Issue #796, opened 2019-05-14
lwan tracking tkt


eyberg:
i don't see a reason to do anything more w/these than provide stubs - lwan wants them

```
eyberg@s1:~/lwan/lwan/build/src/bin/lwan$ cat config.json
{
  "Args": ["./lwan", "-c", "/lwan.conf"],
  "Files": ["lwan.conf"]
}
eyberg@s1:~/lwan/lwan/build/src/bin/lwan$ ops run -d -c config.json -p 8080 lwan
```

```
1 capset
1 nosyscall capset
1 write
Loading configuration file: /lwan.conf
Could not drop capabilities: Function not implemented (error number 38)
1 direct return: 111, rsp 0x769fc870
1 exit_group
exit status 3
```

eyberg:
it gets past the capget calls however the connection doesn't seem to close after an intial request

```
            4 [lwan worker] direct return: 1072, rsp 0x7028009330
            4 [lwan worker] clock_gettime
            4 [lwan worker] clock_gettime: clk_id 6
            4 [lwan worker] direct return: 0, rsp 0x70a31ffcd8
            4 [lwan worker] epoll_wait
            4 [lwan worker] sleep
            4 [lwan worker] wakeup 4->4 0x00000070a9d21bb7
            4 [lwan worker] run frame 0x0000000100a02400, RIP=0x00000070a9d21bb7
            4 [lwan worker] clock_gettime
            4 [lwan worker] clock_gettime: clk_id 6
            4 [lwan worker] direct return: 0, rsp 0x70a31ffcd8
            4 [lwan worker] epoll_wait
            4 [lwan worker] sleep
            4 [lwan worker] wakeup 4->4 0x00000070a9d21bb7
            4 [lwan worker] run frame 0x0000000100a02400, RIP=0x00000070a9d21bb7
            4 [lwan worker] clock_gettime
            4 [lwan worker] clock_gettime: clk_id 6
            4 [lwan worker] direct return: 0, rsp 0x70a31ffcd8
            4 [lwan worker] time
            4 [lwan worker] direct return: 1563911460, rsp 0x70a31ffcf8
            4 [lwan worker] epoll_wait
            4 [lwan worker] sleep
            4 [lwan worker] wakeup 4->4 0x00000070a9d21bb7
            4 [lwan worker] run frame 0x0000000100a02400, RIP=0x00000070a9d21bb7
            4 [lwan worker] close
            4 [lwan worker] close: fd 12
[...]

PR #889, opened 2019-07-03
Fixes for Lwan

trungkiena6:
Set the function for capset/capget to syscall_ignore. So, it always return true.

Signed-off-by: Bui Trung Kien <trungkiena6@gmail.com>


francescolavra:
Both the capget and the sendfile changes are needed for Lwan to work (#796), that's why I put both commits in this PR. I updated the title of the PR to reflect that.

PR #899, opened 2019-07-09
Optimized memcpy implementation

mkwal:
**Implementation Detail:**
1. Converted src and destination pointer to word alligned
2. and then made word copy

**Performance Measurement:**
I used x86 64-bit ubuntu machine to compare the performance of new code with existing approach and gcc libc memcpy.
new code shows considerable improvement.
I also included inline x86 assembly approach (code is not included in review but its performace is even better than gcc libc memcpy).
Let me know if this can be used. I will include the code in review.

**Performance Analysis Report:**	 
Following input data used in evaluating performance with different approach:
Size selection: 64B, 128B, 256B, 512B, 1K, 2K ... 1M
Combination: alligned with 64 byte address, non-aligned destination,  non-aligned destination and src
Number of Loops: From 16777216 to 1024

Complete report is attached. 
[mempcy_performance_analysis.docx](https://github.com/nanovms/nanos/files/3370869/mempcy_performance_analysis.docx)

Following is one sample.
**memcpy 2048B -- 524288 loops**
  **aligned blocks with 64 multiple**
      nanos runtime_memcpy                               2.496000 s
      nanos runtime_memcpy_e                             0.640000 s
      libc memcpy                                        0.060000 s
      memcpy using x86 inline asssembly                  0.036000 s
  **unaligned blocks , src with +0 and dst with +4**
      nanos runtime_memcpy                               2.500000 s
      nanos runtime_memcpy_e                             0.648000 s
      libc memcpy                                        0.060000 s
      memcpy using x86 inline asssembly                  0.048000 s
  **unaligned blocks , src with +10 and dst with +3**
      nanos runtime_memcpy                               2.492000 s
      nanos runtime_memcpy_e                             0.764000 s
      libc memcpy                                        0.060000 s
      memcpy using x86 inline asssembly                  0.052000 s

### DIFF
--- a/boot/Makefile
+++ b/boot/Makefile
@@ -7,6 +7,7 @@ SRCS-stage2.elf= \
 	$(SRCDIR)/runtime/buffer.c \
 	$(SRCDIR)/runtime/extra_prints.c \
 	$(SRCDIR)/runtime/format.c \
+	$(SRCDIR)/runtime/memops.c \
 	$(SRCDIR)/runtime/range.c \
 	$(SRCDIR)/runtime/runtime_init.c \
 	$(SRCDIR)/runtime/pqueue.c \
--- a/mkfs/Makefile
+++ b/mkfs/Makefile
@@ -6,6 +6,7 @@ SRCS-mkfs= \
 	$(SRCDIR)/runtime/extra_prints.c \
 	$(SRCDIR)/runtime/format.c \
 	$(SRCDIR)/runtime/heap/id.c \
+	$(SRCDIR)/runtime/memops.c \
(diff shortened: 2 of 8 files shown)


francescolavra:
I picked up this PR. I approved it because I had previously requested changes on it when it was being worked on by someone else, but this approval now doesn't count. So I'll wait for somebody to review it.

eyberg:
cc @mkhon @wjhun

francescolavra:
OK, I added to the unit test a memcpy test for all combinations of src_cnt and dest_cnt.

Issue #184, opened 2018-11-01
Add unit test for all basic data structures.


tijoytom-zz:
-bitmap
-pqueue
-buffer
-rtrie ( maybe not for now)

uma-sethuraman:
Hello, my team members and I are students at the University of Texas at Austin. We are looking to contribute to Nanos as a part of a project for our virtualization class. Could we take on this issue? Would this be a good first issue to tackle?

eyberg:
sorry just now seeing this, yes please feel free to work on this, i'd just break up individual tests against the various data structures into separate pull requests, please reach out if you need more direction/help

wjhun:
Hello Uma. This issue is very old and was out-of-date, so I updated it. The only item left here is a unit test for bitmaps. Bitmaps are tested indirectly in the id heap test, but an explicit unit test would still be welcome if you are interested in picking that up.

We need a good futex runtime test, too, but that isn't really a data structure test. I'll open a new issue for that in case you or your colleagues are interested in picking that up (see #1306).

uma-sethuraman:
Hi Ian and Will, 
Thank you for the helpful information! My teammates and I will go ahead and start working on this issue. We're also interested in issue #1306 which Will referenced in his reply, so I commented on that issue as well. We'll let you know if we have any additional questions as we start working on this issue.

samantha3pen:
Hello @eyberg and @wjhun , I'm one of Uma's teammates and I'll be helping with this issue. Thank you for the updates! So far, I've begun to familiarize myself with the bitmap.c and id_heap_test.c files, but I'd appreciate having some more direction/help on where I should start with creating unit tests for bitmaps and what your expectations are.

wjhun:
Hello @samantha3pen, and thank you for offering to contribute. The id_heap_test unit test exercises the bitmap code somewhat, but this is only a brute-force (randomized) test, and it doesn't exercise the bitmap interface at all.

A good unit test here should, at least:

- test all bitmap functions and macros (like bitmap_*foreach*) for correct behavior
- for functions that return some error result, stress that the correct result is given for a set of inputs (including invalid arguments)
  - e.g. bitmap_range_check_and_set should fail if validate is set to true and some bits in the query range have already been set (if set is true) or cleared (if !set).
- test for invariant conditions
  - e.g. allocations from the bitmap should be aligned to the power-of-2 greater than or equal to the allocation size

[...]

PR #916, opened 2019-07-12
WIP: Add basic unit testing for random

x64k:
Part of #184.

Add a basic set of unit tests for random(). Currently, this relies on
ent and performs only the most trivial tests.

While this is not sufficient to exhaustively test the quality of our
implementation, it should be reasonably sufficient to expose the most
trivial bugs that we are likely to encounter.

### DIFF
--- a/test/unit/random_test.c
+++ b/test/unit/random_test.c
@@ -1,0 +1,268 @@
+/**
+ * Basic unit testing for random implementation, using Fourmilab's ent.
+ *
+ * Note that this is not a set of "true" randomness unit tests, the way
+ * DieHard is. It's intended to quickly expose pathological and trivial
+ * bugs and to test the API.
+ *
+ * We definitely should look into a true random unit test (TestU01,
+ * DieHard, DieHarder...)
+ */
+
+#include <runtime.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <string.h>
+#include <errno.h>
+
+/* Number of random bytes to output for test. */
+#define BSLEN 16384
+/* Number of bytes used by gen_bytestream's internal buffer */
+#define INTERNAL_BUFLEN 1024
+/* Number of values that ent gives out. */
+#define ENT_NSTATS 6
+/* Length of a stat entry key */
+#define KEY_LEN 64
+
+/* Maximum acceptable compressibility value for a reasonably long bytestream */
+#define COMPRESS_PASS_THRESH 10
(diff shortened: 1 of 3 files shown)


test/unit/random_test.c
```diff
@@ -0,0 +1,228 @@
...
+ */
+static void gen_bytestream(char *path, int len)
+{
+    int fd, err;
+    u64 buf[INTERNAL_BUFLEN / 8];
+    u64 i, nbytes = 0;
+
+    fd = open(path, O_CREAT | O_WRONLY, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
+    if (fd == -1)
+    {
```
wjhun: Kind of a nit, but for the sake of consistency: Could you please keep opening braces following the conditional expression?

test/unit/random_test.c
```diff
@@ -0,0 +1,228 @@
...
+    while (nbytes < BSLEN) {
+        for (i = 0; i < INTERNAL_BUFLEN / 8; i++)
+            buf[i] = random_u64();
+
+        nbytes += INTERNAL_BUFLEN;
+
+        err = write(fd, &buf[0], INTERNAL_BUFLEN);
+        if (err == -1)
+        {
+            msg_err("error writing to bytestream file.\n");
```
wjhun: Maybe print strerror(errno), and also check for EINTR here.

test/unit/random_test.c
x64k: Not a nit at all! I'll fix it before submitting -- sorry for the slip-up, it's the coding style I've used for the last three years or so in a full-time project and I haven't fully flushed it out of my system yet :). I'll make a local git hook with indent or something to make sure I don't end pushing it again.

test/unit/random_test.c
x64k: Sure thing, thanks!

x64k:
Ah, cool, I didn't realize the CI config files are in this tree as well, I thought I'd have to pester someone to figure out how to modify the config files.

I'll push an amended commit tomorrow (I mean, well, it's already today in my timezone :) ).

Thanks for the review!

x64k:
OK, I fixed the indenting, the missing EINTR case (doh!) and fixed the CircleCI dependencies.

I've also added a "-s" flag which, when passed, causes random_test to show the statistics it uses in human-readable format. I figured it would be useful for development/debugging -- pass/fail is useful for automated testing, but this program is also a pretty straightforward way to test the PRNG while hacking on it, when being able to see the actual values is nice. The flag is off by default, of course.

All good?

x64k:
Awesome! I'll do the merge later today. Thanks for the review!

Issue #901, opened 2019-07-09
python 3 failing


eyberg:
```
➜  t  ops load python_3.6.7 -n -c config.json
Downloading.. https://storage.googleapis.com/packagehub/python_3.6.7.tar.gz
 14.50 MiB / 14.50 MiB [=============================================================================================================================================================] 100.00% 9.32 MiB/s 1s
Downloading.. https://storage.googleapis.com/nanos/release/nightly/nanos-nightly-darwin.tar.gz
 588.33 KiB / 588.33 KiB [===========================================================================================================================================================] 100.00% 7.42 MiB/s 0s
[python3 -m http.server]
booting /Users/eyberg/.ops/images/python3.img ...
assigned: 10.0.2.15
Unhandled: 000000000000000d
General protection fault
interrupt: 000000000000000d
frame: 0000000101001600
error code: 0000000000000000
  rax: 0000000100549f48
  rbx: 000000010054a018
  rcx: 0000000100549f48
  rdx: 0000000100803000
  rsi: 000000010054a060
  rdi: 001d001e00400007
  rbp: 000000007fb84ef0
  rsp: 000000007fb84eb8
   r8: 0000000003000000
   r9: 0000000100803000
  r10: 00000000000001ff
  r11: 0000000000000000
  r12: 0000000103804000
  r13: 0000000103806000
  r14: 0000000000000004
  r15: 0000000100776140
  rip: 000000007f016f0d (notify_remove + 000000000000002d/0000000000000061)
flags: 0000000000000006
   ss: 0000000000000010
   cs: 0000000000000008
   ds: 0000000000000000
   es: 0000000000000000
   fs: 0000000000000000
   gs: 0000000000000000

frame trace:
000000007f018e3d        (poll_internal + 000000000000028d/00000000000008fb)
000000007f023feb        (syscall_debug + 000000000000018b/000000000000023c)
000000007f0002d2        (syscall_enter + 000000000000006a/0000000000000072)

stack trace:
000000007f017942        (free_epollfd + 0000000000000162/000000000000022b)
0000000000000000
0000000000000000
0000000103806000
00000000721f81a8
0000000103804000
0000000000000000
000000007fb84f80
000000007f018e3d        (poll_internal + 000000000000028d/00000000000008fb)
0000000103805050
000000010054a160
0000000000000000
0000000000000001
0000000000000000
0000000500000000
0000000103805000
000000007f04a668        (bootstrap_region + 0000000000000b28/0000000000200000)
[...]

PR #926, opened 2019-07-19
WIP: Preserve physical addresses in mremap

x64k:
Ensure that, when resizing a memory region via mremap, the existing
virtual -> physical mappings of the previously mapped pages from the
original region are preserved in the new region.

Tentative fix for #901.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -206,15 +206,10 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void 
     }
     thread_log(current, "   new physical pages at 0x%lx, size %ld", dphys, dlen);
 
-    /* remove old mapping */
-    u64 oldphys = physical_from_virtual(old_address);
-    thread_log(current, "   old mapping at phys addr 0x%lx, unmapping", oldphys);
-    unmap(u64_from_pointer(old_address), old_size, pages);
-
-    /* map existing portion */
+    /* Remap existing portion to new address, preserving flags and
+       physical mappings. */
     u64 mapflags = page_map_flags(vm->flags);
-    thread_log(current, "   mapping existing portion at 0x%lx", vnew);
-    map(vnew, oldphys, old_size, mapflags, pages);
+    vremap(old_address, old_size, vnew, mapflags, pages);
 
     /* map new portion and zero */
     thread_log(current, "   mapping and zeroing new portion at 0x%lx", vnew + old_size);
--- a/src/x86_64/page.c
+++ b/src/x86_64/page.c
@@ -400,6 +400,25 @@ void map(u64 virtual, physical p, int length, u64 flags, heap h)
     map_range(virtual, p, length, flags | PAGE_PRESENT, h);
 }
 
+void vremap(void* old_virtual, u64 old_size, u64 new_virtual, u64 flags, heap h)
+{
+    physical phys;
+    u64 crt_off = 0;
+    u64 stride = PAGESIZE;
+
+    while (crt_off < old_size)
+    {
+        phys = physical_from_virtual(pointer_from_u64(old_virtual + crt_off));
+        if (phys == INVALID_PHYSICAL)
+                continue;
+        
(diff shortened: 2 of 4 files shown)


ssudake21:
I know your changes are currently WIP. Can you please remove skip flag here once you finalize change ? Change will get verified by CI then.
https://github.com/nanovms/nanos/blob/master/test/e2e/e2e.go#L71

x64k:
Opening to gather feedback on the approach. The implementation can probably be optimize -- this is a quick prototype I whipped up to test my analysis of the root cause of #901. I don't see crashes anymore, and the mappings look right now.

I think the "original" problem was that, between mremap and map_range, there was a "hidden" assumption that consecutive pages of virtual memory would be mapped to consecutive regions of physical memory. That's not necessarily the case. Python anonymously mmapped large chunks of memory, but didn't access them linearly, so the virtual -> physical mappings are scattered all over the place.

This adds a vremap function which (naively, for now -- like I said, I'm just testing the overall approach) goes through every page in the "old" region, unmaps its virtual -> physical pair, and then maps the corresponding page in the "new" region to the same old physical page.

It seems to take care of the crash for now.

In addition to the obvious question (i.e. does this look right to you?), I have a few other ones:

- Is there a "right" way to find out what the current page size is (i.e. 4K? 2M?) I'm guessing there's a flag somewhere, which we could maybe abstract in a macro or a small function? I'm not really fond of writing PAGESIZE/PAGESIZE_2M all over the place...
- Not related but I stumbled upon it while looking for the answer to the question above:

```
static void zero_page(u64 addr, u64 * pte)
{
    zero(pointer_from_u64(addr), *pte & PAGE_2M_SIZE ? PAGE_2M_SIZE : PAGESIZE);
}
```

Shouldn't the second PAGE_2M_SIZE be PAGESIZE_2M :)?
- Would it make sense to merge this special(-ish?) case under map_range? I'm thinking mostly no, I think we can generally know when we're mapping a fresh range and when we're remapping an existing range and we can invoke the "right" function (map range vs. vremap). That would possibly allow us to come up with a more efficient implementation of the latter without making the former more complicated (and slower), but I'm always happy to be proven wrong :octocat: .

@ssudake21 yep, thanks for reminding me! I can't say I have a good reason for not including it in the WIP PR in the first place, I think I'll just force push another change :).

wjhun:
This looks like the right diagnosis, though really I think we should add a remap function to the page table code itself, making use of "traverse_range" which will walk the page tables and operate on each leaf PTE at a time. This can address 2M mappings, etc. For an example, see update_map_flags in src/x86_64/page.c: the update_pte_flags closure is called for each leaf pte (you can determine if it's 2M by looking at the passed pte - and yes, that is a bug you found in zero_page).

So a remap() function in the page table code would, with each leaf PTE, extract the physical address, clear the PAGE_PRESENT flag in the PTE (or zero the whole entry), then call map_page to install the new entry. And of course don't forget the invalidate or you'll be chasing heisenbugs for a day (been there, done that). Don't worry about overlapping regions; mremap() doesn't support MREMAP_FIXED yet anyway.

Let me know what you think - I'm happy to whip up a new PR or you can take a crack at it if you wish.

x64k:
Yep, I was primarily interested in figuring out if preserving the mapping is the right fix :). Now that I look at traverse_range & friends, it's obvious that that's the level where the remapping ought to happen, especially since it wouldn't leak abstractions all over the place. In fact, I suspect the naive version in this PR does the same thing but with a far worse complexity, since instead of walking the PTEs it "walks" the page range, deriving the physical address of each one, then doing a map - unmap (thus doing map_range twice).

I'll leave it to you and @eyberg to decide who should whip up the proper PR for this. I can take a crack at it, but I'm pretty new here and this is core functionality, so I definitely understand if you'd rather have a core team member do it. I'm not one of those "my issue, my code" people -- I'm fine with whatever gets us moving forward more efficiently.

In the meantime, I'll make a separate PR with the fix for zero_page (+ I think unmap_page needs the same treatment, too).

x64k:
Obsoleted by https://github.com/nanovms/nanos/pull/932 , closing.

PR #932, opened 2019-07-21
fix mremap(2) / python 3 e2e failure

wjhun:
As @x64k discovered in PR #926, mremap(2) was not properly handling physically-discontiguous mappings. This is resolved in this PR by introducing remap_pages() into the page table code. This walks the page table for a given mapping, adds entries for the target region and removes old ones. A few other subtle bugs in the page table code are addressed as well (missing invalidate and incorrect translation of pte to phys addr in unmap_page, and the wrong 2M pagesize define mentioned in #926).

As in #926, the python_3.6.7 e2e test is re-enabled here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     steps:
       - checkout
       - run: sudo apt-get update
-      - run: sudo apt-get install nasm qemu ent ruby rustc
+      - run: sudo apt-get install nasm qemu-system-x86 ent ruby rustc
       - run: cd .. && git clone git@github.com:nanovms/ops.git
       - run: make
       - run: curl https://ops.city/get.sh -sSfL | sh
@@ -23,7 +23,7 @@ jobs:
     steps:
       - checkout
       - run: sudo apt-get update
-      - run: sudo apt-get install nasm qemu ent ruby rustc
+      - run: sudo apt-get install nasm qemu-system-x86 ent ruby rustc
       - run: make
       - run: curl https://ops.city/get.sh -sSfL | sh
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     steps:
       - checkout
       - run: sudo apt-get update
-      - run: sudo apt-get install nasm qemu ruby rustc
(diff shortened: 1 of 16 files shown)


wjhun:
I included a fix for the CI failure. The docker image must have changed (to debian 10?) and the 'qemu' package wasn't sufficient to pick up qemu-system-x86_64.

x64k:
Yep, I think this is the right stuff :).

Regarding this:

> seems like this path could be simplified / re-written using traverse

I think we would gain a little bit of performance by skipping a level of indirection but I'm not sure the loss in readability would justify it, not at this point anyway, when we still have a bit to go in terms of features (e.g. MREMAP_FIXED). It looks like the kind of thing that we could write up under the proverbial root of all evil :).

wjhun:
> Regarding this:
> 
> > seems like this path could be simplified / re-written using traverse
> 
> I think we would gain a little bit of performance by skipping a level of indirection but I'm not sure the loss in readability would justify it, not at this point anyway, when we still have a bit to go in terms of features (e.g. MREMAP_FIXED). It looks like the kind of thing that we could write up under the proverbial root of all evil :).

Oh, that comment of mine was kind of obtuse. What I meant was that the page table code could stand to use a little refactoring so that we don't unnecessarily have two different ways of walking the tables (the original recursive force_entry for creating entries and the iterative traverse_range for updating/unmapping). traverse_range could be changed to call the closure for all ptes, not only ones with PAGE_PRESENT set, thus enabling new entries to be installed. Of course, middle entries would need similar treatment so that new pages can be inserted where needed.

wjhun:
So I just pushed another change that alters the behavior of page traversal (traverse_range() -> traverse_entries()). The new routine will call the handler for each entry in the traversal path, regardless of table level or presence. This allows us to detect unmapped areas in remap_pages() and correspondingly remove any pages that may be mapped in the target area. Now validate_virtual() is written in terms of traverse_entries(). (I didn't change force_entry, etc., but as mentioned this can later be refactored and written using traverse as well.) Also included some other debug cleanup.

@x64k, would you mind taking a look at the update? Thanks!

x64k:
I think I like traverse_entries :).

PR #934, opened 2019-07-22
adding in logo

eyberg:


wjhun:
If I view the README.md, the logo has a black rectangular background and grey text which is pretty low-contrast. Did you mean to make the background transparent?

eyberg:
no - not sure why imgur uploaded it that way - it's supposed to be transparent but then github blew up - i'm going to fix it in a bit

PR #939, opened 2019-07-23
Make open_internal UTF-8-aware

### LINKED ISSUE #567: fs doesn't see utf-8 files
i was looking to verify this -> 

https://github.com/nanovms/nanos/issues/389

and while that seems to work for whatever reason tfs doesn't seem to see it ->

```
eyberg@s1:~/g$ ops run -c config.json g
Finding dependent shared libs
booting /home/eyberg/.ops/images/image ...
warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
assigned: 10.0.2.15
open /✔.txt: no such file or directory
/
/dev
/dev/null
/dev/urandom
/etc
/etc/resolv.conf
/g
/kernel
/proc
/proc/sys
/proc/sys/kernel
/proc/sys/kernel/hostname
/some.txt
exit_group
exit status 1
eyberg@s1:~/g$ cat config.json
{
  "Files": ["✔.txt", "some.txt"]
}
```

```
eyberg@s1:~/g$ cat main.go
package main

import (
    "fmt"
    "io/ioutil"
    "os"
    "path/filepath"
)

func main() {

    dat, err := ioutil.ReadFile("/✔.txt")
    if err != nil {
        fmt.Println(err)
    }

    fmt.Print(string(dat))

    var files []string

    root := "/"
    err = filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
        files = append(files, path)
        return nil
    })
    if err != nil {
        panic(err)
    }
    for _, file := range files {
        fmt.Println(file)
    }

}
```

x64k:
open_internal was attempting to lookup a file's tuple by doing
resolve_cstring. However, resolve_cstring is not parsing UTF-8 strings
correctly; it works only on one byte at a time, and while push_character
is UTF-8-aware, it can't properly encode values given one byte at a
time.

It's not unthinkable that we might want to restrict some things to
ASCII-only. So this commit adds an UTF-8 counterpart called
resolve_utf8, and amends open_internal to use it.

Closes #567.

### DIFF
--- a/src/runtime/text.h
+++ b/src/runtime/text.h
@@ -40,7 +40,7 @@ static int inline string_character_length(char *s) {
     for (character __i; __i = pop_character(__s), __i != CHARACTER_INVALID;)
 
 
-static inline character utf8_decode(u8 *x, int *count)
+static inline character utf8_decode(const u8 *x, int *count)
 {
     if ((x[0] & 0xf0) == 0xf0) {
         *count = 4;
@@ -67,7 +67,6 @@ static inline character utf8_decode(u8 *x, int *count)
     return *x;
 }
 
-
 static inline void push_character(string s, character c)
 {
     if (c<0x80) {
@@ -86,6 +85,23 @@ static inline void push_character(string s, character c)
         buffer_write_byte(s, 0x80 | (c&0x3f));
     }
 }
+
+/**
+ * Push an UTF-8 character from a sequence of bytes p and return the
+ * number of bytes consumed from p.
+ */
+static inline int push_utf8_character(string s, const char *p)
+{
+    int nbytes = 0;
+    u32 crt_char;
+
+    crt_char = utf8_decode((const u8 *)p, &nbytes);
+
+    push_character(s, crt_char);
+
+    return nbytes;
+}
+
 
 // xxx - check
 #define CHARACTER_INVALID 0xfffffffful
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -263,19 +263,27 @@ static inline tuple resolve_cstring(tuple cwd, const char *f)
 
     buffer a = little_stack_buffer(NAME_MAX);
     char y;
-    while ((y = *f++)) {
+    int nbytes;
+
+    while ((y = *f)) {
         if (y == '/') {
             if (buffer_length(a)) {
                 t = lookup(t, intern(a));
                 if (!t)
                     return t;
                 buffer_clear(a);
-            }                
+            }
+            f++;
         } else {
-            push_character(a, y);
+            nbytes = push_utf8_character(a, f);
+            if (!nbytes) {
+                thread_log(current, "Invalid UTF-8 sequence.\n");
+                return 0;
+            }
+            f += nbytes;
         }
     }
-    
+
     if (buffer_length(a)) {
         t = lookup(t, intern(a));
     }
@@ -289,7 +297,9 @@ static inline tuple resolve_cstring_parent(tuple cwd, const char *f)
     tuple parent = 0;
     buffer a = little_stack_buffer(NAME_MAX);
     char y;
-    while ((y = *f++)) {
+    int nbytes;
+    
+    while ((y = *f)) {
         if (y == '/') {
             if (buffer_length(a)) {
                 if (!t) {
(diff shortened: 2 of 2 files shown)


x64k:
This is a limited (but expandable) fix for #567. For now, it just enables us to open paths that contain UTF-8 characters. It should be followed by one, probably two refactoring commits, but it works okay in this manner, too, which is why I'm not labeling it as WIP.

At first, I considered amending resolve_cstring to do the right thing and work on UTF-8 streams, too. Unfortunately, that required either a bit of refactoring on push_characters (which is used in a lot of places and would have messed up a lot of things) so as to return the number of bytes consume from the byte sequence that was passe to it, or a pretty messy version of resolve_cstring so as to clumsily attempt to resolve UTF-8 sequences on its own.

Since it also makes sense that we'd want to restrict some things to ASCII-only, I just went ahead and added an utf-8 counterpart to resolve_cstring.

We probably want to have the rest of the fs code use resolve_utf8, but I'd rather do that in another commit (not necessarily another PR, I can reuse this one).

Refactoring resolve_cstring & friends is still on the table if we want it, of course...

src/runtime/text.h
```diff
@@ -87,6 +86,23 @@ static inline void push_character(string s, character c)
     }
 }
 
+/**
+ * Push an UTF-8 character from a sequence of bytes p and return the
+ * number of bytes consume from p.
```
francescolavra: Nitpick: s/consume/consumed

src/runtime/text.h
```diff
@@ -87,6 +86,23 @@ static inline void push_character(string s, character c)
...
+ */
+static inline int push_utf8_character(string s, const char *p)
+{
+    int nbytes = 0;
+    u32 crt_char = 0x00;
```
francescolavra: There is no need to initialize crt_char here, since you are going to overwrite its value in the next line.

src/unix/syscall.c
```diff
@@ -283,6 +283,45 @@ static inline tuple resolve_cstring(tuple cwd, const char *f)
...
+{
+    tuple t = *f == '/' ? filesystem_getroot(current->p->fs) : cwd;
+
+    buffer a = little_stack_buffer(3 * NAME_MAX);
+    char y;
+    int nbytes;
+
+    if (buffer_length(a)) {
+        t = lookup(t, intern(a));
+    }
```
francescolavra: Here, the buffer has just been allocated and is guaranteed to be empty (buffer_length() will always return 0), so I would remove this if() block.

src/runtime/text.h
x64k: Oops! Will fix!

src/runtime/text.h
x64k: You're right, it's a leftover from a previous version of this function :). I'll remove the initialization.

src/unix/syscall.c
x64k: This one is _definitely_ a copy-paste error, it's the same block as at the end of the function. I have no good excuse for why it made it into a PR :(.

x64k:
> I'm personally in favor of amending resolve_cstring (and resolve_cstring_parent) to properly support UTF8. 

Let me run a quick test tomorrow -- I'll check to see if resolve_cstring is used anywhere outside the runtime fs code, and run a few tests to see if resolve_utf8 works instead. If amending it doesn't end up touching code that I can't test, and if nothing breaks, I'll push another commit with an amended version.

I try to be careful with changes that could have a deep impact when I'm new in a project, but in this case I think I've been overly cautious and ended up pushing something that's not as good as it could be.

x64k:
Just pushed a revised version. I fixed the typos and the leftovers (though I'm still embarrassed about the latter...), and revised resolve_cstring, resolve_cstring_parent and filepath_is_ancestor so that they're all UTF-8 aware. 

The additional revise functions (resolve_cstring_parent, filepath_is_ancestor) are called as part of rename & friends. I only tried to run very basic tests (whipped up a go program that just called os.Rename on a couple of paths, with and without unicode characters in the path) and I poked at some files from the Python image. Between that and the passing tests, I'm reasonably convinced that I haven't  introduced any regressions.

src/unix/syscall.c
```diff
@@ -263,19 +263,28 @@ static inline tuple resolve_cstring(tuple cwd, const char *f)
 
     buffer a = little_stack_buffer(NAME_MAX);
     char y;
-    while ((y = *f++)) {
+    int nbytes;
+
+    while ((y = *f))
+    {
```
francescolavra: The opening brace should be after the while() clause in the preceding line.

x64k: Right! Sorry, I'm still getting the coding style from my last project out of my system!

PR #941, opened 2019-07-23
clock_gettime(2): properly handle monotonic clock

### LINKED ISSUE #357: clock issues
taken form https://github.com/nanovms/nanos/pull/354

```
eyberg@s1:~/cl$ cat main.c
#include <stddef.h>
#include <stdio.h>
#include <time.h>
#include <sys/sysinfo.h>

int main(void){
    struct sysinfo info;
    sysinfo(&info);

    const time_t boottime = time(NULL) - info.uptime;

    struct timespec monotime;
    clock_gettime(CLOCK_MONOTONIC, &monotime);

    time_t curtime = boottime + monotime.tv_sec;

    struct timespec realtime;
    clock_gettime(CLOCK_REALTIME, &realtime);

    printf("Boot time = %s", ctime(&boottime));
    printf("Current time = %s", ctime(&curtime));
    printf("Real Time = %s", ctime(&realtime.tv_sec));

    return 0;
}
```

```OpenJDK 64-Bit Server VM warning: No monotonic clock was available - timed services may be adversely affected if the time-of-day clock changes
```

```
eyberg@s1:~/cl$ ./main
0
1
Boot time = Wed Aug 22 18:38:07 2018
Current time = Thu Jan 10 00:40:00 2019
Real Time = Thu Jan 10 00:40:01 2019
```

```
eyberg@s1:~/cl$ ops run -p8083 -f main
Finding dependent shared libs
booting image ...
warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
pages heap: 000000007fb75000, length 00000000002f7000
physical memory:
   base 0000000000200000, length 000000007f600000
assigned: 10.0.2.15
0
1
Boot time = Thu Jan 10 00:40:15 2019
Current time = Thu Jan 19 01:20:31 2068
Real Time = Thu Jan 10 00:40:16 2019
exit_groupexit status 1
```

and if you look @ http://man7.org/linux/man-pages/man3/sleep.3.html you can see it uses nanosleep which uses... http://man7.org/linux/man-pages/man2/nanosleep.2.html

&&&.... https://golang.org/src/time/time.go

francescolavra:
clock_gettime(2): properly handle monotonic clock

The now() function returns the wall clock time, so it will be subject to discontinuities if the system time is changed (when support for changing the system time is actually implemented), and for this reason it should not be used as time source in clock_gettime() in case the requested clock is a monotonic clock. This commit uses uptime() as time source in such cases.

Closes #357.

### DIFF
--- a/src/unix/system_structs.h
+++ b/src/unix/system_structs.h
@@ -168,6 +168,14 @@ typedef struct iovec {
 
 typedef int clockid_t;
 
+#define CLOCK_REALTIME              0
+#define CLOCK_MONOTONIC             1
+#define CLOCK_PROCESS_CPUTIME_ID    2
+#define CLOCK_THREAD_CPUTIME_ID     3
+#define CLOCK_MONOTONIC_RAW         4
+#define CLOCK_REALTIME_COARSE       5
+#define CLOCK_MONOTONIC_COARSE      6
+
 struct timespec {
 	u64 ts_sec;
 	u64 ts_nsec;
--- a/src/unix/unix_clock.c
+++ b/src/unix/unix_clock.c
@@ -48,7 +48,25 @@ sysreturn times(struct tms *buf)
 sysreturn clock_gettime(clockid_t clk_id, struct timespec *tp)
 {
     thread_log(current, "clock_gettime: clk_id %d", clk_id);
-    timespec_from_time(tp, now());
+    timestamp t;
+    switch (clk_id) {
+    case CLOCK_REALTIME:
+    case CLOCK_REALTIME_COARSE:
+        t = now();
+        break;
+    case CLOCK_PROCESS_CPUTIME_ID:
+        t = proc_utime(current->p) + proc_stime(current->p);
+        break;
+    case CLOCK_MONOTONIC:
+    case CLOCK_MONOTONIC_COARSE:
+    case CLOCK_MONOTONIC_RAW:
+        t = uptime();
+        break;
+    default:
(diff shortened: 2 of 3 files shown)


src/unix/system_structs.h
```diff
@@ -168,6 +168,14 @@ typedef struct iovec {
 
 typedef int clockid_t;
 
+#define CLOCK_REALTIME              0
+#define CLOCK_MONOTONIC             1
+#define CLOCK_PROCESS_CPUTIME_ID    2
```
wjhun: Can this be implemented using the time accounting you added for times(2)?

src/unix/unix_clock.c
```diff
@@ -48,7 +48,21 @@ sysreturn times(struct tms *buf)
...
+    case CLOCK_REALTIME_COARSE:
+        t = now();
+        break;
+    case CLOCK_MONOTONIC:
+    case CLOCK_MONOTONIC_COARSE:
+    case CLOCK_MONOTONIC_RAW:
+        t = uptime();
+        break;
+    default:
+        return -EINVAL;
```
wjhun: Should we emit a warning for unimplemented clk_ids?

src/unix/system_structs.h
francescolavra: Yes, it can, good point. I added support for this clock type too.

src/unix/unix_clock.c
francescolavra: OK, I added a call to msg_warn().

PR #943, opened 2019-07-25
replace fixed timeout in e2e tests with retry with exponential backoff

wjhun:
This is an attempt to fix the spurious "connection reset" errors during CI runs. This removes the fixed 10 second delay instead tries the first http request of each test up to 5 times (with exponential backoff) before giving up.


wjhun:
Actually, this should only retry on the first request per tt, not subsequent ones. Will update later.

PR #954, opened 2019-08-14
add runtime test for mmap/munmap/mincore/mremap

briankoco:
New tests added un test/runtime/mmap.c
Run with "make TARGET=mmap run"

### DIFF
--- a/test/runtime/mmap.c
+++ b/test/runtime/mmap.c
@@ -1,0 +1,533 @@
+/* tests for mmap, munmap, mremap, and mincore */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdbool.h>
+
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include <runtime.h>
+
+#define NR_MMAPS	3000
+#define ALLOC_AT_A_TIME	150
+
+typedef struct {
+    void * addr;
+    unsigned long size;
+} mmap_t;
+
+/* round up to nearest page address */
+#define round_up_page(addr) \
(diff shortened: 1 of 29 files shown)


test/runtime/mmap.c
```diff
@@ -0,0 +1,533 @@
...
+	map_addr = tmp;
+	map_size = new_size;
+    }
+
+    printf("** all mremap tests passed\n");
+}
+
+int main(int argc, char * argv[])
+{
+    mmap_test();
```
wjhun: You may want to unbuffer stdout here, e.g. "setbuf(stdout, NULL)" so that printfs are in line with any possible error messages from nanos. (And then the fflushes above can be removed.)

test/runtime/mmap.c
```diff
@@ -0,0 +1,533 @@
...
+    vec = malloc(sizeof(u8) * 10);
+    if (vec == NULL) {
+	perror("malloc failed");
+	exit(EXIT_FAILURE);
+    }
+
+    /* test something on the stack */
+    printf("  performing mincore on stack address...\n");
+    fflush(stdout);
+    addr = (void *)round_down_page(mincore_test);
```
wjhun: Wouldn't mincore_test be an address on a code page from the program ELF?

briankoco: yep, my mistake

test/runtime/mmap.c
briankoco: thanks for the tip

test/runtime/mmap.c
```diff
@@ -0,0 +1,533 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdbool.h>
+
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include <runtime.h>
```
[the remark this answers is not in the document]
briankoco: ok thanks, I was using it to pull in PAGESIZE and other macros, but those are easy enough to add directly to the source

PR #955, opened 2019-08-14
add sanity checks for mincore/mmap arguments

briankoco:


src/unix/mmap.c
```diff
@@ -224,6 +224,10 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
 
 static sysreturn mincore(void *addr, u64 length, u8 *vec)
 {
+    u64 where = u64_from_pointer(addr);
+    if (where & MASK(PAGELOG))
+        return -EINVAL;
+
     if (validate_virtual(addr, length)) {
```
wjhun: Could we also pad the length up here? e.g. length = pad(length, PAGESIZE);

So actually, this implementation of mincore has long bitrotted and isn't really sufficient. I just filed an issue for this - #956.

src/unix/mmap.c
```diff
@@ -557,16 +561,19 @@ static sysreturn mmap(void *target, u64 size, int prot, int flags, int fd, u64 o
...
         range q = irange(where, where + len);
         if (!mmap_reserve_range(p, q)) {
-            if (fixed) {
-                thread_log(current, "   fail: fixed address range %R outside of lowmem or virtual_page heap\n", q);
-                return -ENOMEM;
-            } else {
-                where = 0;
-            }
+	    thread_log(current, "   fail: fixed address range %R outside of lowmem or virtual_page heap\n", q);
+	    return -ENOMEM;
```
wjhun: Yes, that conditional was redundant.

Issue #956, opened 2019-08-15
re-implement mincore(2)


wjhun:
The existing mincore(2) implementation was written before we had demand paging and has since bit-rotted. A new implementation could add a function to src/x86/page.c which uses traverse_entries() to act on each present page within a range (similar to zero_mapped_pages, but calling a closure passed from the outside the page table code), then calling said function with a mincore-specific closure (e.g. "mincore_page") to fill the vector as necessary.

briankoco:
Addressed in PR #961

PR #957, opened 2019-08-15
Reimplementation of mincore (as noted in #956)

briankoco:
Note that we are not currently able to return -ENOMEM in cases where
the region covers unallocated memory, as this would require tracking of
memory in stack/heap/bss

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -16,42 +16,6 @@ static boolean vmap_attr_equal(vmap a, vmap b)
     return a->flags == b->flags;
 }
 
-/* Page faults may be caused by:
-
-   - user program instructions
-
-   - syscall top halves accessing unmapped anonymous pages
-
-   - syscall bottom halves accessing unmapped anonymous pages (while
-     in the interrupt handler!)
-
-     - can consider manually faulting in user pages in top half to
-       avoid faults from interrupt handler; this is debatable
-
-   Therefore:
-
-   - allocating a physical page must be fast and safe at interrupt
-     level
-
-     - as elsewhere in the kernel, if/when we move from the bifurcated
-       runqueue / interrupt processing scheme, we need to put the
-       proper locks in place
-
-     - we can easily build a free list of physical pages
-
-       - also note that, given that id deallocations don't need to
-         match their source allocations, we can take any size
-         deallocation and bust it up into single pages to cache
-
-   - map() needs to be safe at interrupt and non-interrupt levels
-
-   - the page fault handler runs on its own stack (set as IST0 in
-     TSS), given that the user stack may live on an anonymous mapping
-     and need to have pages faulted in on its own behalf - otherwise
-     we eventually wind up with a triple fault as the CPU cannot push
-     onto the stack when invoking the exception handler
-*/
-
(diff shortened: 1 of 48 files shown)


src/x86_64/page.c
```diff
@@ -423,6 +423,32 @@ void zero_mapped_pages(u64 vaddr, u64 length)
...
+                vec[pgoff + i] = 1;
+        } else if (entry_is_pte(level, e)) {
+            vec[pgoff] = 1;
+        }
+    }
+
+    return true;
+}
+
+void mincore_pages(u64 vaddr, u64 length, u8 *vec)
```
wjhun: In general, we've tried to keep the unix-specific stuff confined to the unix world (i.e. src/unix) and the low-level kernel stuff as generic as possible. I think in this case it would make sense to make traverse_entries() global (and make the name more descriptive, e.g. traverse_ptes or traverse_page_tables) and then move is_page_mapped back to mmap.c and invoke the traverse directly from within mincore().

src/unix/mmap.c
```diff
@@ -222,15 +222,49 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
...
-        return 0;
+    u64 start, i;
+
+    start = u64_from_pointer(addr);
+    if (start & MASK(PAGELOG))
+        return -EINVAL;
+
+    length = pad(length, PAGESIZE);
+
+    /* XXX: we have no way to know whether there is unmapped memory in this range,
```
wjhun: It looks like perhaps this was part of an earlier, abandoned solution, but it's not clear to me why we even need to look at vmaps here. The page tables should tell us all we need to know, regardless of the type of mapping or meta.

The comment is correct for other purposes, however, in that we're not keeping a vmap for stuff like loaded ELF segments, the stack, heap, etc.. At some point - e.g. if we need to emulate /proc/self/maps - we should be able to easily add these. But I don't think it matters for mincore.

src/unix/mmap.c
```diff
@@ -557,16 +591,19 @@ static sysreturn mmap(void *target, u64 size, int prot, int flags, int fd, u64 o
             thread_log(current, "   attempt to map zero page");
             return -ENOMEM;
         }
+
+	/* Must be page-aligned */
```
wjhun: Is this hunk from a different PR?

src/unix/mmap.c
briankoco: mincore specifies that it wants -ENOMEM if anything in the specified range is not mapped. Conversely, if everything is mapped but some of the pages are not currently in RAM (e.g., not yet demand paged), fill the vec with 0. Unless I'm missing something, making the distinction between not mapped and not yet demand-paged required more info than we can get from the page tables -- e.g., a vmap

src/x86_64/page.c
briankoco: OK, makes sense

src/unix/mmap.c
wjhun: Sorry, my mistake. Of course it wouldn't make sense for mincore to allow the inspection of a range of pages if the process didn't have it mapped. I think we can add those extra vmaps without much trouble.

PR #961, opened 2019-08-20
PTE traversal update and mincore reimplementation

briankoco:
This PR externalizes an interface to walk page tables via src/x86_64/page.h:walk_ptes(), and uses
the interface to implement mincore. We also externalize various PTE helper functions in page.h

Also included are minor updates to the mincore portion of the mmap runtime test.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -192,15 +192,61 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void 
     return sysreturn_from_pointer(vnew);
 }
 
+static CLOSURE_3_3(mincore_fill_vec, boolean, u64, u64, u8 *, int, u64, u64 *);
+boolean mincore_fill_vec(u64 base, u64 nr_pgs, u8 * vec, int level, u64 addr, u64 * entry)
+{
+    u64 e = *entry;
+    u64 pgoff, i;
+
+    if (pt_entry_is_present(e)) {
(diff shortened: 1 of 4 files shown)


src/unix/mmap.c
```diff
@@ -192,15 +192,59 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
...
     }
-    return -ENOMEM;
+
+    return true;
+}
+
+static CLOSURE_0_1(mincore_vmap_gap, void, range);
+static void mincore_vmap_gap(range r)
+{
+    rprintf("mincore: found gap [0x%lx, 0x%lx)\n", r.start, r.end);
```
wjhun: This output should probably be debug-only. If you wish, you could modify rangemap_range_find_gaps() to accept an optional (maybe null) range_handler if no real callback on a gap is needed.

src/unix/mmap.c
```diff
@@ -192,15 +192,59 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
...
+    nr_pgs = length >> PAGELOG;
+
+    /* -ENOMEM if any unmapped gaps in range */
+    if (rangemap_range_find_gaps(
+            current->p->vmaps, 
+            (range){start, start + length},
+            stack_closure(mincore_vmap_gap)))
+        return -ENOMEM;
+
+    for (i = 0; i < nr_pgs; i++)
```
wjhun: could use runtime_memset() here

Issue #479, opened 2019-02-08
spring boot hello world


eyberg:
this is a pretty generic feature request - sorry

once we have java hello world working it'd be great to iron out anything to get spring/spring boot working - 
https://spring.io/guides/gs/spring-boot/

briankoco:
#973 should get us there

eyberg:
for anyone in the future that sees this 1.8 needs the tmp dir ->

```
{
  "Args": ["-Dos.name=Linux", "-jar", "/target/myproject-0.0.1-SNAPSHOT.jar"],
  "Dirs": ["target", "tmp"]
}
```

PR #964, opened 2019-09-03
SIGSEGV support. Includes:

briankoco:
- raise SIGSEGV from mmap fault handler
 - add ucontext save/restore to signal delivery/return code
 - update signal runtime test with SIGSEGV test

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -74,20 +74,14 @@ boolean unix_fault_page(u64 vaddr, context frame)
     u64 error_code = frame[FRAME_ERROR_CODE];
 
     vmap vm = (vmap)rangemap_lookup(p->vmaps, vaddr);
-    if (vm == INVALID_ADDRESS) {
-        if (error_code & FRAME_ERROR_PF_P) {
-            msg_err("page protection violation at 0x%lx: no vmap found\n", 
-                    vaddr);
-            return false;
-        }
 
+    /* no vmap --> send access violation */
+    if (vm == INVALID_ADDRESS) {
         deliver_segv(vaddr, SEGV_MAPERR); /* does not return */
         assert(0);
     }
 
-    /* we found a vmap -- must be a protection violation or 
-     * demand paging fault
-     */
+    /* vmap found, with protection violation set --> send prot violation */
     if (error_code & FRAME_ERROR_PF_P) {
         if (error_code & FRAME_ERROR_PF_RSV) {
             /* no SEGV on reserved PTEs */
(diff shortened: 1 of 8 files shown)


briankoco:
This gets java hello world to run on JDK 9, which relies on sigsegv to boot the JVM.
It also gets us part of the way to spring boot (#479) but there are other issues still

src/unix/system_structs.h
```diff
@@ -276,6 +276,24 @@ struct rlimit {
...
 #define SIGINFO_UNION_ALIGN 16
 
+/*
+ * SIGSEGV si_codes
+ */
+#define SEGV_MAPERR 1   /* address not mapped to object */
+#define SEGV_ACCERR 2   /* invalid permissions for mapped object */
+//#ifdef __bfin__
+//# define SEGV_STACKFLOW 3   /* stack overflow */
+//#else
```
wjhun: Cruft?

src/unix/system_structs.h
```diff
@@ -448,6 +468,113 @@ typedef struct {
...
+    u64 rbp;
+    u64 rbx;
+    u64 rdx;
+    u64 rax;
+    u64 rcx;
+    u64 rsp;
+    u64 rip;
+    u64 eflags; /* RFLAGS */
+    u16 cs;
+    u16 fs;
```
wjhun: Not sure what the correct source of this struct is, and we're just filling them with zero anyway, but the order of fs and gs is reversed here vs. linux arch/x86/include/uapi/asm/sigcontext.h.

The comments in that header are somewhat confusing - does Linux actually use these fields for anything meaningful?

src/unix/signal.c
```diff
@@ -298,18 +298,96 @@ static inline void reset_sigmask(thread t)
...
+    mcontext->eflags = f[FRAME_FLAGS];
+    mcontext->cs = f[FRAME_CS];
+    mcontext->fs = 0;
+    mcontext->gs = 0;
+    mcontext->ss = 0; /* FRAME[SS] if UC_SIGCONTEXT SS */
+    mcontext->err = f[FRAME_ERROR_CODE];
+    mcontext->trapno = f[FRAME_VECTOR];
+    mcontext->oldmask = sa->sa_mask.sig[0];
+
+    /* XXX: this should probably come from somewhere in the
```
wjhun: Yes, or feel free to add an entry in frame.h if it makes sense to do so.

src/unix/mmap.c
```diff
@@ -48,34 +65,39 @@ boolean unix_fault_page(u64 vaddr, context frame)
         return true;
     } else {
         /* page protection violation */
+        vmap vm = (vmap)rangemap_lookup(p->vmaps, vaddr);
```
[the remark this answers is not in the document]
briankoco: makes sense, updated

src/unix/system_structs.h
briankoco: Nice catch --- don't know how that happened ...

PR #968, opened 2019-09-18
Xen hypervisor support / network driver

wjhun:
This PR introduces support for running nanos on a guest HVM domain under a Xen hypervisor. A driver for the PV "xennet" device is included. With these changes, a nanos build with webg serves requests on both a stock, Debian-based Xen install and an EC2 T2 instance.

There are a lot of changes here, including general bug fixes and cleanup, copious header files imported from Xen sources (located under src/xen/public - licensed with an MIT-style license apart from Xen GPL sources), and reorganization of stage3 and platform-specific setup.

Note that there is preliminary support for Xen PV timers. However, due to unresolved issues on t2, it is presently unused. HPET is used in its place as a clocksource and runloop timer. HPET emulation is said to be inefficient, so this issue should be resolved before performance assessments are made.


src/x86_64/pvclock.c
```diff
@@ -0,0 +1,39 @@
+#include <runtime.h>
```
briankoco: Any reason why this is sitting in its own file rather than kvm_platform.c?

src/x86_64/pvclock.h
```diff
@@ -0,0 +1,20 @@
+struct pvclock_vcpu_time_info {
```
briankoco: Likewise for kvm_platform.h?

src/x86_64/kvm_platform.c
```diff
@@ -21,3 +39,56 @@ void halt(char *format, ...)
...
+    if (vc->system_time == 0) {
+        /* noise, but we should know if this happens */
+        msg_err("kvm pvclock probe failed: detected kvm pvclock, but system_time == 0\n");
+        return false;
+    }
+    init_pvclock(heap_general(kh), vc);
+    return true;
+}
+
+extern void lapic_runloop_timer(timestamp interval);
```
briankoco: For this and for other apic functions:
`configure_lapic_timer`
`lapic_eoi`
`init_apic`
 .... would it make sense to refactor into apic.h?

src/x86_64/service.c
```diff
@@ -237,27 +246,60 @@ static void __attribute__((noinline)) init_service_new_stack()
...
+    /* scheduling queues init */
     runqueue = allocate_queue(misc, 64);
     bhqueue = allocate_queue(misc, 2048); /* XXX will need something extensible really */
-    init_clock(kh);
+
+    /* interrupts */
+    init_debug("start_interrupts");
+    start_interrupts(kh);
+
+    /* platform detection and early init */
```
briankoco: Probably not something we need to worry about yet, but it's possible we'll run on instances with a combination of kvm and xen virtualization -- e.g., KVM as the core hypervisor doing CPU/EPT virtualization but also exposing xen PV devices. Just something to think about, in those cases we'll need to think about how to further refine our platform notion

src/x86_64/interrupt.c
```diff
@@ -330,65 +267,23 @@ void common_handler()
...
-
-void configure_lapic_timer(heap h)
+void register_interrupt(int vector, thunk t)
 {
-    apic_write(APIC_TMRDIV, 3 /* 16 */);
-    int v = allocate_u64(interrupt_vectors, 1);
-    apic_write(APIC_LVT_TMR, v); /* one shot */
-    register_interrupt(v, closure(h, int_ignore));
-    calibrate_lapic_timer();
+    handlers[vector] = t;
```
briankoco: may be helpful to have a debug check for erroneously overwriting assigned vectors ..

src/xen/public/arch-x86/cpuid.h
```diff
@@ -0,0 +1,118 @@
+/******************************************************************************
```
briankoco: fyi -- I did not look through all these headers. I assume it's all copy/paste

src/xen/xen.c
```diff
@@ -0,0 +1,847 @@
...
+    int rv = HYPERVISOR_set_timer_op(expiry);
+    if (rv != 0) {
+        msg_err("failed; rv %d\n", rv);
+    }
+}
+
+static CLOSURE_0_0(xen_runloop_timer_handler, void);
+static void xen_runloop_timer_handler(void)
+{
+    rprintf("%s: now %T\n", __func__, nanoseconds(pvclock_now_ns()));
```
briankoco: How often does this event channel fire? Should this rprintf be a debug?

src/xen/xen.c
```diff
@@ -0,0 +1,847 @@
...
+    xen_hvm_param.domid = DOMID_SELF;
+    xen_hvm_param.index = HVM_PARAM_STORE_PFN;
+    rv = HYPERVISOR_hvm_op(HVMOP_get_param, &xen_hvm_param);
+    if (rv < 0) {
+        msg_err("failed to get xenstore page address (rv %d)\n", rv);
+        return false;
+    }
+    xen_info.xenstore_paddr = xen_hvm_param.value << PAGELOG;
+
+    xen_debug("xenstore page at phys 0x%lx; allocating virtual page and mapping");
```
briankoco: missing address in this debug

src/xen/xen.c
```diff
@@ -0,0 +1,847 @@
...
+
+    /* timer setup */
+    xen_debug("stopping periodic tick timer...");
+    /* attempt to disable periodic (tick) timer; won't work in older Xens... */
+    rv = HYPERVISOR_vcpu_op(VCPUOP_stop_periodic_timer, 0 /* vcpu0 */, 0);
+    if (rv < 0) {
+        msg_err("unable to stop periodic timer (rv %d)\n", rv);
+        goto out_unregister_irq;
+    }
+// XXX - disabled due to incorrect behavior on t2
```
briankoco: in what way is the behavior incorrect?

src/xen/xen.c
```diff
@@ -0,0 +1,847 @@
...
+    status s = traverse_directory(xen_info.h, "device", &node);
+    if (!is_ok(s))
+        return s;
+    if (!node)
+        return timm("result", "failed to parse directory");
+
+    xen_info.device_tree = node;
+    xen_debug("success; result: %v", node);
+
+    table_foreach(node, k, v) {
```
briankoco: If I'm reading this correctly, the leaves here are never deallocated ...

src/xen/xennet.c
```diff
@@ -0,0 +1,778 @@
...
+    heap h;
+    heap contiguous;                /* physically */
+    int if_id;
+    domid_t backend_id;
+    buffer frontend;
+    buffer backend;
+
+    u8 mac[ETHARP_HWADDR_LEN];
+    u16 mtu;
+    netif_rx_front_ring_t rx_ring;
```
briankoco: I guess these rings do not also need to be manipulated via volatile pointers?

src/xen/xennet.c
```diff
@@ -0,0 +1,778 @@
...
+    xd->rx_ring_gntref = GRANT_INVALID;
+    xd->tx_ring_gntref = GRANT_INVALID;
+
+    /* allocate shared rings */
+    netif_rx_sring_t *rx_ring = allocate_zero(xd->contiguous, PAGESIZE);
+    netif_tx_sring_t *tx_ring = allocate_zero(xd->contiguous, PAGESIZE);
+    assert(rx_ring != INVALID_ADDRESS);
+    assert(tx_ring != INVALID_ADDRESS);
+    
+    SHARED_RING_INIT(rx_ring);
```
briankoco: Do these rings not also need to be manipulated via volatile pointers? I know this is public code, but just curious why the other hypervisor interfaces (in `xen_platform_info`) are declared volatile and these aren't

src/x86_64/pvclock.c
wjhun: Yes, they both use pvclock, so I split it out.

src/x86_64/kvm_platform.c
wjhun: Fixed.

src/x86_64/service.c
wjhun: That is true. That is why I didn't create static notions of platforms and left it kind of loose. But since we don't have any such instances to support, it's not clear how even probing such a platform would work. Presently both KVM and Xen are detected via inspecting cpuid leaves - in fact each has a signature on the same leaf node, if I'm not mistaken. So I think we should punt on this for now.

src/xen/public/arch-x86/cpuid.h
wjhun: Yes, they are imported (not all headers - only those required) verbatim from the xen tree (public headers only) with no changes.

src/x86_64/interrupt.c
wjhun: Fixed.

src/xen/xen.c
wjhun: Thanks, fixed.

src/xen/xen.c
wjhun: The event channel is fired by Xen when a timer expires. That should be a debug, but the code is still in flux and presently is not in use (see PR description).

src/xen/xen.c
wjhun: The PV timer appears to work fine on a local Xen instance, however on t2 I'm seeing that the event channel interrupt fires well before the timer expiry set (well beyond TIMER_SLOP as defined in Linux). Since we can use hpet for the time being, I decided to address this in a follow-up PR once I've debugged it.

src/xen/xen.c
wjhun: The intention here was that the device probe routine should consume the passed tuple entirely, but I'm sure I'm not doing that. I'll clean it up before merging.

src/xen/xennet.c
wjhun: This is a good question. In the case of these rings, I believe it's not necessary to use volatile because the macros which push the ring changes (e.g. RING_PUSH_REQUESTS_AND_CHECK_NOTIFY, RING_FINAL_CHECK_FOR_RESPONSES) use the various memory barrier macros, which in turn specify the "memory" clobber in their respective asm inlines. Memory barriers are also used prior to the ring processing loops as well. Neither Linux nor NetBSD uses volatiles for the rings. I suspect it would be advantageous for performance to avoid hitting memory with each read/write within these loops as well; the pattern here is to only update the ring pointers after each linear run of processing ring elements. The other shared page accesses mentioned above are akin to device register accesses, where it's safer to avoid stashing values in registers and instead always hit memory/cache with each access.

src/xen/xen.c
wjhun: Sadly, this depends on #974.

PR #969, opened 2019-09-19
only restore ucontext for sigactions with SA_SIGINFO in rt_sigreturn; align stack in setup_sigframe

wjhun:
An unconditional restore_ucontext in rt_sigreturn was causing a trashed thread frame when a ANSI/BSD-style signal handler (non-siginfo) was used. This led to the signal runtime test hanging in one case where the return to the (corrupted) thread frame was triggering an endless succession of page fault / SIGSEGVs. This change makes the restore conditional on the presence of SA_SIGINFO in the associated sigaction.

Incidentally, setup_sigframe should enforce a 16-byte alignment for the stack pointer before pushing the siginfo / ucontext and/or returning to the signal trampoline. It was, on some test setups, leading to an unaligned stack pointer and incorrect execution for e.g. 'movaps %xmm0, 0x50(%rsp)'.

### DIFF
--- a/src/unix/signal.c
+++ b/src/unix/signal.c
@@ -703,8 +703,8 @@ static void setup_sigframe(thread t, int signum, struct siginfo *si, void * ucon
         t->sigframe[FRAME_RSP] = 0; /* TODO */
         halt("SA_ONSTACK ...\n");
     } else {
-        /* must avoid redzone */
-        t->sigframe[FRAME_RSP] -= 128;
+        /* 16-byte alignment; avoid redzone */
+        t->sigframe[FRAME_RSP] = (t->sigframe[FRAME_RSP] & ~15) - 128;
     }
 
     /* arguments to trampoline */
--- a/src/unix/signal.c
+++ b/src/unix/signal.c
@@ -369,15 +369,18 @@ static void restore_ucontext(struct ucontext * uctx, context f)
 sysreturn rt_sigreturn(void)
 {
     struct rt_sigframe * frame;
+    sigaction sa;
     thread t = current;
 
     /* sigframe sits at top of the stack */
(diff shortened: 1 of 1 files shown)


briankoco:
Looks good, nice catch.

One other thing we should pay attention to is SA_RESTORER, which indicates that userspace has a signal trampoline to be invoked on delivery. I guess we should probably use this instead of our own, in case glibc does something like cache a value in the TLS.

I have this change in my local branch which seems to do the trick, though I haven't done extensive testing. 
"""
    /* must avoid redzone */
    t->sigframe[FRAME_RSP] -= 128;
    
    /* create space for rt_sigframe */
    t->sigframe[FRAME_RSP] -= pad(sizeof(struct rt_sigframe), 16);
    
    /* setup sigframe for user sig trampoline */
    frame = (struct rt_sigframe *)t->sigframe[FRAME_RSP];
    frame->pretcode = sa->sa_restorer;
    runtime_memcpy(&(frame->info), si, sizeof(struct siginfo));
    setup_ucontext(&(frame->uc), sa, si, t->frame);

   /* setup regs for signal handler */
    t->sigframe[FRAME_RIP] = u64_from_pointer(sa->sa_handler);
    t->sigframe[FRAME_RDI] = signum;

    if (sa->sa_flags & SA_SIGINFO) {
        t->sigframe[FRAME_RSI] = (u64)&(frame->info);
        t->sigframe[FRAME_RDX]  = (u64)&(frame->uc);
    } else {
        t->sigframe[FRAME_RSI] = 0;
        t->sigframe[FRAME_RDX] = 0;
    }
"""

wjhun:
I agree that we should be supporting the user trampoline. I'll merge this and follow up with another PR for that. Thanks!

PR #972, opened 2019-09-19
Support for SA_RESTORER and glibc signal trampoline

wjhun:
This pulls in @briankoco's change from #969:

https://github.com/nanovms/nanos/pull/969#issuecomment-533066602

This obviates the in-kernel trampoline in favor of the user trampoline set in the sigaction's sa_restorer field. Mandatory for x86-64, this support insures that the correct trampoline (in glibc) is used on handler exit, which is necessary for proper gdb backtraces through signal handlers.

### DIFF
--- a/src/unix/signal.c
+++ b/src/unix/signal.c
@@ -368,12 +368,12 @@ static void restore_ucontext(struct ucontext * uctx, context f)
 
 sysreturn rt_sigreturn(void)
 {
-    struct rt_sigframe * frame;
+    struct rt_sigframe *frame;
     sigaction sa;
     thread t = current;
 
-    /* sigframe sits at top of the stack */
-    frame = (struct rt_sigframe *)(t->sigframe[FRAME_RSP]);
+    /* sigframe sits at %rsp minus the return address word (pretcode) */
+    frame = (struct rt_sigframe *)(t->sigframe[FRAME_RSP] - sizeof(u64));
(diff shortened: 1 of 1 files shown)


src/unix/signal.c
```diff
@@ -693,45 +668,38 @@ static void setup_sigframe(thread t, int signum, struct siginfo *si, void * ucon
...
-        t->sigframe[FRAME_RDX] = t->sigframe[FRAME_RSP];
-
-        t->sigframe[FRAME_RSP] -= pad(sizeof(struct ucontext), 16);
-        dest_uctx = pointer_from_u64(t->sigframe[FRAME_RSP]);
-        sig_debug("copying ucontext to [%p, %p)\n", dest_uctx, ((void *)dest_uctx) + sizeof(struct ucontext));
-        setup_ucontext(dest_uctx, sa, si, t->frame);
-        t->sigframe[FRAME_R8] = t->sigframe[FRAME_RSP];
+        runtime_memcpy(&(frame->info), si, sizeof(struct siginfo));
+        setup_ucontext(&(frame->uc), sa, si, t->frame);
+        t->sigframe[FRAME_RSI] = (u64)&(frame->info);
```
briankoco: I guess for consistency you could replace these with u64_from_pointer() ...

briankoco: I guess for the sake of consistency you could replace with u64_from_pointer() ...

PR #973, opened 2019-09-21
Series of bugfixes to enable spring boot

briankoco:


briankoco:
err, sorry for the messy reversions. Basically, I just pulled out 48dd6ce9fe1225c4ba14a604528ca8859a10bf3a  which erroneously modified the tfs mkentry function.

src/unix/thread.c
```diff
@@ -2,6 +2,11 @@
 
 thread current;
 
+struct futex {
```
wjhun: This looks like a major improvement to the futex code. Would it be a pain to split it out into a futex.c? It's substantial enough and doesn't seem like something inseparably tied to the thread code.

PR #979, opened 2019-09-26
xennet: handle EAGAIN temporary failure when posting transaction to backend

wjhun:
An XS_TRANSACTION_END request may return an XS_ERROR with a value of EAGAIN, indicating that the transaction should be retried. This amends xennet_inform_backend to retry some number of times after receiving EAGAIN.

### DIFF
--- a/src/xen/xennet.c
+++ b/src/xen/xennet.c
@@ -99,60 +99,74 @@ struct xennet_dev {
     struct list tx_free;
 };
 
+#define XENNET_INFORM_BACKEND_RETRIES 128
+
 static status xennet_inform_backend(xennet_dev xd)
 {
     status s = STATUS_OK;
     xennet_debug("%s: dev id %d", __func__, xd->if_id);
 
     u32 tx_id;
-    boolean again = false;
     char *node;
-    do {
-        s = xenstore_transaction_start(&tx_id);
-        if (!is_ok(s))
-            return s;
+    int retries = XENNET_INFORM_BACKEND_RETRIES;
 
-        node = "vifname";
-        s = xenstore_sync_printf(tx_id, xd->frontend, node, "vif%d", xd->if_id);
-        if (!is_ok(s))
-            goto abort;
+  again:
+    node = "transaction start";
+    s = xenstore_transaction_start(&tx_id);
+    if (!is_ok(s))
+        return s;
 
-        node = "rx-ring-ref";
-        s = xenstore_sync_printf(tx_id, xd->frontend, node, "%d", xd->rx_ring_gntref);
-        if (!is_ok(s))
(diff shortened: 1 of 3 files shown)


src/xen/xennet.c
```diff
@@ -99,60 +99,74 @@ struct xennet_dev {
...
+        value v = table_find(s, sym(errno));
+        if (v) {
+            if (!runtime_strcmp("EAGAIN", buffer_ref((buffer)v, 0))) {
+                deallocate_tuple(s);
+                if (retries-- == 0) {
+                    s = timm("result", "%s failed: transaction end returned EAGAIN after %d tries",
+                             __func__, XENNET_INFORM_BACKEND_RETRIES);
+                    goto abort;
+                }
+                goto again;
```
briankoco: I don't really know anything about this interface -- but is there any merit in pausing for a bit before retrying? What causes the EAGAINs?

src/xen/xennet.c
```diff
@@ -99,60 +99,74 @@ struct xennet_dev {
...
+            }
         }
-    } while (again);
+        goto abort;
```
briankoco: I think you're leaking the tuple here

src/xen/xennet.c
wjhun: It's not clear to me if there is merit in pausing. Neither NetBSD nor Linux implementations pause before retrying.

Here the xennet driver is attempting to issue a xenstore "transaction" to the xenstore daemon. A transaction consists of multiple asynchronous (although since we're only using the xenstore for init, we're polling) writes to the store, delineated by transaction beginning and end messages. The xenstore daemon keeps a generation count of the database (or maybe some portion of it, I'm not sure). If a transaction begins and the generation count advances before the transaction end occurs, the transaction is aborted and an XS_ERROR message with a response of "EAGAIN" is sent to indicate to the caller that the transaction should be retried.

I suppose if there's a lot of contention on writes to the store then some kind of backoff delay might be in order, but I'm satisfied to play follow-the-leader here.

src/xen/xennet.c
wjhun: Thanks, that error handling was messed up. I actually took the opportunity to sneak in "timm_up," which will allow us to trivially build up chains of statuses after a failure (that roughly follow the call stack). More on that later - we can do a lot to take advantage of these status tuples.

Deallocating tuples in general requires a bit more work to make it so that any buffer (wrapped / not wrapped, shared data, etc.) can be safely deallocated. Maybe also loop detection would be in order, as there's nothing to prevent cycles from being formed.

src/xen/xennet.c
briankoco: ok gotcha

PR #984, opened 2019-10-01
closure deallocation roundup

wjhun:
Self-explanatory follow up to #981. Tuple parser updates will be addressed separately.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -371,7 +371,10 @@ closure_function(7, 2, sysreturn, sock_read_bh,
                  sock, s, thread, t, void *, dest, u64, length, struct sockaddr *, src_addr, socklen_t *, addrlen, io_completion, completion,
                  boolean, blocked, boolean, nullify)
 {
-    return sock_read_bh_internal(bound(s), bound(t), bound(dest), bound(length), bound(src_addr), bound(addrlen), bound(completion), blocked, nullify);
+    sysreturn rv = sock_read_bh_internal(bound(s), bound(t), bound(dest), bound(length), bound(src_addr), bound(addrlen), bound(completion), blocked, nullify);
+    if (rv != infinity)
+        closure_finish();
+    return rv;
 }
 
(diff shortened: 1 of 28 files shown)


wjhun:
BTW, the memory savings here are huge. In a master tree, if I serve 5,000 requests with webg, the general heap occupancy shoots up from 1.2M at init state to 12.5M after serving all requests. With the closure deallocations, the same operation causes a rise from 1.2M to 2.2M.

Undoubtedly there are more leaks to address, but cleaning up closure allocations has already brought the waste down an order of magnitude for this test case.

wjhun:
Given that closures are presently completely general with respect to arguments and return type - and thus return values aren't necessarily integers - I can't really see adding an encoding the way other very specific interfaces in the kernel do (e.g. blockq).

In all the cases I could think of and find in the system, it made sense to me to have the "producer" side (side creating/allocating the closure, versus the applier) also handle the deallocation (or ignore deallocating if it is known that the closure is only ever a stack closure or a closure allocated on a transient heap). In those cases, it was readily knowable on the producer side when a closure would no longer be used - but not necessarily on the consumer side. This is also why I changed the notify_remove a little bit; I preferred to have it optionally direct the consumer side (notify) to apply its stored closure with the 'release' flag to trigger the freeing of that handler within the closure rather than have the consumer try to deallocate it directly. I couldn't see why the consumer would need to know anything about - or perform some kind of maintenance on behalf of - a foreign closure object that got sent down the pike.

Keep in mind too that it's entirely up to the author to use this approach or, say, never free on the producer side and use "deallocate_closure" on the apply side. So even though I only ran into cases that were easy to resolve on the producer side, I think there is nothing preventing consumer side deallocating too. And, though not in some canonical way, the producer can still encode a return value to indicate to the consumer that a deallocate or whatever kind of release is in order.

If there is ever a need to pass more return values, I suppose they could spill into the closure environment. The applier won't have the specific type of the closure necessarily (just the closure_type with return value and rhs types), but we could say that such-and-such return values of fixed types are invariants (i.e. in _closure_common), or unpack them in some other way.

PR #985, opened 2019-10-01
remap_pages() bugfix

briankoco:
bugfix: remap_pages() can overwrite PTEs beyond the range of the target destination address range when it encounters missing PTE entries. This leads to occasional failures in the mmap runtime test, or anything using mremap

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -160,7 +160,6 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void 
         new_size == 0)
         return -EINVAL;
 
-    /* XXX should determine if we're extending a virtual32 allocation... */
     heap vh = p->virtual_page;
     heap physical = heap_physical(kh);
     heap pages = heap_pages(kh);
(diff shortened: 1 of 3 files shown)


src/unix/mmap.c
```diff
@@ -169,6 +168,33 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
     if (new_size <= old_size)
         return sysreturn_from_pointer(old_address);
 
+    /* verify we have a single vmap for the old address range */
+    vmap old_vm = (vmap)rangemap_lookup(p->vmaps, old_addr);
```
wjhun: I suppose there could be adjacent vmaps (e.g. an anonymous one and a fixed one immediately after) that are both affected by a remap. But we can track this with an issue and circle around...this should at least bail out sufficiently if we run into such a case.

src/unix/mmap.c
```diff
@@ -169,6 +168,33 @@ sysreturn mremap(void *old_address, u64 old_size, u64 new_size, int flags, void
...
+        return -EFAULT;
+
+    /* XXX should determine if we're extending a virtual32 allocation...
+     * - for now only let the user move anon mmaps 
+     */
+    if (! ((old_vm->flags & VMAP_FLAG_MMAP) &&
+           (old_vm->flags & VMAP_FLAG_ANONYMOUS))
+       )
+    {
+        msg_err("mremap only supports anon mmap regions at the moment\n");
```
wjhun: Would this not work for mappings backed by a file? At least at this time all the pages should be present for a backed mapping.

src/unix/mmap.c
briankoco: From `man 2 mremap:`

"EFAULT "Segmentation  fault."  Some  address  in the range old_address to old_address+old_size is an invalid virtual memory address for this process.  You can also get EFAULT even if there exist mappings that cover the whole address space requested, but those mappings are of different types."

So it seems like the adjacent maps issued of different types can safely fail. As to the other issues, yeah, if we see a need for them we can add them, shouldn't be too difficult.

src/unix/mmap.c
briankoco: Yeah I don't see why not.  As soon as we see something that needs it, should just be a one line fix

Issue #456, opened 2019-01-25
do the right thing on thread exit


wjhun:
The system still has a number of carryovers from the original "don't ever free anything" approach that was taken during initial bringup. In exit_thread(), we should be cancelling blocking operations, freeing futexes and tearing down mappings. We're not doing any of this at the moment.

wjhun:
Resolved in #990

PR #990, opened 2019-10-08
thread exit cleanup

wjhun:
This contains a variety of clean up around unix land to allow us to free up structures on thread exit. Mostly this consists of straightening out refcounts in various structures (facilitated by adding a refcount interface - a small improvement over fetch_and_add) and finishing cleanup for various types of objects (particularly with poll waiting). Note that yet more additions have been made to closures - specifically the ability to embed a closure struct within another struct, which I used for the various refcount completions.

See #456

### DIFF
--- a/src/unix/blockq.c
+++ b/src/unix/blockq.c
@@ -107,6 +107,7 @@ static void blockq_item_finish(blockq bq, blockq_item bi)
         /* XXX acquire spinlock */
     }
 
+    refcount_release(&bi->t->refcount);
     free_blockq_item(bq, bi);
 }
 
@@ -227,6 +228,7 @@ sysreturn blockq_check_timeout(blockq bq, thread t, blockq_action a,
 
     bi->a = a;
     bi->t = t;
+    refcount_reserve(&t->refcount);
 
     if (timeout > 0) {
         bi->timeout = register_timer(timeout,
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -6,14 +6,18 @@
 #define epoll_debug(x, ...)
 #endif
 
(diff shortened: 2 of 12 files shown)


src/unix/blockq.c
```diff
@@ -227,6 +228,7 @@ sysreturn blockq_check_timeout(blockq bq, thread t, blockq_action a,
 
     bi->a = a;
     bi->t = t;
+    refcount_reserve(&t->refcount);
```
briankoco: Should probably be released if the call to register_timer() fails a couple of lines below

src/unix/poll.c
```diff
@@ -356,9 +399,14 @@ static epoll_blocked alloc_epoll_blocked(epoll e)
     epoll_blocked w = unix_cache_alloc(get_unix_heaps(), epoll_blocked);
     if (w == INVALID_ADDRESS)
 	return w;
-    w->refcnt = 1;
+    epoll_debug("w %p\n", w);
+
+    /* initial reservation released on thread wakeup (or direct return) */
+    init_refcount(&w->refcount, fill_closure(&w->free, epoll_blocked_free, w));
     w->t = current;
```
briankoco: This should be taking a ref on current, right?

src/unix/blockq.c
wjhun: Thanks; I added a check for register_timer failure.

src/unix/poll.c
```diff
@@ -6,14 +6,18 @@
 #define epoll_debug(x, ...)
```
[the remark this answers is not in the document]
wjhun: Yes, I intend to add them here and in a few other places in a follow-up PR.

PR #993, opened 2019-10-17
convert uninterruptible waits to blockq where possible

wjhun:
This rounds out conversion of the remaining cases of uninterruptible blocks that should instead be on a blockq, including the poll waiters (epoll_wait, *select, *poll), nanosleep(2) and TCP connect. The only uninterruptible blocking that remains are filesystem I/O and file-backed mmaps - which are uninterruptible by design.

The blockq actions now take a 'timedout' flag to indicate timer expiry so that the blocking actions don't need to unnecessarily store timestamps and manually check for expiry.

nanosleep(2) implementation should now be complete, and signal interruption will cause the remaining time to be stored in *rem.

The socket code will attempt to abort a TCP connection if a signal was received before lwIP reported connection complete. This could use an explicit test.

Some more exhaustive testing of the poll waiters is in order to verify correct behavior on interruption, so I may either tack on another runtime test(s) before merging or follow up with another PR of tests. A unit test for interrupting nanosleep(2) is in order as well.

### DIFF
--- a/contgen/contgen.c
+++ b/contgen/contgen.c
@@ -86,11 +86,13 @@ void cblock()
 
     p("#define CLOSURE_DEFINE_%_%(_rettype, _name^~)|", nleft, nright, ", _lt%, _ln%", ", _rt%, _rn%");
     p("static _rettype (**_fill_##_name(heap h, struct _closure_##_name* n, bytes s^))(void *~) {|", ", _lt% l%", ", _rt%");
-    p("  n->__apply = _name;|");
-    p("  n->__c.name = #_name;|");
-    p("  n->__c.h = h;|");
-    p("  n->__c.size = s;|");
+    p("  if (n != INVALID_ADDRESS) {|");
+    p("    n->__apply = _name;|");
+    p("    n->__c.name = #_name;|");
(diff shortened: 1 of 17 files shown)


briankoco:
Looks good -- my only comment is that, for the blockq closure applications, we could use a single status flags variable as a bitwise-OR of the various conditions (nullify/blocked/etc.), rather than having a separate parameter for each. This might increase our chances of having the args passed in registers rather than on the stack, and also seems a little cleaner and more easily extensible to me.

wjhun:
> we could use a single status flags variable as a bitwise-OR of the various conditions (nullify/blocked/etc.)

Agreed and fixed. Other changes:

- The return value type from closure fill was being clobbered in the closure macro, preventing compiler errors that should occur when an allocated closure is assigned to a variable of a mismatched closure type. Move the closure allocation check to the fill function (yes, every fill function via contgen - but who cares?).
- Using "infinity" (-1ull) as an indicator of blocking required in the blockq_actions clobbers a legitimate return of -EPERM; instead use BLOCKQ_BLOCK_REQUIRED and define it as a value that can't possibly coincide with an errno (i.e. is outside the range of int).

PR #994, opened 2019-10-23
ftrace 

briankoco:
Introduce basic ftrace functionality
 - This relies on the "-pg" flag to build the kernel with tracing. All kernel functions are interposed with a call to mcount(), which is responsible for saving the running register context and collecting information about the function being invoked and storing in a ringbuffer.
 - Currently, a user process has to set itself to be traced and then read the /sys/ files to collect tracing information. A future PR will provide external tracing abilities

### DIFF
--- a/src/unix/ftrace.c
+++ b/src/unix/ftrace.c
@@ -1,0 +1,1103 @@
+#include <runtime.h>
+#include <unix_internal.h>
+#include <ftrace.h>
+#include <x86_64.h>
+#include <symtab.h>
+
+#ifdef CONFIG_FTRACE
+
+/* 1MB default size for the user's trace array */
+#define DEFAULT_TRACE_ARRAY_SIZE        (1ULL << 20)
+#define DEFAULT_TRACE_ARRAY_SIZE_KB     (DEFAULT_TRACE_ARRAY_SIZE >> 10)
+
+#define TRACE_TASK_WIDTH                15
+#define TRACE_PID_WIDTH                 5
+
+/* prevent buffers from growing to unbounded sizes */
+#define TRACE_PRINTER_INIT_SIZE         (1ULL << 12) /* 4KB */
+#define TRACE_PRINTER_FLUSH_SIZE        (1ULL << 18) /* 256KB */
+
+static heap ftrace_heap;
+
+/* whether or not to write into the ring buffer */
+static boolean tracing_on = false;
+
+struct rbuf_entry {
+    unsigned long ip;
+    unsigned long parent_ip;
+    unsigned long tsc;
+    unsigned short cpu;
+    int tid;
+    char name[15]; /* unfortunate but safer than accessing a thread pointer */
+};
+
+struct rbuf {
+    struct rbuf_entry * trace_array;
+    unsigned long count; /* current number of unconsumed items */
+    unsigned long total_written; /* total items ever written */
+    unsigned long max_entries; /* size of buffer */
+    unsigned long read_idx;
+    unsigned long write_idx;
+
+    unsigned long disable_cnt;
+    unsigned long wait_cnt;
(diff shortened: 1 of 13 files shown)


src/unix/ftrace.c
```diff
@@ -0,0 +1,1103 @@
+#include <runtime.h>
+#include <unix_internal.h>
+#include <ftrace.h>
+#include <x86_64.h>
+#include <symtab.h>
+
+#ifdef CONFIG_FTRACE
```
wjhun: Maybe conditional compile rather than #ifdef for the entire file?

src/unix/ftrace.c
```diff
@@ -0,0 +1,1103 @@
...
+    }
+
+    rv = do_trace_pipe_read(
+        bound(p), rbuf, bound(buf), bound(length), bound(offset)
+    );
+
+    if (rv == 0) {
+        if (!blocked)
+            rbuf_wait(rbuf);
+        rv = infinity;
```
wjhun: Will need to change this to BLOCKQ_BLOCK_REQUIRED

src/unix/ftrace.c
```diff
@@ -0,0 +1,1103 @@
...
+
+    unsigned long disable_cnt;
+    unsigned long wait_cnt;
+    timer trace_timer;
+    blockq bq;
+};
+
+/* This structure is designed to simply the process of efficiently
+ * writing buffers to userspace when the user issues multiple reads
+ * with changing offsets
```
wjhun: I think we can sort of safely assume that there won't be any reading of this ftrace data from userspace - unless we want to consider the case of having a diagnostic user program that happens to inspect ftrace data? Seems doubtful for the short-term anyway.

More likely is that buffers of data will be shipped to be sent out remotely, as in the case of the http and/or direct socket interface. In this case, said buffers will sit on a queue waiting to transmitted, and offsets within are tracked using buffer_consume (i.e. b->start). Of course, this is destructive, so any read of persistent data will need to do a clone_buffer or something similar. However this is all fine to merge as-is; we can make adjustments as we wire up the remote interface(s). And we may still want that kernel diagnostic userland test down the road (e.g. a performance / memory regression test).

src/x86_64/ftrace.s
```diff
@@ -0,0 +1,98 @@
...
+;;  [ tracee retaddr in parent ]  <--- moved into rsi 
+%macro save_mcount_regs 0
+    push rbp
+    push qword [rsp+8]
+    push rbp
+    mov rbp, rsp
+ 
+    ;; we add enough stack to save all the regs, but only need those
+    ;; needed and potentially clobbered by mcount
+    sub rsp, (SAVE_REG_SIZE - SAVE_FRAME_SIZE)
```
wjhun: I haven't worked through all this, but can we be assured that rsp always stays 16-byte aligned?

stage3/Makefile
```diff
@@ -40,6 +40,7 @@ SRCS-stage3.img= \
 	$(SRCDIR)/unix/blockq.c \
 	$(SRCDIR)/unix/exec.c \
 	$(SRCDIR)/unix/eventfd.c \
+	$(SRCDIR)/unix/ftrace.c \
```
wjhun: Again maybe make these conditional on CONFIG_FTRACE?

src/unix/ftrace.c
briankoco: Yeah, this was all done before realizing that we didn't really need a userland interface. I do think the ftrace_printer stuff might be worth separating into a standalone interface at some point, if we have an application that needs to read large virtual files. So I'll leave it as is for the time being.

src/x86_64/ftrace.s
briankoco: good call, thanks

PR #996, opened 2019-10-24
in-kernel tcp connection / http interface

wjhun:
This introduces net/direct.c for handling TCP connections within the kernel. http handling code is moved from src/unix_process to src/http and also built into the kernel. As a (temporary) test, stage3.c contains optional telnet and http test drivers (enabled by setting telnet and http flags in manifest, respectively). We can delete these once we've wired up ftrace and other debug services.

A few general changes are also included:
- A "deferqueue" has been added to service.c to allow bottom half code to re-schedule thunks for execution in the following call to process_bhqueue. This is used, for instance, to allow queued data in a direct_conn to spool out to lwIP as space becomes available in the TCP send window. It also becomes possible to approximate a kernel thread by continuously re-scheduling a run method via 'enqueue(deferqueue, closure_self())'.
- buffer_handlers now return a status. Status handling in the connection code isn't thoroughly sorted out, but this is a start.

Obviously, this connection interface is meant to be used for debugging, with ftrace control/trace download and tuple r/w in mind. As such, it is quite rudimentary and designed for handling request-response transactions by exchanging buffers (using buffer_handlers on either end). We can build this out as needed, e.g. passing 'from' sockaddr info to connection_handlers, etc., but don't expect the flexibility of unix sockets here...

### DIFF
--- a/src/x86_64/service.c
+++ b/src/x86_64/service.c
@@ -58,8 +58,9 @@ static u64 bootstrap_alloc(heap h, bytes length)
     return result;
 }
 
-queue runqueue;
-queue bhqueue;
+queue runqueue;                 /* dispatched in runloop */
+queue bhqueue;                  /* dispatched to exhaustion in process_bhqueue */
+queue deferqueue;               /* same as bhqueue, but only for previously queued items */
 
 static void timer_update(void)
 {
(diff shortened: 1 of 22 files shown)


src/x86_64/service.c
```diff
@@ -74,10 +75,19 @@ void process_bhqueue()
 {
     /* XXX - we're on bh frame & stack; re-enable ints here */
     thunk t;
-    while((t = dequeue(bhqueue))) {
```
briankoco: Does it make more sense to process the deferqueue first, so that if it generates work on the bhqueue we'll be able to process it? This would of course depend on how we decide to use the two queues, so it's not clear that such a scenario will happen, but it might be important for making sure that any time-sensitive work that is spawned off of the deferqueue is processed quickly

wjhun: So the idea was that the bhqueue is used for high priority events triggered by interrupt handlers and other exceptions. The deferqueue is for when a condition requires blocking (deferred execution - e.g. deferring the sending of output data for a connection when the TCP send window is full). It's more likely that an event triggered by the bhqueue will queue up work on the deferqueue rather than the other way around.

PR #1000, opened 2019-10-30
Support for querying trace data via http

briankoco:
This PR adds support for querying ftrace data. A couple of notes
- This disables the stage3 debug HTTP interface, since we also use port 9090. It would be trivial to move to a different port to leave the debug interface untouched if we want
 - Only http GET is supported currently. A future PR will be needed to support PUT/POST
 - Only supports transfers of a single buffer per http GET request, which means data transfer is limited to ~1MB. Adding support for chunked transfers (thanks to #999) should resolve this


wjhun:
FYI, I'm getting errors caused by unused static functions when building with TRACE=ftrace:

```
/Users/wjhun/src/nanos/ftrace/src/unix/ftrace.c:323:1: error: unused function 'rbuf_wait' [-Werror,-Wunused-function]
rbuf_wait(struct rbuf * rbuf)
^
/Users/wjhun/src/nanos/ftrace/src/unix/ftrace.c:329:1: error: unused function 'rbuf_release' [-Werror,-Wunused-function]
rbuf_release(struct rbuf * rbuf)
^
/Users/wjhun/src/nanos/ftrace/src/unix/ftrace.c:336:1: error: unused function 'rbuf_has_waiters' [-Werror,-Wunused-function]
rbuf_has_waiters(struct rbuf * rbuf)
^
3 errors generated.
make[1]: *** [/Users/wjhun/src/nanos/ftrace/output/stage3/src/unix/ftrace.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [stage3] Error 2
```

PR #1001, opened 2019-11-04
ftrace: send chunked http response

wjhun:
ftrace printer semantics change somewhat: A local rbuf index is set to read_idx on rbuf init and incremented with each print. If a printer returns true, indicating more data to print, another chunk response is scheduled in the future via a runloop timer. False indicates that printing of the rbuf has completed (whether destructive or not), upon which point the http response is finished and rbuf and printers are reset.

One issue is that queued, asynchronous processing of output buffers gets traced after issuing of output has completed on the ftrace side - thus adding a lot of noise to subsequent trace gets. This should only be a concern for trace_pipe, and can be ameliorated by adding an optional completion (status_handler) to buffer_handlers and then tying buffer responses together with a merge. The merge completion would then perform the rbuf deinit, thus re-enabling tracing.

### DIFF
--- a/src/unix/ftrace.c
+++ b/src/unix/ftrace.c
@@ -45,7 +45,7 @@ struct rbuf {
     unsigned long disable_cnt;
 };
 
-/* This structure is designed to simply the process of efficiently flushing
+/* This structure is designed to simplify the process of efficiently flushing
  * buffers to userspace/http response handlers
  *
  * This is also needed to persist rbuf contents across multiple destructive
@@ -54,8 +54,12 @@ struct rbuf {
  * least that half line of data in the buffer b for the likely subsequent call
  * to read with an incremented offset
  */
+
+typedef closure_type(printer_flush, status, boolean);
+
 struct ftrace_printer {
     buffer b;
+    printer_flush flush;
     u64 local_offset;
     u64 max_size; /* stop printing if the buffer reaches this size */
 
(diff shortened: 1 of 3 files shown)


src/unix/ftrace.c
```diff
@@ -1211,8 +1239,69 @@ ftrace_send_http_server_error(buffer_handler handler)
...
+    }
+
+    /* ret > 0, so more to send */
+    return true;
+  send_http_chunk_failed:
+    msg_err("%s: send_http_chunk failed with %v\n", __func__, s);
+    return false;
+}
+
+#define SEND_HTTP_CHUNK_INTERVAL_MS     (milliseconds(75))
```
briankoco: Can you explain the reason for this timer here? Why do we need one and why is 75 ms appropriate?

You mention simultaneous requests below, not really sure what that means. Are you saying there are internal
data structures in the http response code that could be corrupted by sending a bunch of chunks one after another?

src/unix/ftrace.c
```diff
@@ -6,7 +6,7 @@
 #include <http.h>
 
 /* 1MB default size for the user's trace array */
```
briankoco: maybe update this comment to reflect the new size

src/unix/ftrace.c
wjhun: It is kind of kludgey, but I wanted to provide some kind of back-off for the printing and spooling of output buffers. Just running continuously off of the bhqueue or runqueue starves out the draining of queued buffers into lwIP TCP for long periods, leading to a very slow transfer rate until everything is printed. 75ms was selected empirically.

I think we could avoid this kludge once we have the completions in buffer_handlers. But maybe it's ok for now?

As for the comment about simultaneous http requests, just disregard. I didn't realize that it was made exclusive in the inits anyway.

wjhun:
Ok, maybe we can address the printer offset in another branch. It does (I think?) make sense to have a byte offset for supporting the file accesses in case a read boundary falls in the middle of a printed item. When the basic functionality and function_graph stuff is folded in we could circle around and fix up the file access operations if need be.

PR #1002, opened 2019-11-05
function_graph support for ftrace

briankoco:
Reworks some of the printer setup. Each tracer (function and function_graph) now implement callbacks for both printing headers and entries. These both return void. The functionality added in a previous PR (#1001) whereby a boolean is returned to indicate non empty rbuf is still used, but is pulled out into ftrace_print_rbuf(), which invokes the appropriate tracer-specific callbacks.

The rbuf_entry data structure is reworked to support the different tracer requirements.

This also adds a new data structure to the thread struct: a call stack which is needed to track original function return addresses, which the function_graph tracer needs to recover the return addresses it 
overwrites.

Beyond that, most everything else is specific to function_graph code within ftrace.c and ftrace.s


src/unix/ftrace.c
```diff
@@ -459,47 +575,200 @@ function_print_entry(struct ftrace_printer * p, struct rbuf_entry * entry)
...
+
+    /* check for special switch event */
+    if (graph->depth == TRACE_GRAPH_SWITCH_DEPTH) {
+        print_switch_event(p, &(entry->sw));
+        return;
+    }
+
+    printer_write(p, " %d) ", graph->cpu);
+
+    /* XXX symbol */
```
wjhun: What is this in reference to?

src/unix/ftrace.c
```diff
@@ -1475,3 +1768,176 @@ ftrace_deinit(void)
...
+ * self  : IP in mcache_alloc() where mcount is invoked
+ * parent: stack address where return address sits (from mcache_alloc
+ *      back into init_unix; i.e., *parent is an address in init_unix)
+ * frame_pointer: frame_pointer during call to mcache_alloc()
+ *      - i.e., *(frame_pointer+8) is the return address to get back
+ *       from init_unix() to its own parent (in this case, startup())
+ *
+ * What this function does is moves *parent so that mcache_alloc() returns
+ * to a stub of ours that can record the time, before jumping back into
+ * init_unix()
```
wjhun: Could there be an issue here with tail call elimination? Or will the compiler rule them out when -pg is specified?

src/x86_64/ftrace.s
```diff
@@ -86,16 +93,63 @@ extern __current_ftrace_graph_return
 
 global ftrace_stub
 global mcount
+global return_to_handler
+
 mcount:
-    cmp qword [__current_ftrace_trace_fn], ftrace_stub
-    jnz trace
+    cmp qword [__ftrace_function_fn], ftrace_stub
```
wjhun: Could these just be function vectors that are called unconditionally?

src/unix/ftrace.h
```diff
@@ -95,6 +98,24 @@ static inline void
...
+{
+    return 0;
+}
+
+static inline void
+ftrace_thread_deinit(thread t)
+{}
+
+static inline void
+ftrace_thread_switch(thread old, thread in)
```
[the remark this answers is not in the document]
briankoco: This is just used to record an event in the rbuf when a context switch happens. If you look at a function_graph trace file, you'll see something like
----------------------------------------
    tid-0  >=  tid-1
----------------------------------------

src/unix/ftrace.c
```diff
@@ -15,25 +15,60 @@
 /* 4KB; will grow dynamically */
 #define TRACE_PRINTER_INIT_SIZE (1ULL << 12)
 
-/* MAX size: larger than 1MB buffers not currently supported on general heap */
-#define TRACE_PRINTER_MAX_SIZE  (1ULL << 19)
+/* Larger than 1MB buffers not currently supported on general heap */
+#define TRACE_PRINTER_MAX_SIZE  (1ULL << 18)
```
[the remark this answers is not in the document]
briankoco: sorry, didn't mean to push that in

src/unix/ftrace.c
```diff
@@ -322,8 +438,10 @@ rbuf_enabled(struct rbuf * rbuf)
 static inline boolean
 __rbuf_acquire_write_entry(struct rbuf * rbuf, struct rbuf_entry ** acquired)
 {
-    if (rbuf->count == rbuf->size - 1)
+    if (rbuf->count == rbuf->size - 1) {
```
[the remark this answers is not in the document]
briankoco: just debug cruft -- removed

src/unix/ftrace.c
briankoco: linux adds various symbols that correspond to latency ranges, making it easier to search for long running functions. I'll add it in, not a big effort

src/x86_64/ftrace.s
briankoco: Yeah we could do that. The idea here though is that doing three compares is quicker than doing function calls that just immediately ret, which will issue a few extra instructions to manipulate the stack and frame pointers

src/unix/ftrace.c
briankoco: I doubt gcc would mcount() a function that it also tail optimizes -- but I will double check

src/unix/ftrace.c
```diff
@@ -247,19 +322,51 @@ printer_print_right_adjusted(struct ftrace_printer * p, char * str,
...
         printer_write(p, "<< unknown symbol >>");
 }
 
+/* write timestamp in "num" to the printer using at most "width" spaces */
+static inline void
+printer_print_duration_usec(struct ftrace_printer * p, timestamp num, u16 width)
+{
+    u64 usec, nsec;
+
+    /* this is kind of nasty, but would rather not do memory allocation on every
```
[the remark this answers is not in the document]
briankoco: done

src/unix/ftrace.c
```diff
@@ -15,25 +15,60 @@
...
-struct rbuf_entry {
+struct rbuf_entry_function {
     unsigned long ip;
     unsigned long parent_ip;
-    unsigned long tsc;
+    unsigned long ts; /* XXX only supports tsc at the moment */
     unsigned short cpu;
     int tid;
-    char name[15]; /* unfortunate but safer than accessing a thread pointer */
+    char name[15];
```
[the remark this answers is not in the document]
briankoco: updated to use symbols, thanks for the tip

Issue #1003, opened 2019-11-05
node closing socket after failing getsockopt w/SO_ERROR


eyberg:
https://github.com/nanovms/ops/issues/403

```
eyberg@:~/nz/nt/node-v11.5.0-linux-x64/bin$ cat config.json
{
  "Args": ["node", "hi.js"],
  "Files": ["/lib/x86_64-linux-gnu/libresolv.so.2", "/lib/x86_64-linux-gnu/libnss_dns.so.2", "hi.js"]
}
eyberg@:~/nz/nt/node-v11.5.0-linux-x64/bin$ cat hi.js
const http = require('http');

http.get('http://api.nasa.gov/planetary/apod?api_key=DEMO_KEY', (resp) => {
  let data = '';

  // A chunk of data has been received.
  resp.on('data', (chunk) => {
    data += chunk;
  });

  // The whole response has been received. Print out the result.
  resp.on('end', () => {
    console.log(JSON.parse(data).explanation);
  });

}).on("error", (err) => {
  console.log("Error: " + err.message);
});
```

```
1 socket
[1592884962.364664]  NET: socket: new tcp fd 18, pcb 0x0000000101801100
1 direct return: 18, rsp 0x76fc2ae8
1 connect
[1592884962.368102]  NET: connect_tcp: sock 18, tcp state 1, addr ffffffffb65b0734, port 80
[1592884962.370788]  NET: connect_tcp_bh: sock 18, tcp state 2, thread 1, lwip_status 0, flags 0x0
1 sleep interruptible (on "sock receive")
[1592884962.433713]  NET: connect_tcp_complete: sock 18, tcp state 2, pcb 0x0000000101801100, err 0
[1592884962.436679]  NET: connect_tcp_bh: sock 18, tcp state 4, thread 1, lwip_status 0, flags 0x1
1 thread_wakeup: 1->1 blocked_on sock receive, RIP=0x70c9c11777
1 run frame 0x0000000101201800, RIP=0x00000070c9c11777
1 clock_gettime
1 clock_gettime: clk_id 6
1 direct return: 0, rsp 0x76fc7098
1 epoll_ctl
1 direct return: 0, rsp 0x76fc3fb8
1 epoll_pwait
1 direct return: 1, rsp 0x76fc3f70
1 clock_gettime
1 clock_gettime: clk_id 6
1 direct return: 0, rsp 0x76fc3f98
1 getsockopt
[...]

PR #1006, opened 2019-11-08
various bug fixes

wjhun:
These bugs were found while investigating https://github.com/nanovms/ops/issues/403 and #1003.

- getsockopt needed support for SO_ERROR
- missing poll events method for std files
- futex:
  - FUTEX_CMP_REQUEUE was incorrectly interpreting val2 as a timeout, and was also waking waiters before re-queuing them to the uaddr2 futex when it should only re-queue (use blockq_transfer_waiters to re-queue)
  - on thread exit, don't call futex syscall directly with a wake (causing a soft create); use futex_wake instead - if the futex isn't found, just ignore it
- file syscalls: explicitly check if file operation completed before putting thread to sleep (triggered by a file_read spanning only a file hole in the write runtime test - not related to node)

### DIFF
--- a/src/unix/futex.c
+++ b/src/unix/futex.c
@@ -112,7 +112,6 @@ static timestamp get_timeout_timestamp(int futex_op, u64 val2)
 {
     switch (futex_op) {
     case FUTEX_WAIT:
-    case FUTEX_CMP_REQUEUE:
     case FUTEX_WAIT_BITSET:
         return (val2) 
             ? time_from_timespec((struct timespec *)pointer_from_u64(val2)) 
@@ -156,14 +155,14 @@ sysreturn futex(int *uaddr, int futex_op, int val,
             false, ts
         );
     }
-            
+
     case FUTEX_WAKE: {
         if (verbose)
             thread_log(current, "futex_wake [%ld %p %d] %d",
                 current->tid, uaddr, *uaddr, val);
         return set_syscall_return(current, futex_wake_many(f, val));
     }
-        
+
     case FUTEX_CMP_REQUEUE: {
         int wake1, wake2;
         struct futex * f2;
--- a/src/unix/futex.c
+++ b/src/unix/futex.c
@@ -1,10 +1,5 @@
 #include <unix_internal.h>
 
-struct futex {
-    heap h;
-    blockq bq;
-};
-
 static u64 futex_key_function(void *x)
 {
     return u64_from_pointer(x);
@@ -57,6 +52,18 @@ static thread futex_wake_one(struct futex * f)
     return w;
 }
 
+boolean futex_wake(process p, int *uaddr)
+{
+    struct futex * f;
+
+    f = table_find(p->futices, (void *)uaddr);
+    if (!f)
+        return false;
+
+    futex_wake_one(f);
+    return true;
+}
+
 /*
  * Wake up to 'val' waiters
  * Return the number woken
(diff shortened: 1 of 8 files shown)


src/unix/futex.c
```diff
@@ -156,50 +164,37 @@ sysreturn futex(int *uaddr, int futex_op, int val,
...
-                false, ts
-             );
+        woken = futex_wake_many(f, val);
+
+        requeued = 0;
+        if (val2 > 0) {
+            struct futex * new = soft_create_futex(current->p, u64_from_pointer(uaddr2));
+            if (new == INVALID_ADDRESS)
+                return set_syscall_error(current, ENOMEM);
+            int requeued = blockq_transfer_waiters(new->bq, f->bq, val2);
```
briankoco: Is transferring waiters semantically any different from waking and queuing on a different blockq? Certainly the interface
now is cleaner, just wondering how the original functionality would have lead to problems

src/unix/futex.c
```diff
@@ -1,10 +1,5 @@
 #include <unix_internal.h>
 
-struct futex {
```
briankoco: Any reason for making this public?

src/unix/thread.c
```diff
@@ -130,13 +130,15 @@ closure_function(1, 0, void, run_thread,
 
 void thread_sleep_interruptible(void)
 {
+    disable_interrupts();
```
briankoco: Is this just here in preparation of having interrupts on in kernel? Interrupts should already be off, right?

src/unix/futex.c
```diff
@@ -57,6 +52,18 @@ static thread futex_wake_one(struct futex * f)
     return w;
 }
 
+boolean futex_wake(process p, int *uaddr)
```
briankoco: I know we already have a function called futex_wake_one(), but it might be nice to have the name of this function express the fact that at most one waiter is being woken here

src/unix/futex.c
wjhun: I think they are different. I think that's the point of FUTEX_CMP_REQUEUE - that is, prevent a thundering herd problem by only waking up a specified number (val) waiters and then partitioning existing waiters by moving them into a new futex (but without waking them).

In my case, node was crashing after these additional waiters were woken up. I may have to reproduce to give you the play-by-play as I don't remember the exact sequence - but I think the first wakeup lead to the thread exit, and then a subsequent wakeup (for one of the requeued items) was leading to scheduling of an already terminated thread, causing runloop to crash.

I'll try to repro this crash again to make sure I didn't miss another issue, but I think it's pretty clear that only val waiters should be woken up.

futex(2) man page:

> Typical  values to specify for val are 0 or 1.  (Specifying INT_MAX is not useful, because it would make the FUTEX_CMP_REQUEUE operation equivalent to FUTEX_WAKE.)

> Both  FUTEX_REQUEUE  and FUTEX_CMP_REQUEUE can be used to avoid "thundering herd" wake-ups that could occur when using FUTEX_WAKE in cases where all of the waiters that are woken need to acquire another futex.  Consider the following scenario, where multiple waiter threads are waiting on B, a wait queue implemented using a futex:
> 
>                   lock(A)
>                   while (!check_value(V)) {
>                       unlock(A);
>                       block_on(B);
>                       lock(A);
>                   };
>                   unlock(A);
> 
> If a waker thread used FUTEX_WAKE, then all waiters waiting on B would be woken up, and they would all try to acquire lock A.  However, waking all of the threads in this manner would be pointless because all except  one of  the  threads  would immediately block on lock A again.  By contrast, a requeue operation wakes just one waiter and moves the other waiters to lock A, and when the woken waiter unlocks A then the next waiter can proceed.

src/unix/futex.c
wjhun: Ah, that was from an earlier version of (global) futex_wake which took a pointer to the struct as an argument. Will fix, thanks.

src/unix/thread.c
wjhun: Yes, they will be off for now, though I think we should continue to write things as if interrupts could potentially be on for syscall / bh processing at some point.

src/unix/futex.c
wjhun: Agreed - changed to more descriptive names

PR #1013, opened 2019-11-13
cpuid check for rdseed/rdrand on init only

wjhun:
try_hw_seed was unnecessarily checking for rdseed/rdrand availability on each call. This trivial fix should confine all (costly) uses of cpuid to initialization.

### DIFF
--- a/src/x86_64/service.c
+++ b/src/x86_64/service.c
@@ -203,8 +203,8 @@ static void read_kernel_syms()
 
 extern void install_gdt64_and_tss();
 
-static boolean have_rdseed;
-static boolean have_rdrand;
+static boolean have_rdseed = false;
+static boolean have_rdrand = false;
 
 static boolean hw_seed(u64 * seed, boolean rdseed)
 {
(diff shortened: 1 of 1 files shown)


src/x86_64/service.c
```diff
@@ -233,13 +225,24 @@ static boolean try_hw_seed(u64 * seed, boolean rdseed)
...
-    if (try_hw_seed(&seed, true))
+    if (have_rdseed && hw_seed(&seed, true))
         return seed;
-    if (try_hw_seed(&seed, false))
+    if (have_rdrand && hw_seed(&seed, false))
         return seed;
     return (u64)now();
 }
 
+static void init_hwrand(void)
```
briankoco: - We have U64_FROM_BIT() that you could use for this
- have_rdseed / have_rdrand should either be initialized to false when declared or set to false based on the cpuid calls here

wjhun: Results from cpuid() are u32s (eax, ebx etc). Updated to explicitly initialize booleans (though all of bss is zeroed anyway).

PR #1017, opened 2019-11-15
KVM: use TSC-Deadline timer as runloop timer, if available

wjhun:
Profiling has shown significant latency when programming the local APIC timer. Such writing to the APIC registers is trapped, leading to a vm exit and subsequent device emulation within QEMU. This is quite expensive, taking anywhere from 4 to tens of microseconds in traces of webg performance.

KVM has the capability to emulate the TSC-Deadline timer within the host kernel. This is available in qemu whenever kernel-irqchip is set to on or split (default when KVM is enabled). While an APIC register needs to be written once to enable the timer on initialization, interval updates occur through an MSR which is then handled by the kernel irqchip. Avoiding costly device emulation by qemu in userspace, such updates are measured to take ~1.4us with much less variation.

On my machine, a test run against webg of "ab -n 9000 -c 50" leads to ~11.7k reqs/s with TSC Deadline, as opposed to ~9.8k reqs/s with the local APIC timer - a nearly 20% increase in performance.

There are tricks that we can do to ameliorate the expense of programming the timer, such as adding a small margin to the timer expiry in an attempt to aggregate more timer events together per interrupt. Xen hypervisors achieve this with a configurable "timer_slop" parameter that defaults to 50-100us, depending on the distribution. We can consider adding a similar option to allow the user to trade off accuracy for the sake of improved performance.

### DIFF
--- a/src/x86_64/pvclock.c
+++ b/src/x86_64/pvclock.c
@@ -12,7 +12,8 @@ u64 pvclock_now_ns(void)
     u64 result;
 
     do {
-        version = vclock->version & ~1; /* mask enable */
+        /* mask update-in-progress so we don't match */
+        version = vclock->version & ~1;
         read_barrier();
         u64 delta = rdtsc() - vclock->tsc_timestamp;
         if (vclock->tsc_shift < 0) {
@@ -55,7 +56,8 @@ closure_function(0, 1, void, tsc_deadline_timer,
     u64 count = 0;
 
     do {
-        version = vclock->version & ~1; /* mask enable */
+        /* mask update-in-progress so we don't match */
+        version = vclock->version & ~1;
         read_barrier();
         timestamp subsec = truncate_seconds(interval);
         count = (nsec_from_timestamp(subsec) << 32) / vclock->tsc_to_system_mul;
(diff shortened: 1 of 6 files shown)


src/x86_64/pvclock.c
```diff
@@ -34,5 +40,50 @@ void init_pvclock(heap h, struct pvclock_vcpu_time_info *vti)
...
+    /* debug here */
+}
+
+closure_function(0, 1, void, tsc_deadline_timer,
+                 timestamp, interval)
+{
+    u32 version;
+    u64 count = 0;
+
+    do {
```
briankoco: can you just invoke pvclock_now() here?

src/x86_64/pvclock.c
```diff
@@ -34,5 +40,50 @@ void init_pvclock(heap h, struct pvclock_vcpu_time_info *vti)
...
+    pvclock_heap = h;
     register_platform_clock_now(closure(h, pvclock_now));
 }
+
+closure_function(0, 0, void, tsc_deadline_interrupt)
+{
+    /* debug here */
+}
+
+closure_function(0, 1, void, tsc_deadline_timer,
```
briankoco: Minor quibble, but Is it possible to abstract the vclock stuff out from here so that pvclock_now() and tsc_deadline_timer() don't have to duplicate this logic?

wjhun: Aside from the version check which is a single conditional, these two are actually the reverse of each other (TSC interval -> timestamp for pvclock_now and the reverse for the timer).

PR #1019, opened 2019-11-17
unix files: handle fs (I/O) operations that complete without blocking

wjhun:
Use of the ATA driver, which poll-waits for data, exposed discrepancies in unix file operations which expected all tfs operations to block. These changes will detect if a completion has taken place before blocking the calling thread, in which case a direct return will take place instead of a thread sleep.

Note that there are still issues with the iovec operations (readv, writev) which will be addressed in a subsequent PR.


wjhun:
Resolves #1018 crash

PR #1020, opened 2019-11-19
revised iovec (readv, writev) and sendfile implementations

wjhun:
The poll-and-wait ATA driver exposed flaws in the way io_completions were being handled in the iovec and sendfile syscalls. Previously, these completions were applied for blocking operations, but internal versions of those completions were called directly after a direct return from an I/O op. It's likely that this logic was never really tested when using the virtio storage devices - which always block on a read or write. Additionally, certain cases were not being handled, such as partial buffer reads/writes in iovec operations and partial writes in sendfile(2). readv, writev and sendfile runtime tests were all failing when testing with the ATA driver.

These routines have been re-written to eliminate the multiple completion paths. They also take advantage of the new-style closures, which allow a closure to re-use itself as an io_completion. Said routines now only create a single closure in the top half and release it after the entire operation has completed. Note that the now-writable environment (enclosed variables) are used for state-keeping across completion applications. I/O operations for all fdesc types now apply any passed completion, whether the operation requires blocking or returns directly (most done via blockq_handle_completion helper). Previously, completions were only applied when waking up from blocking in a bottom-half call.

All runtime tests now pass with the ATA driver, and webg and node have been verified on t2.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -349,9 +349,8 @@ static sysreturn sock_read_bh_internal(sock s, thread t, void * dest, u64 length
 
     rv = xfer_total;
   out:
-    if (flags & BLOCKQ_ACTION_BLOCKED)
-        blockq_set_completion(s->rxbq, completion, t, rv);
-    net_debug("   returning %ld\n", rv);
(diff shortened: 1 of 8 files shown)


briankoco:
It all seems good to me

Issue #829, opened 2019-05-24
implement flock(2)


eyberg:
we need something for this but there might be considerations for this

https://gavv.github.io/articles/file-locks/

https://github.com/jruby/jruby/issues/828

eyberg:
also badgerdb wants this 

```
 go get github.com/dgraph-io/badger/...
```

```
  item, err := txn.Get([]byte("answer"))
  handle(err)

  var valNot, valCopy []byte
  err := item.Value(func(val []byte) error {
    // This func with val would only be called if item.Value encounters no error.

    // Accessing val here is valid.
    fmt.Printf("The answer is: %s\n", val)

    // Copying or parsing val is valid.
    valCopy = append([]byte{}, val...)

    // Assigning val slice to another variable is NOT OK.
    valNot = val // Do not do this.
    return nil
  })
```

```
1 flock
1 nosyscall flock
```

i haven't tried but since this is really for multiple processes but we could prob. just stub this out now, for a quick win

jonericcook:
I tried the above code and got an error from badger saying that it couldnt get a lock on the badger folder as someone else was using it.

cdosoftei:
:+1: for `flock` since it's commonly used by daemons to manage their PID files ... not that they are useful in an unikernel environment but it's nice to be able to run them out of the box, without unikernel specific tweaks.

cdosoftei:
The stub above definitely solves the issue for the `flock` syscall, thanks! However, I just realized in my case the implementation handles the file locks via `fcntl` instead (e.g. `F_SETLK` and friends).

Adding a simple successful return [here](https://github.com/nanovms/nanos/blob/4d781c541634ee10186a0d8713678b573c8afb0a/src/unix/syscall.c#L1679) does the job -- I am more than happy to provide a PR (or just open a separate dedicated ticket, let me know)

wjhun:
@cdosoftei: Sure, you are welcome to submit a PR. It looks like F_SETLK and F_SETLKW can just return zero, and F_GETLK can write F_UNLCK in lock->l_type if lock is nonzero (and return zero).

May I ask what application you are trying to run?

cdosoftei:
Thanks @wjhun I'll take a stab at the PR!

The application in question is [FreeSWITCH](https://github.com/signalwire/freeswitch), I realize it may be a longer shot given the implications of handling real-time media (timers for example) but I am sure it's doable.

PR #1021, opened 2019-11-20
some fixes for badger

wjhun:
- stub out flock (just returns 0)
- A file-backed mmap(2) was leading to a buffer allocation equal to the given mmap length, even if the file length itself is much less. This caused an issue in badger whereby a request for a large amount of address space (2gb in this case) was leading to a buffer alloc failure. Going forward, treat the map as two parts: the file-backed portion, sized at the minimum of the file size or given length, and an anonymous portion if the given length is greater than the file size.

Keep in mind that we do not yet support writing to files through a backed mmap, and any program which does this will not see persistence across reboots (or maps, for that matter).

With these changes, the badger test in #829 gives:

```
assigned: 10.0.2.15                                                                                                                                                                                                                            
badger 2019/11/20 14:42:39 INFO: All 0 tables opened in 1ms                                                                                                                                                                                    
The answer is: 42                                                                                                                                                                                                                              
badger 2019/11/20 14:42:39 DEBUG: Storing value log head: {Fid:0 Len:29 Offset:45}                                                                                                                                                             
badger 2019/11/20 14:42:39 INFO: Got compaction priority: {level:0 score:1.73 dropPrefix:[]}                                                                                                                                                   
badger 2019/11/20 14:42:39 INFO: Running for level: 0                                                                                                                                                                                          
badger 2019/11/20 14:42:39 DEBUG: LOG Compact. Added 2 keys. Skipped 0 keys. Iteration took: 1.585ms                                                                                                                                           
badger 2019/11/20 14:42:40 DEBUG: Discard stats: map[]                                                                                                                                                                                         
badger 2019/11/20 14:42:40 INFO: LOG Compact 0->1, del 1 tables, add 1 tables, took 49.518ms                                                                                                                                                   
badger 2019/11/20 14:42:40 INFO: Compaction for level: 0 DONE                                                                                                                                                                                  
badger 2019/11/20 14:42:40 INFO: Force compaction on level 0 done                                                                                                                                                                              
exit status 1
```

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -324,7 +324,6 @@ closure_function(6, 2, void, mmap_read_complete,
 
     kernel_heaps kh = (kernel_heaps)&t->uh;
     heap pages = heap_pages(kh);
-    heap physical = heap_physical(kh);
 
     // mutal misalignment?...discontiguous backing?
     u64 length_padded = pad(length, PAGESIZE);
@@ -338,15 +337,6 @@ closure_function(6, 2, void, mmap_read_complete,
 
     if (length < length_padded)
         zero(pointer_from_u64(where + length), length_padded - length);
-
-    if (length_padded < mmap_len) {
-        u64 bss = pad(mmap_len, PAGESIZE) - length_padded;
-        if (!mapped)
-            map(where + length_padded, allocate_u64(physical, bss), bss, mapflags, pages);
-        else
-            update_map_flags(where + length_padded, bss, mapflags);
-        zero(pointer_from_u64(where + length_padded), bss);
-    }
 
     if (mapped) {
         deallocate_buffer(b);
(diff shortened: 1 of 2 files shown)


src/unix/mmap.c
```diff
@@ -680,26 +670,39 @@ static sysreturn mmap(void *target, u64 size, int prot, int flags, int fd, u64 o
...
     thread_log(current, "  read file at 0x%lx, %s map, blocking...", where, mapped ? "existing" : "new");
+    u64 flen = MIN(pad(f->length, PAGESIZE), len);
+
+    /* Paint file portion */
+    struct vmap q;
+    q.flags = vmflags;
+    q.node.r = irange(where, where + flen);
+    vmap_paint(h, p->vmaps, &q);
+
+    /* And any remaining portion as anonymous (temporary until we have true file demand paging / write-back) */
```
briankoco: Do we need two separate regions here? Seems that we don't really rely on the anonymous flag anywhere other than for sanity checking in the page fault handler, which could just drop that check. We'll need to do so anyways once we put support in for demand paging the file-backed portion itself as well

wjhun: That's true, it doesn't necessarily need to be two different regions, and we can just fault in a zeroed page whenever a fault happens in a valid vmap regardless of flags. In the future, the fault handler would be wired up to a buffer / page cache which can propagate writes (dirty pages) back to disk, etc.

PR #1027, opened 2019-11-22
fcntl: add F_GETLK, F_SETLK, F_SETLKW

cdosoftei:
Per the conversation in https://github.com/nanovms/nanos/issues/829#issuecomment-557301456, this PR implements the support for the `F_GETLK`, `F_SETLK`, `F_SETLKW` _fcntl_ operations.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -1670,7 +1670,7 @@ sysreturn close(int fd)
     return 0;
 }
 
-sysreturn fcntl(int fd, int cmd, int arg)
+sysreturn fcntl(int fd, int cmd, u64 arg)
 {
     fdesc f = resolve_fd(current->p, fd);
 
@@ -1695,7 +1695,7 @@ sysreturn fcntl(int fd, int cmd, int arg)
         return set_syscall_return(current, 0);
     case F_GETLK:
         if (arg) {
-            ((struct flock *)(long)arg)->l_type = F_UNLCK;
+            ((struct flock *)arg)->l_type = F_UNLCK;
         }
         return set_syscall_return(current, 0);
     case F_SETLK:
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -50,7 +50,7 @@ void register_other_syscalls(struct syscall *map)
     register_syscall(map, msgsnd, 0);
     register_syscall(map, msgrcv, 0);
     register_syscall(map, msgctl, 0);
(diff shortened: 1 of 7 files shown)


src/unix/syscall.c
```diff
@@ -1697,6 +1697,14 @@ sysreturn fcntl(int fd, int cmd, int arg)
 
         f->flags = arg & ~O_CLOEXEC;
         return set_syscall_return(current, 0);
+    case F_GETLK:
+        if (arg) {
+            ((struct flock *)(long)arg)->l_type = F_UNLCK;
```
wjhun: I think perhaps arg should be changed to a u64 or void *; otherwise the address could possibly be truncated to 32 bits. (It probably passes in the runtime test below as the stack is in 32-bit space, but could fail for mmapped memory in high mem.)

test/runtime/fcntl.c
```diff
@@ -0,0 +1,89 @@
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdio.h>
+#include <fcntl.h>
+#include <errno.h>
+
+/* expect = 0 for success, errno otherwise */
+void getlk(int fd, struct flock *lock, int expect)
```
wjhun: Minor nit, but these three functions are all nearly identical and could be folded into one with parameterized op (aside the check for F_UNLCK).

src/unix/syscall.c
cdosoftei: Thanks for the review!

I went with `u64` on this one after taking a quick look at the Linux implementation; I think it makes sense here too since `arg` its mostly used in numeric capacity rather than a pointer, so this would avoid excessive type casting.

test/runtime/fcntl.c
cdosoftei: Perfect! I just squashed all three functions into one specific to these three `*LK` commands -- later on, more tests can be added for the rest of the commands.

wjhun:
@cdosoftei, would you mind doing a force push to this branch? I had to flip a switch to allow CI builds for forked repos, and a push should hopefully re-trigger a build.

cdosoftei:
Hmm, the force push alone didn't cut it so I added an empty commit (https://github.com/nanovms/nanos/pull/1027/commits/5f0905d2f72d0327aab79dabd7852cf6748604b6) -- I believe this approach works.

PR #1039, opened 2019-12-05
Add ftrace utilities and readme

briankoco:


briankoco:
fyi, this is all self contained in a new top level directory (trace-utilities). Easiest way to check it out is to just checkout the ftrace-graph branch and navigate to trace-utilities in github

PR #1042, opened 2019-12-09
pci_discover: probe 32 devices per PCI bus

cdosoftei:
Allow `pci_discover` to scan all 32 devices per PCI bus.


cdosoftei:
Thank you for the input, I've just made the necessary change. This also leads me to two observations:

1. In the `for` loop comparison, I've changed the operator from _greater_ to _greater or equal_ since the meaning of the `PCI_SLOTMAX` constant isn't the total number of possible slots but the highest supported slot number;

2. In the same spirit, shouldn't `msi_map` be allocated an additional entry? i.e.

https://github.com/nanovms/nanos/blob/ea92930270a68eec22d9353bee9cb330e9f74ef9/src/x86_64/pci.c#L47

changed to:

```C
static u32 *msi_map[PCI_SLOTMAX + 1];
```

Issue #776, opened 2019-04-29
Implement a vdso-safe version of now() for vdso gettimeofday(2)


wjhun:
PR #775 introduces a ".vdso" section for functions which should be callable from userspace, i.e. via a vdso mapping. Our kernel's implementation of now() is not safe to use from userspace at this point. The vdso gettimeofday falls back to a syscall for the time being.

To make this a legitimate vdso gettimeofday:

- Look at making the kvm pvclock struct user-accessible. This would allow one to write a now() that is safe to use from userspace (provided no other kernel pages are accessed).
- Much like kvm does on vm entry with pvclock, we could stash a TSC offset and scaling factor in a user-readable page on each entry into userspace (thread run, return from syscall, etc). This would obviate the need to share the pvclock and perhaps be more portable in situations where we don't have a pvclock struct available (are there any? will we ever run on bare metal?)
- determine if we should even be using HPET as a fallback clock source - this seems dubious given the platforms we're looking at

Related: #126

Issue #126, opened 2018-10-08
make a proper vdso


convolvatron:
currently we're counting on the runtime to use a fallback to an older version that depends on symbols being at particular physical address (in the pic-gap)

build and link in a vdso and advertise it through the aux element so we can augment the user space with direct calls and debugging

PR #1044, opened 2019-12-10
vdso support

briankoco:
This PR introduces a VDSO implementation (and thus addresses #126 and #776). There are also some other minor bugfixes and an update to ftrace

For the most part, there is nothing surprising here and most of the major changes are in the following files:
- `src/x86_64/vdso.c`
- `src/x86_64/vdso-now.c`
- `src/x86_64/clock.c`

1. `vdso.c`: the actual VDSO functions exported to userspace.
2. `vdso-now.c`: implements a function called `vdso_now()` which performs a VDSO-safe version of our `now()` callback, invoking the appropriate "backend" routine (e.g., pvclock if we have it)
3. `clock.c`: `now()`, `uptime()`, and our TSC routines, which must all now access/modify variables via the custom VDSO mapping that is shared with userspace

Currently, the only VDSO-accessible clock source is the pvclock. If it is not present, we fallback to syscalls. Though the framework is designed such that others should be easy to add if desired


briankoco:
hmm, just discovered that glibc 2.23 does not use the VDSO with statically linked executables ... perhaps related to this: https://lists.debian.org/debian-glibc/2016/04/msg00131.html

I'll try and test on a machine with a newer glibc but don't have one available at the moment. I don't believe that should require any additional kernel support though

src/x86_64/vdso.c
```diff
@@ -0,0 +1,142 @@
...
+{
+    return do_syscall(SYS_time, t, 0);
+}
+
+static sysreturn
+do_vdso_clock_gettime(clockid_t clk_id, struct timespec * tp)
+{
+    timestamp ts;
+
+    switch (clk_id) {
```
wjhun: Not clear to my why there is a switch here on clk_id when vdso_now does so also; could it be confined to vdso_now()?

briankoco:
@wjhun ok, I've addressed these. Do you mind taking a quick look through the updates to make sure this is what you wanted for the clock routines? I re-inlined now(), uptime(), and rdtsc()

PR #1048, opened 2019-12-11
signalfd(2) support

### LINKED ISSUE #931: implement signalfd(2)
signalfd(2) provides an fd for receiving signal events.

#925

wjhun:
This introduces support for signalfd(2). Notes:
- The notify_set has uses outside of the context of poll waiters. In order to use it for signalfd notifications, the eventmask was expanded to 64-bits.
- State and events for fds (fdescs) were previously process-wide. However a signalfd, which may be used in any thread in a process, may reflect state that is specific to a particular thread (e.g. for a signal that is delivered at the thread (not thread group / process) level). As such, the poll waiters now specify the blocking thread when checking for fdesc events. event_handlers also now take an optional thread argument so that posted events are only relevant if the collecting waiter is blocked on behalf of the specified thread.
- sigstate_flush_queue was broken, causing an assertion failure if a thread exited without handling all pending signals

Resolves #931


src/unix/blockq.c
```diff
@@ -201,6 +201,22 @@ thread blockq_wake_one(blockq bq)
     return t;
 }
 
+boolean blockq_wake_one_for_thread(blockq bq, thread t)
```
briankoco: doesn't seem like this is used anywhere, am I missing something?

PR #1050, opened 2019-12-16
implement timerfd(2)

wjhun:
This introduces support for timerfd(2) as well as some generic timer infrastructure for the unix world. The runtime timer interface has been changed slightly to allow differing start and periodic intervals as well as absolute timestamps.

Resolves #992

Issue #992, opened 2019-10-15
timerfd_create(2)


eyberg:
apparently both foundationdb && mongo want it 

```
1 [initandlisten] timerfd_create
1 [initandlisten] nosyscall timerfd_create
1 [initandlisten] epoll_ctl
```

Issue #789, opened 2019-05-03
setitimer(2) support


eyberg:
every now and then running the built-in php webserver will stall - and it looks like it might be correlated to calls to this

```
1 rt_sigaction
1 direct return: 0, rsp 0x7d7fe5f8
1 rt_sigprocmask
1 direct return: 0, rsp 0x7d7fe6c8
1 getcwd
1 direct return: 2105525696, rsp 0x7d7fc490
1 chdir
1 direct return: 0, rsp 0x7d7fc548
1 setitimer
1 nosyscall setitimer
1 time
1 direct return: 1556901838, rsp 0x7d7f7e68
1 open
1 open: "/index.php", flags 0, mode 1b6
1    fd 5, length 33, offset 0
1 direct return: 5, rsp 0x7d7fb128
1 fstat
```

```
➜  p  ops load php_7.2.13 -d -c config.json -p 3030
➜  p  cat config.json
{
  "Args":["-S", "0.0.0.0:3030"],
  "Files": ["index.php"]
}
```

eyberg:
doesn't seem to be failing on this anymore but it's a diff timer than we have support for now ->

```
1 direct return: 4, rsp 0x76f5aea8
1 setitimer
setitimer error: timer type 0 not yet supported
1 direct return: -95, rsp 0x76f5b078
1 setitimer
setitimer error: timer type 0 not yet supported
1 direct return: -95, rsp 0x76f5b078
1 close
1 close: fd 4
```

little bit of  strange error msg considering the defs -> 

https://github.com/nanovms/nanos/blob/e1145092c30397fca77992e5e43efe6d8e502ebf/src/unix/system_structs.h#L209-L211
https://github.com/nanovms/nanos/blob/e1145092c30397fca77992e5e43efe6d8e502ebf/src/unix/timer.c#L567

PR #1054, opened 2019-12-20
unix timers

### LINKED ISSUE #448: implement timer_create(2)
haskell wants it...

```
cat h.hs
main = putStrLn "Hello, World!"
```

```
eyberg@s1:~/h$ ops run -f -d h
*snip*
nosyscall timer_create
 1 write
:  1 write
timer_create 1 write
: Function not implemented
 1 exit_group
exit_groupexit status 1
```

low priority but the first os contributor on ops looks like a haskell user

wjhun:
This PR introduces support for itimer, timerfd and posix timer interfaces, implementing the following system calls:

- alarm
- getitimer
- setitimer
- timer_create
- timer_settime
- timer_gettime
- timer_getoverrun
- timer_delete
- timerfd_create
- timerfd_gettime
- timerfd_settime

In the process of implementing these, a number of improvements have been made to the runtime timer facility as well as signal handling:

timers
- Runtime timers accept an initial interval as well as an optional, periodic interval to better match the unix interfaces.
- Timer callbacks (of type timer_handler) now accept a count of overruns. When a periodic timer expiry is more than one interval behind the current time, the next expiry is advanced past the current time and the number of overruns is computed from the delta.
- Timer registration, and in turn other interfaces using it, now accept a flag to indicate if the expiry is absolute.

signals
-  The signal delivery path detects posix timer signals (type SI_TIMER) and accumulates overruns accordingly. As per spec, only one signal is queued at a timer per posix timer - even if the signal is a realtime (queueable) signal.
- A bug / false assumption in setup_sigframe was causing crashes in signal handlers that used SSE instructions (e.g. movaps). Rather than align the stack to 16 bytes, align it to 8 - but not 16. The x86-64 ABI dictates that the stack must be aligned to 16 bytes before a call, however we are entering the handler directly - which is analogous to entering after a call (which saves a return address, causing the 8 byte alignment). A function prologue will re-align to 16 before executing the body.
- Signal dispatching was previously occurring only on return from a syscall or in run_thread, which was clearly insufficient for timely delivery. In order to dispatch signal handlers after interrupt handling, a hook has been added to process_bhqueue back into the unix world to invoke dispatch_signals(current) if it so happens that a running_frame was a thread frame. rt_sigreturn was also missing a dispatch call to pick up any additional signals to process.

misc
- added missing thread release in epoll_blocked_free

TODO:
- Process and thread-based clocks are not yet implemented (itimer: ITIMER_VIRTUAL, ITIMER_PROF; posix: CLOCK_PROCESS_CPUTIME_ID, CLOCK_THREAD_CPUTIME_ID). Runtime timer expiries are made monotonic before adding them to the timer heap, but there isn't really a direct mapping from a process/thread clock to a system monotonic time. However, there are ways around this. TBD
- There is a half-finished hook into the unix timer code for notification of an RTC time change event. This is required to implement TFD_TIMER_CANCEL_ON_SET for timerfds. This isn't presently wired up and needs further work on the RTC side. (See notify_unix_timers_of_rtc_change(), which would probably be better implemented as a generic closure registration.)
- An attempt has been made to make a robust runtime test for all these interfaces, but there are surely some corner cases and error conditions that are not being checked.

supercedes #1050, and resolves #448, #992, and #789

### DIFF
--- a/src/runtime/timer.c
+++ b/src/runtime/timer.c
@@ -22,7 +22,7 @@ define_closure_function(1, 0, void, timer_free,
     deallocate(theap, bound(t), sizeof(struct timer));
 }
 
-timer register_timer(clock_id id, timestamp val, boolean absolute, timestamp interval, thunk n)
+timer register_timer(clock_id id, timestamp val, boolean absolute, timestamp interval, timer_handler n)
 {
     timer t = allocate(theap, sizeof(struct timer));
     if (t == INVALID_ADDRESS) {
@@ -38,27 +38,31 @@ timer register_timer(clock_id id, timestamp val, boolean absolute, timestamp int
 
     init_refcount(&t->refcount, 1, init_closure(&t->free, timer_free, t));
     pqueue_insert(timers, t);
-    timer_debug("register timer: %p, expiry %T, interval %T, thunk %p\n", t, t->expiry, interval, t);
+    timer_debug("register timer: %p, expiry %T, interval %T, handler %p\n", t, t->expiry, interval, n);
     return t;
 }
 
-/* Presently called with ints off. Address thread safety with
-   pqueue before using with ints enabled.
-*/
 timestamp timer_check()
 {
-    timestamp here = 0;
-    timer t = 0;
+    timer t;
+    s64 delta;
+    timestamp here = now(CLOCK_ID_MONOTONIC);
 
-    while ((t = pqueue_peek(timers)) &&
-           (here = now(CLOCK_ID_MONOTONIC), timer_expiry(t) <= here)) {
+    while ((t = pqueue_peek(timers)) && (delta = here - timer_expiry(t), delta >= 0)) {
         pqueue_pop(timers);
         if (!t->disabled) {
-            apply(t->t);
             if (t->interval) {
-                t->expiry += t->interval;
-                pqueue_insert(timers, t);
-                continue;
+                u64 overruns = 1;
+                if (delta > t->interval)
+                    overruns += delta / t->interval;
+                apply(t->t, overruns);
+                if (!t->disabled) {
+                    t->expiry += t->interval * overruns;
+                    pqueue_insert(timers, t);
+                    continue;
+                }
+            } else {
+                apply(t->t, 1);
             }
         }
         refcount_release(&t->refcount);
--- a/src/runtime/timer.c
+++ b/src/runtime/timer.c
@@ -39,7 +39,7 @@ timer register_timer(clock_id id, timestamp val, boolean absolute, timestamp int
     /* one ref for consumer, one for pqueue */
     init_refcount(&t->refcount, 2, init_closure(&t->free, timer_free, t));
     pqueue_insert(timers, t);
-    timer_debug("register timer: %p, expiry %T, interval %T, thunk %p\n", t->expiry, interval, t);
+    timer_debug("register timer: %p, expiry %T, interval %T, thunk %p\n", t, t->expiry, interval, t);
     return t;
 }
 
--- a/src/runtime/timer.c
+++ b/src/runtime/timer.c
@@ -52,9 +52,7 @@ timestamp timer_check()
         pqueue_pop(timers);
         if (!t->disabled) {
             if (t->interval) {
-                u64 overruns = 1;
-                if (delta > t->interval)
-                    overruns += delta / t->interval;
+                u64 overruns = delta > t->interval ? delta / t->interval + 1 : 1;
                 apply(t->t, overruns);
                 if (!t->disabled) {
                     t->expiry += t->interval * overruns;
--- a/src/runtime/timer.c
+++ b/src/runtime/timer.c
@@ -36,8 +36,7 @@ timer register_timer(clock_id id, timestamp val, boolean absolute, timestamp int
     t->disabled = false;
     t->t = n;
 
-    /* one ref for consumer, one for pqueue */
-    init_refcount(&t->refcount, 2, init_closure(&t->free, timer_free, t));
+    init_refcount(&t->refcount, 1, init_closure(&t->free, timer_free, t));
     pqueue_insert(timers, t);
     timer_debug("register timer: %p, expiry %T, interval %T, thunk %p\n", t, t->expiry, interval, t);
     return t;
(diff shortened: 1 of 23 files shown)


src/runtime/timer.c
```diff
@@ -6,136 +6,74 @@
...
-/* Presently called with ints off. Address thread safety with
-   pqueue before using with ints enabled.
-*/
-timestamp timer_check()
-{
-    timestamp here = 0;
-    timer t = 0;
+    timer t;
+    s64 delta;
+    timestamp here = now(CLOCK_ID_MONOTONIC);
```
convolvatron: its a bit fussy, and I don't know the details. but is CLOCK_ID_MONOTIC really something that's fundamental or is it semantics carried from the linux interface?

src/runtime/timer.h
```diff
@@ -15,53 +30,95 @@ static inline void runloop_timer(timestamp duration)
...
 
-// danger - truncation, should always be subsec
-
-static inline timestamp seconds(u64 n)
+/* Convert to monotonic. Not clear yet how to map process and thread
+   times to monotonic scale. Should the process have its own timer heap? */
+static inline timestamp timer_expiry(timer t)
 {
-    return n * TIMESTAMP_SECOND;
+    switch (t->id) {
```
convolvatron: right. sorry, this is really the block that seems like it should be in 'unix'

src/unix/timer.c
```diff
@@ -0,0 +1,655 @@
...
+    return bound(ut)->overruns > 0 ? EPOLLIN : 0;
+}
+
+closure_function(1, 0, sysreturn, timerfd_close,
+                 unix_timer, ut)
+{
+    unix_timer ut = bound(ut);
+    remove_unix_timer(ut);
+    deallocate_blockq(ut->info.timerfd.bq);
+    deallocate_closure(ut->f.read);
```
convolvatron: this is an opportunity for struct closure?

not really a comment, just curious

src/unix/unix.c
```diff
@@ -185,6 +185,9 @@ process create_process(unix_heaps uh, tuple root, filesystem fs)
     p->start_time = now(CLOCK_ID_MONOTONIC);
     init_sigstate(&p->signals);
     zero(p->sigactions, sizeof(p->sigactions));
+    p->posix_timer_ids = create_id_heap(h, 0, U32_MAX, 1);
+    p->posix_timers = allocate_vector(h, 8);
```
convolvatron: I guess ostensibly these limit would be better as a define

src/x86_64/def64.h
```diff
@@ -13,6 +13,15 @@ typedef __uint128_t u128;
 typedef u64 word;
 typedef u64 bytes;
 
+// XXX verify
```
convolvatron: strictly speaking these are true regardless of how many bits wide the registers are

src/runtime/timer.c
wjhun: I believe it's from posix 1b. To support timer_create, clock_nanosleep, etc, we need to somehow parameterize the clock source used as a reference. Not clear that this is the best way to do it; we could have instead used a vft per clock id.

src/unix/signal.c
```diff
@@ -228,16 +226,35 @@ static void deliver_signal(sigstate ss, struct siginfo *info)
     heap h = heap_general(get_kernel_heaps());
     int sig = info->si_signo;
 
-    /* check if we can post */
-    if (sig < RT_SIG_START && sigstate_is_pending(ss, sig)) {
-        /* Standard signal already posted. Unless a particular signal
-           would allow the updating of posted siginfo, just return here. */
-        sig_debug("already posted; ignore\n");
-        return;
+    /* Special handling for pending signals */
```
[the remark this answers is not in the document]
wjhun: Right, perhaps the signal delivery can return the queued_signal as a handle and a call can be added to update it. I'll take a crack at that before merging.

src/unix/signal.c
```diff
@@ -336,7 +353,7 @@ void deliver_signal_to_process(process p, struct siginfo *info)
        signal first, so this could cause a spurious wakeup (EINTR) ... care? */
     if (can_wake) {
         sig_debug("attempting to interrupt thread %d\n", can_wake->tid);
-        if (!thread_attempt_interrupt(t))
+        if (!thread_attempt_interrupt(can_wake))
```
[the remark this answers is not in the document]
wjhun: If the thread is in an interruptible wait on a blockq, the bottom half closure is removed from the queue and invoked with the nullify flag. The various bottom halves in turn return to thread with -EINTR.

src/unix/timer.c
wjhun: Yes, as they are one-to-one. Since this is a pattern throughout our unix world, it would be best done in another PR.

src/unix/unix.c
wjhun: These aren't limits, just default alloc sizes for the vectors.

src/unix/signal.c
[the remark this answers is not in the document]
wjhun: On second thought, this would then require that a signal dequeue call back into the timer code to nullify the reference. If we were going to do that, we may as well never push overrun updates and instead add special handling for SI_TIMER on dequeue to call into the timer code for the latest value of overrun. I believe this is what Linux does. That's kind of kludgey, too, but it would be less overhead per timer expiry.

I guess I'm inclined to punt for now if there isn't a correctness issue. I also kind of wonder how likely it is that this search would be for N > 1. If a program is combining SI_TIMER and other queueable events on the same rt signal (signo >= 32), then yes it does mean unnecessary searching.

I guess if we actually see very high rate posix timers in real-world use, we can revisit this. Maybe a little mechanism to install a general closure callback to update a siginfo before finally dispatching to a thread would be a nice thing to have?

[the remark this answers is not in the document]
eyberg: i wouldn't block this if it's functional - can always open an issue and track future work desired

src/runtime/timer.h
wjhun: I guess I started off assuming that these various clocksources would need sorting out at the lowest levels, so I made the internal CLOCK_ID_* identifiers line up with the unix CLOCK_* ones. But perhaps that assumption is wrong, as it seems in the end to just boil down to monotonic or realtime. So I guess we could revert to now() with no parameter and expose rtc_offset to the unix world. I supposed this could be addressed when implementing CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID.

Issue #1060, opened 2020-01-10
potential code generation issue with clang on macos


wjhun:
While developing in the https://github.com/nanovms/nanos/tree/kernlock branch on macos/clang (commit https://github.com/nanovms/nanos/commit/6ce5efc05f6bd0a96d80a95458a62b7259ae6992), I get the following crash soon after the runloop is started and interrupts are enabled:

```
qemu-system-x86_64 -m 2G -display none -serial stdio -drive if=none,id=hd0,format=raw,file=/Users/wjhun/src/nanos/kernlock2/output/image/disk.raw -device virtio-scsi-pci,id=scsi\
0 -device scsi-hd,bus=scsi0.0,drive=hd0 -device isa-debug-exit -no-reboot -smp 16 -d int -D int.log -device virtio-net,netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostf\
wd=tcp::9090-:9090,hostfwd=udp::5309-:5309 || exit $(($?>>1))
SMP test: 15 APs online

no fault handler for frame 0000000100a02200 (misc frame)

       cpu: 0000000000000000
 interrupt: 000000000000000e (Page fault)
     frame: 0000000100a02200
error code: 0000000000000000
   address: 0000000fa881644c

  rax: 0000000000000000
  rbx: 000000001999999a
  rcx: 000000007f261d60 (cpuinfos + 0000000000000000/0000000000000480)
  rdx: 0000000fa8816428
  rsi: 0000002af3258c66
  rdi: 0000000000000028
  rbp: 000000010001ff80
  rsp: 000000007701ffc8
   r8: 0000000000000028
   r9: 0000000100c00800
  r10: 0000000000000009
  r11: 0000000000000001
  r12: 0000000100000000
  r13: 0000020000000000
  r14: 000000000000a7c6
  r15: 000000007f061a20 (bootstrap_region + 0000000000000000/0000000000200000)
  rip: 000000007f0002c0
flags: 0000000000000046
   ss: 0000000000000000
   cs: 0000000000000008
   ds: 0000000000000000
   es: 0000000000000000
   fs: 0000000000000000
   gs: 0000000000000000

frame trace:
000000007f03d382        (runloop + 00000000000000e2/0000000000000118)
000000007f03ddbf        (init_service_new_stack + 000000000000055f/000000000000056b)
000000007f03d45b

stack trace:
000000007f03a676        (kernel_sleep + 0000000000000006/0000000000000009)
[...]

PR #1061, opened 2020-01-10
ap initialization and per-cpu structures

wjhun:
This PR is the union of the mpinit branch, the addition of per-CPU data structures (to which the various stacks and contexts have been migrated for the most part) and other ingredients for the ongoing SMP work.

This seemed as good a point as any for checkpointing into the mainline; SMP is disabled by default, and none of the AP initialization code is called. However, AP bringup can be tested out by adding "-smp n" to the qemu command line (where n is the total number of cpus) and defining SMP_TEST in the build.

An unrelated issue was found in the delivery of sigsegv within the unix fault handler. The fault handler was delivering the signal and then calling thead_yield - and in turn runloop - from within an interrupt context. This has been changed to allow the fault handler to exit gracefully while being assured that the signal will be delivered on return from exception handling.

Please note that an issue arose in the macos/clang build which leads to incorrect construction of IDT tables. A hack (making write_idt noinline) was added as a temporary workaround, but no cause has been isolated as of this point. See https://github.com/nanovms/nanos/issues/1060 for details.


src/x86_64/mp.c
```diff
@@ -0,0 +1,78 @@
...
+    void *n = allocate_stack(pages, 16);
+    asm volatile("mov %0, %%rsp": :"m"(n));
+    ap_new_stack();
+}
+
+void start_cpu(heap h, heap p, int index, void (*ap_entry)()) {
+    pages = p ;
+    start_callback = ap_entry;
+    if (!apboot) {
+        apboot =  (void *)0x8000;
```
convolvatron: I guess this is questionable. i..kind of assumed it would be safe since its the load
address for stage2 (?)

src/x86_64/page.h
```diff
@@ -18,7 +18,7 @@
 #define PAGEMASK           MASK(PAGELOG)
 #define PAGE_FLAGS_MASK    (PAGE_NO_EXEC | PAGEMASK)
 #define PAGE_PROT_FLAGS    (PAGE_NO_EXEC | PAGE_USER | PAGE_WRITABLE)
-#define PAGE_DEV_FLAGS     (PAGE_WRITABLE | PAGE_WRITETHROUGH | PAGE_NO_EXEC)
+#define PAGE_DEV_FLAGS     (PAGE_WRITABLE | PAGE_CACHE_DISABLE | PAGE_NO_EXEC)
```
convolvatron: this was a total rider....I thought cache disable was pretty mandatory for devices...that would imply write through...but its probably academic since there isn't actually any hardware back there?

src/x86_64/service.c
```diff
@@ -362,9 +424,24 @@ static void __attribute__((noinline)) init_service_new_stack()
...
+#ifdef SMP_TEST
+    ipi_vector = allocate_interrupt();
+    assert(ipi_vector != INVALID_PHYSICAL);
+    register_interrupt(ipi_vector, closure(heap_general(kh), ipi_interrupt));
+
+    init_debug("performing SMP test");
+    idle_cpu_queue = allocate_queue(misc, 64);
+    start_cpu(misc, pages, TARGET_EXCLUSIVE_BROADCAST, new_cpu);
+
+    kernel_delay(seconds(1));
```
convolvatron: I know this is just for testing, but I had a question about this in general. do we really ever need to barrier on 'all cpus available'? it seems like we can just let them trickle into the idle queue and pick up work? I guess that makes reporting the number of cpus difficult..but maybe that should come from the mptables?

src/x86_64/x86_64.h
```diff
@@ -66,9 +68,94 @@ static inline void disable_interrupts()
...
+       must be the third field (+16). */
+    void * syscall_stack;
+
+    /* common_handler switches to this stack when calling process_bhqueue */
+    void * bh_stack;
+
+    /* The default frame for when we're not in a thread or bh context. */
+    context misc_frame;
+
+    /* for bh processing, where page faults must be supported */
```
convolvatron: oh...I see, I found this all a bit confusing....but if bh context has to be able to take page faults then I'm totally lost. ... this is for mmap file? no, that can't be it.

src/x86_64/x86_64.h
```diff
@@ -66,9 +68,94 @@ static inline void disable_interrupts()
...
+{
+    return current_cpu()->running_frame;
+}
+
+static inline void set_running_frame(context f)
+{
+    current_cpu()->running_frame = f;
+}
+
+static inline void frame_push(context new)
```
convolvatron: ok...I have to think about that, I didn't realize that the frames were chained. doesn't that already solve some of the special case handling that I was complaining about? I guess I'd been advocating that we just merge this stack of frames into the scheduling queue. does that really make sense?

src/x86_64/x86_64.h
```diff
@@ -66,9 +68,94 @@ static inline void disable_interrupts()
...
+    new[FRAME_FAULT_HANDLER] = get_running_frame()[FRAME_FAULT_HANDLER];
+    frame_push(new);
+}
+
+static inline void frame_pop(void)
+{
+    set_running_frame(pointer_from_u64(get_running_frame()[FRAME_SAVED_FRAME]));
+}
+
+#define switch_stack(__s, __target) {                           \
```
convolvatron: mp.c should probably use this, right?

src/x86_64/mp.c
```diff
@@ -0,0 +1,78 @@
...
+        map((u64)apboot, (u64)apboot, 4096, PAGE_WRITABLE, h);
+    }
+    asm("sgdt %0": "=m"(ap_gdt_pointer));
+    asm("sidt %0": "=m"(ap_idt_pointer));    
+    mov_from_cr("cr3", ap_pagetable);
+    // just one function call
```
convolvatron: just going to mention that there is plenty of room at 0x8000 to make this call, so we don't really need to leak this page..but it doesn't really .. matter?

src/x86_64/mp.c
```diff
@@ -0,0 +1,78 @@
...
+    void *rsp = allocate_stack(pages, 1);
+    ap_stack = rsp;
+
+    runtime_memcpy(apboot, &apinit, &apinit_end - &apinit);
+    u8 vector = (((u64)apboot) >> 12) & 0xff;
+    
+    apic_ipi(index, ICR_TYPE_INIT);
```
convolvatron: apparently these delays and the repeat of the startup are traditional. but no real justification is given and there is some consternation in the various forums that the second startup could lead to some kind of race. its unlikely that the second startup is necessary on modern hardware, and almost certainly not for the hypervisor

src/x86_64/interrupt.c
```diff
@@ -194,77 +200,104 @@ void print_frame(context f)
...
+
+    /* bsp only right now */
+    if (ci->id == 0 && !ci->in_bh) {
+        frame_push(ci->bh_frame);
+        ci->in_bh = true;
+        switch_stack(ci->bh_stack, process_bhqueue);
     }
+    return;
+  exit_fault:
+    // XXX need to terminate bsp if on ap
```
convolvatron: I'm pretty sure you can set the nmi type on an ipi...thats probably appropriate

src/x86_64/apic.h
```diff
@@ -1,4 +1,40 @@
...
+#define ICR_DEST_ALL_EXC_SELF 0x000C0000
+
+#define TARGET_EXCLUSIVE_BROADCAST 0xfffffffful
+#define IA32_APIC_BASE_MSR 0x1b
+
 void lapic_eoi(void);
 void init_apic(kernel_heaps kh);
 void lapic_set_tsc_deadline_mode(u32 v);
 clock_timer init_lapic_timer(void);
+void apic_ipi(u32 target, u64 icr);
```
convolvatron: add vector argument?

src/x86_64/kernlock.c
```diff
@@ -0,0 +1,63 @@
...
+static inline int dequeue_cpu(void)
+{
+    u64 src = fetch_and_add(&kernlock.tail, 1);
+    return (int)kernlock_ring[src % kernlock.size];
+}
+
+static inline void enqueue_cpu(int cpu)
+{
+    u64 dest = fetch_and_add(&kernlock.head, 1);
+    kernlock_ring[dest % kernlock.size] = cpu; /* cas for paranoia? */
```
convolvatron: this has a race...it matches the original queue implementation that ... Levent I guess sort of made worse. 

otherwise the actual locked queue is fine. it seems like its a specialized version of something we will need more of? so I guess we could allocate it as an object rather than a global?

idk..maybe just a more primitive queue that's thread safe that we can use to implement queues of whatever?

convolvatron:
umm...just thought of something. we have that little spin lock in mp.c (which is going to kind of suck if we ever start running a lot of cores)....that makes sure that only one of the aps 
is using the shared stack, and also that only one ap is in the kernel.

but aside form the sleep(1) that we almost certainly want to remove, there isn't anything
right now stopping the bsp and one of the aps from using the unsafe kernel interfaces
simultaneously

src/x86_64/kernlock.c
convolvatron: yeah. thinking about it. this is really our mutex implementation, except that as we drop it into the more general part of the kernel it will be queuing closures instead of cpus...and if we fixed up the other queue (runqueue and I guess bhqueue?) that could probably be merged as well. not as sure about all the stuff in the block queue?

src/unix/syscall.c
```diff
@@ -1920,12 +1920,17 @@ struct syscall {
 static struct syscall _linux_syscalls[SYS_MAX];
 struct syscall *linux_syscalls = _linux_syscalls;
 
-static context syscall_frame;
+// this can actually change to the per-cpu miscframe; if the syscall
```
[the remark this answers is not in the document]
wjhun: syscall is under cli, but page faults need to be handled - in both top and deferred portions (thus the bh frame). Certainly these various frames are convoluted. We need to switch contexts when we might possibly take an exception or interrupt and don't want to lose the existing saved context. Whereas linux or some other os just pushes the frame to the stack within a thread context, we shuffle these context frames around. So the syscall context allows us to e.g. take page faults without trashing the saved thread context.

The peculiar thing about the syscall context right now is that, unless we do a direct return, we continue running on the sysall context (and stack!) after we call runloop and switch to something else - just by oversight, not design. So, while the syscall top half itself will be protected by the kernlock, that lock would be released as soon as we need to block, and we can't then call runloop and continue running kernel code on the (global) syscall_frame ... so we may as well just use the per-cpu miscframe. So clearly this needs to be cleaned up before actually handling syscalls on APs...but not necessarily for this PR.

I guess this is why I made that comment about the stack-of-frames. There isn't an absolute need to have a dedicated frame for the syscall, bhqueue processing, etc...so we could just say "push frame" any time we want to protect a saved context and might take an exception and have it grab the next free per-cpu frame...the depth wouldn't be more than a few levels anyway. And the fault handlers don't necessarily need to be per-frame, do they? We could just stick the single fault handler in the cpuinfo rather than copy them to each frame.

src/x86_64/crt0.s
```diff
@@ -9,17 +9,33 @@
...
+;; rflags
+;; cs
+;; rip
+;; [error code - if vec 0xe or 0xd]
+;; vector <- rsp
+
+%macro interrupt_common_top 0
         push rbx
-        mov rbx, [running_frame]
+        mov rbx, [gs:0]
```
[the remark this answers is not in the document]
wjhun: Yes...comment is on the mov rbx, [rbx + 8] below. The swapgs is outside of this macro as the stack offset varies with the exception.

src/x86_64/crt0.s
```diff
@@ -188,6 +222,7 @@ frame_return:
         mov rsp, [rax+FRAME_RSP*8]
         mov rcx, [rax+FRAME_RIP*8]
         mov rax, [rax+FRAME_RAX*8]
+        swapgs
```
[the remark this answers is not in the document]
wjhun: Thanks, I meant to add gs base in the context restore. Sad to say, that was always missing. (fs base is there - that's tls - not sure how gs is used in userspace).

We don't bother with save, as arch_prctl(2) is used to set %fs and %gs in the thread frame explicitly.

src/x86_64/interrupt.c
```diff
@@ -194,77 +200,104 @@ void print_frame(context f)
...
-context intframe;               /* for context save on exception within interrupt */
-context bhframe;
-
-void kernel_sleep()
+void kernel_sleep(void)
 {
-    running_frame = miscframe;
-    enable_interrupts();
-    __asm__("hlt");
+    __asm__("sti; hlt" ::: "memory");
```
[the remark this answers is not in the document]
wjhun: Not sure I understand here...we do need current_cpu()->running_frame to point to a valid frame any time we can take an interrupt, right? Or maybe you're referring to some other area?

src/x86_64/interrupt.c
```diff
@@ -194,77 +200,104 @@ void print_frame(context f)
...
-        if (i < 25) {
-            running_frame = apply(f, running_frame);
+        fault_handler fh = pointer_from_u64(f[FRAME_FAULT_HANDLER]);
+        if (fh) {
+            set_running_frame(apply(fh, f));
+        } else {
+            console("\nno fault handler for frame ");
+            print_u64(u64_from_pointer(f));
+            /* make a half attempt to identify it short of asking unix */
+            if (f == current_cpu()->misc_frame)
```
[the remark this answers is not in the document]
wjhun: The more I think of it, the more I like the idea of a stack of generic frames that aren't specific to any context - they don't need to be and are only transient containers of state while taking an exception. And again, I don't see why we need to have the fault handler attached to the frame. The thread frame fault handlers enclose the thread, but it's not even used - the page faults occur within the context of the process address space (singular in our case). And anyway, t->frame is only installed as running frame if current == t, right? Just trying to think if there's any reason to associate the fault handler with the context at all.

We should add names to the registered interrupts, though. Would greatly aid debugging.

src/x86_64/interrupt.c
```diff
@@ -379,7 +390,14 @@ void start_interrupts(kernel_heaps kh)
     void *idt_desc = idt_from_interrupt(n_interrupt_vectors); /* placed after last entry */
     *(u16*)idt_desc = 2 * sizeof(u64) * n_interrupt_vectors - 1;
     *(u64*)(idt_desc + sizeof(u16)) = u64_from_pointer(idt);
-    asm("lidt %0": : "m"(*(u64*)idt_desc));
+    asm volatile("lidt %0": : "m"(*(u64*)idt_desc));
+
+    u64 v = allocate_interrupt();
```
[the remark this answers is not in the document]
wjhun: But we don't - the one idt is shared by all cpus. Assuming that makes sense? Would we ever want to have different handlers on different cpus for a given vector? Even if we handle external ints on aps?

src/x86_64/service.c
wjhun: Absolutely. We have both the idle queue and the ci->online flag to tell us what cpus are available. This was just a test to see if they would all come up.

src/x86_64/x86_64.h
wjhun: The bhqueue processing can and does take page faults. Imagine lwip processing invokes a callback which wakes a socket, causing a blocked read to fill data in a user buffer over a non-present anonymous page.

But it is confusing, and again the stack of frames approach could clean this up.

src/x86_64/x86_64.h
wjhun: Well they weren't really chained - just one level. :) But, yes, if it was nestable to a few levels it would remove those specialized frames.

I guess it's confusing because in these transient cases we're not really trying to save a context to be restored at a later time, as with thread frames. We just want to retain one level of register save while possibly catching another - before popping our way back out. If we were to do our register saves to the stack as opposed to discrete frame records, this would be all automatic.

As far as I know, the only case where we want to hang on to a context to restore at some indeterminate later time is the thread frame...because, well, that's the only real notion of a thread that we have in the system.

src/x86_64/mp.c
wjhun: Interesting. I saw this exact process laid out in the intel sdm and just didn't question it - but I don't know how much of that is a catch-all for legacy devices.

src/x86_64/kernlock.c
wjhun: Not sure where the race is on the enqueue. There's no check for a semaphore maximum because it is assumed that the ring is sized to the maximum number of cpus, and thus cannot overflow.

Granted, it's possible that one cpu could bump the semaphore and that another cpu could immediately after bump the semaphore and also grab the next slot on the ring first. But as the advancing of the head is atomic, the first cpu would just get enqueued second. At first I thought this would be a problem, but it's just very close contention between two cpus - does it matter? It's not a priority inversion, right?

I think someone tried to make the generic queue safe for multiple readers as well...but that side is clearly broken. Here, we don't have multiple readers as there is only one lock holder at a time. I figured it would be cheaper to just implement it within rather than going to the general queue with the heavy dequeue, even if fixed.

Or maybe I'm totally missing something else?

The blockq is just a unix-specific thing to contain thread waiters on a resource. It could use a ring queue, but I just didn't want to deal with the hard upper limit. It may at some point need finer-grained locking around it if we get to that level of granularity in the kernel locking in general.

convolvatron: yeah, advancing the head is atomic, so there isn't any race between competing writers, they each get a unique and consecutive storage slot. the only issue is a race between the reader and a writer, since the write to the slot isn't atomic wrt the pointer advance.

so the reader could see that there is something on the queue, read it, and get back the old contents before the writer has a chance to update the queue entry

convolvatron:
I see, sorry. you've pre-allocated all the cpu structures...for some reason I thought we were doing allocations here

src/unix/syscall.c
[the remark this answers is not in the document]
convolvatron: just so I understand..the page faults here aren't from swapped pages clearly...they are from oversubscribed allocations? missing pages from an mmaped file? it seems possible, not suggesting necessarily better, to check for this rather than take the fault if its structurally less problematic

src/x86_64/kernlock.c
wjhun: Thanks...don't know how I missed that. I'll take this back to the drawing board - and maybe remove it for this PR.

src/unix/syscall.c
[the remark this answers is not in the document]
wjhun: Yes, oversubscribed allocations. There are programs that mmap vast regions of anonymous address space, expecting that the kernel supports (anonymous) demand paging. So at some point we needed to fault in anonymous pages. (I believe it was for ruby/sinatra...mmaping something like 700+MB of address space at a time.) We don't yet support demand paging for file-backed mmaps.

We certainly could manually install these pages in the syscall top half for any given user buffer and avoid taking a fault in a deferred operation. That is, unless there are reads that also point to huge buffers, expecting that only the needed pages would be faulted in.

I guess I don't see what the fundamental problem is with allowing deferred operations to fault in pages.

src/x86_64/x86_64.h
wjhun: Yes...and maybe stage3 too.

wjhun:
> I see, sorry. you've pre-allocated all the cpu structures...for some reason I thought we were doing allocations here

Yes, though I think down the road we can instead create per-cpu heaps and let them allocate for themselves so we can scale up and avoid MAX_CPUS.

src/unix/syscall.c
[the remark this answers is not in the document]
convolvatron: not a problem. just a choice. I'm not sure its super important if the end goal is to allow for interrupts in kernel space for other reasons. it does seem to complicate things...so if interrupts for page faults in i/o the only reason, then we could work around that.

how does this work if the user buffer (virtual, but without backing) is a direct target of a dma?

src/x86_64/kernlock.c
convolvatron: it is a pretty* small race. this is the one the levent was tasked with closing. if we know that the queue entry are always pointers that we have control over (not random ints, or user specified pointers), then there are plenty of bits to play with to mark invalid entries (which would require a spin to detect? I guess there is still an issue with progress if the writer took an interrupt right there (the only time this would really happen))...right, just INVALID pointer after the read and before advancing the read pointer would be enough here if we could tolerate the spin

src/unix/syscall.c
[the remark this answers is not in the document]
wjhun: It is admittedly a lazy approach. Plus, validation of user buffers is in order. In such a validation process, missing pages can be installed and protection violations can be caught preemptively without resorting to a fault...which of course would be absolutely necessary for the dma case as you mention (which presently isn't an issue as we don't implement any zero-copy interfaces). So I think that, once the ingredients are in order, we could avoid relying on page faults in the kernel outside of error handling.

src/x86_64/mp.c
wjhun: It kind of is an issue since that area is supposed to be open to userspace...right now there's nothing to prevent a collision with it.

So maybe we can set some deadline for the APs to come up and then just unmap it?

Also - any reason that ap.s is located in boot when it's linked into stage3?

convolvatron: well, we should fix it. ideally by using an allocator of some kind. I can look it up again, but I think it needs to be located in the lowest 1<<20? 

no, absolutely no reason...it wasn't clear how it was going to land when I started and it seemed to be more related to boot than anything else. and I was quite surprised that the makefile just worked..x86_86 is fine

convolvatron: what about 0? that's virtually unusable elsewhere, and would solve the offset issue :)

wjhun: So I was able to keep the address defined in a single place - the linker script. Which is kind of odd, because nothing actually gets linked at that address, but it does allow us to have a single place to define it.

Surprisingly, the zero page works. I say surprisingly because I thought the real mode interrupt vectors were in that page. But maybe it doesn't matter for the whole ap trampoline scheme? In any case, it's easy enough to move around now.

PR #1062, opened 2020-01-10
add optional reboot on exit

wjhun:
If the flag reboot_on_exit is specified in the build manifest, nanos will reboot (using an induced triple fault) on vm_exit (as a result of exit_group(2), an assertion failure, halt or unrecoverable fault) - as opposed to the default qemu exit with code.

Tested on GCE with a webg image which induces halt after a timeout (not included here). The server restarts and continues serving requests as expected.

(CI failure appears to occur in lepton go build:)
```
go get github.com/nanovms/ops/lepton
# google.golang.org/api/option
/go/src/google.golang.org/api/option/option.go:153:14: undefined: grpc.RoundRobin
/go/src/google.golang.org/api/option/option.go:154:42: undefined: grpc.WithBalancer
make[2]: *** [Makefile:29: deps] Error 2
make[2]: Leaving directory '/home/circleci/project/test/go'
make[1]: *** [Makefile:7: all] Error 2
make[1]: Leaving directory '/home/circleci/project/test'
make: *** [Makefile:86: test-noaccel] Error 2

Exited with code exit status 2
```

### DIFF
--- a/boot/stage2.c
+++ b/boot/stage2.c
@@ -4,6 +4,7 @@
 #include <page.h>
 #include <region.h>
 #include <x86_64.h>
+#include <kvm_platform.h>
 #include <serial.h>
 #include <drivers/ata.h>
 
@@ -250,7 +251,6 @@ region fsregion()
     halt("invalid filesystem offset\n");
 }
 
-
 closure_function(4, 2, void, filesystem_initialized,
                  heap, h, heap, physical, tuple, root, buffer_handler, complete,
                  filesystem, fs, status, s)
@@ -282,6 +282,11 @@ void newstack()
                       closure(h, filesystem_initialized, h, physical, root, bh));
     
     halt("kernel failed to execute\n");
+}
+
+void vm_exit(u8 code)
+{
+    QEMU_HALT(code);
 }
 
 static struct heap working_heap;
--- a/src/x86_64/interrupt.c
+++ b/src/x86_64/interrupt.c
@@ -384,3 +384,13 @@ void start_interrupts(kernel_heaps kh)
     /* APIC initialization */
     init_apic(kh);
 }
+
+void triple_fault(void)
+{
+    disable_interrupts();
+    /* zero table limit to induce triple fault */
+    void *idt_desc = idt_from_interrupt(n_interrupt_vectors);
+    *(u16*)idt_desc = 0;
+    asm volatile("lidt %0; int3": : "m"(*(u64*)idt_desc));
+    while (1);
+}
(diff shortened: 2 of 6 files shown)


convolvatron:
looks like exit() just goes back to runloop and exit_group() calls vm_exit. idk if exit() should really as well.

convolvatron:
...i...dont get the failure reported by circleci? maybe because I have particular versions of dependencies in the cache?

maybe we should consider using the new modules system to fix the dependency versions?

eyberg:
i thought ops was using modules - but i'll try and take a look tmo || mon

wjhun:
> looks like exit() just goes back to runloop and exit_group() calls vm_exit. idk if exit() should really as well.

To my knowledge, exit(2) only terminates the running thread, whereas exit_group(2) is used to terminate the whole process/thread group (used by exit(3)).

Not sure if we're covering all the corner cases though - i.e. if the last thread exits - shouldn't it do a vm_exit?

convolvatron:
you're absolutely right. what an odd thing to not know.

convolvatron:
the HN thread had several people complaining about 'seconds to restart', which I think is pretty valid given the suggested use case. I do certainly agree with Ian that from a security perspective, restarting the whole OS on each request, or on a regular amortized basis a really big lever. 

it seems like the seconds here come from GCE (as Ian implied). is it worth considering starting stage3 over again. maybe not as full a reset, but certainly would clean out quite a bit and not take very long. idk if nanos text is RO, but one could guarentee that all the runtime structures were rebuilt.

briankoco:
It would be really, really cool, but also very non-trivial, to be able to do something like this for stateful apps. i.e., checkpoint app (memory, open files, kernel data, etc.), reboot, then re-instantiate the address space layout for the app but a new layout for the kernel. The hard part would be serializing the in-kernel data structures (e.g., thread structs and all relevant fields) before reboot and then re-instantiating then somehow at a different region of the address space.

Of course, an easier route would just be to checkpoint+restart/migrate the whole VM, which won't give you a different guest VA layout but would likely move you to different physical machine memory. Given nanos' small mem footprint this might be able to be done more quickly than eg Linux. This would be useful for mitigating side channel attacks (e.g., rowhammer) that are cropping up everywhere

wjhun:
> It would be really, really cool, but also very non-trivial, to be able to do something like this for stateful apps. i.e., checkpoint app (memory, open files, kernel data, etc.), reboot, then re-instantiate the address space layout for the app but a new layout for the kernel. The hard part would be serializing the in-kernel data structures (e.g., thread structs and all relevant fields) before reboot and then re-instantiating then somehow at a different region of the address space.

It would be a significant effort, though if we continue to make progress with introspection and classifying the various kernel objects - we could get a handle on it. Basically any state that would be required to repopulate a fresh kernel instance and pick up execution would need to be typed and serialized/deserialized. We are kind of heading in that direction anyway, though for other reasons.

But even before attempting that, we can do a lot to speed up kernel initialization.

PR #1066, opened 2020-01-15
new lock-free queue

wjhun:
This is an implementation of a lock-free, multiple producer / multiple consumer ring queue with single producer (enqueue_single) and single consumer (dequeue_single) variants. It replaces src/x86_64/queue.c. The interface is largely unchanged outside of the addition of the single variants, however a return value of INVALID_ADDRESS is now used to indicate an empty queue from a call to dequeue, dequeue_single or queue_peek, and a null value may be legitimately stored in the queue. Queues are instantiated in only one way despite use of multi or single variants; the caller must only be consistent in use of the interface. MPMC, MPSC, SPMC and SPSC combinations are all supported.

Two sets of head and tail pointers are used for both producing and consuming, and the typical ring enqueue and dequeue operations are split internally into reserve and commit phases (for the multi variants). See https://doc.dpdk.org/guides/prog_guide/ring_lib.html for a visual description of this data structure.

A pthread-based unit test is included. Each thread operates on a common queue, continuously alternating between enqueueing and dequeueing randomly-picked numbers from a global vector for a set duration (default 8 threads for 5 sec). Parity between enqueues and dequeues and conservation of the picked numbers are asserted.


wjhun:
On reflection I realize that the queue indices do not need to be 64-bit at all. So I may take a stab at changing to 32-bit indices and then using 64-bit accesses to atomically read and update pairs (prod_head, cons_tail) and (cons_head, prod_tail) with the hope of eliminating any concerns with head and tail reads getting out of sync.

src/runtime/queue.c
```diff
@@ -0,0 +1,24 @@
...
+/* will round up size to next power-of-2 */
+queue allocate_queue(heap h, u64 size)
+{
+    if (size == 0)
+        return INVALID_ADDRESS;
+    int order = find_order(size);
+    queue q = allocate(h, _queue_alloc_size(order));
+    if (q == INVALID_ADDRESS)
+        return q;
+    void *buf = ((void *)q) + sizeof(struct queue);
```
convolvatron: in some world, it would be nice to have the body line-aligned

src/runtime/queue.h
```diff
@@ -0,0 +1,173 @@
...
+    int order;
+} *queue;
+
+#define _queue_size(q) (1ul << (q)->order)
+#define _queue_idx(q, i) ((i) & MASK((q)->order))
+#define _queue_assert(x) assert(x)
+#define _queue_pause kern_pause
+
+/* little endian */
+#define _head_from_combined(c) ((u32)((c) & MASK(32)))
```
convolvatron: it doesn't matter..but doesn't the union take care of this?

wjhun: Taking from the union defeats the 64-bit copy but, yes, it makes more sense to overlay the struct onto the copy and extract from that (endian safe).

src/runtime/queue.c
wjhun: made a statically-defined default cacheline size of 64 and aligned to it

Are there x86_64 implementations with other sizes? We could probably find it in a cpuid leaf if we really cared to...

convolvatron: looking right now apparently "netburst" machines have a 128byte line, but apparently its somehow split?

since its an optimization, fixed 64 byte alignment is just fine...i'm not even sure if it was
worth bothering you with the idea.

but yeah, I think there are probably other cases where cache aligning some object will be helpful

PR #1067, opened 2020-01-16
fix GDT entries for sysret

wjhun:
Not a serious problem: Incorrect GDT and STAR MSR setup was causing the wrong selector to be loaded on sysret. Eliminates corresponding kludge in common_handler.

### DIFF
--- a/src/x86_64/crt0.s
+++ b/src/x86_64/crt0.s
@@ -299,54 +299,61 @@ install_gdt64_and_tss:
         ret
 .end:
 
+%define SEG_DESC_G          (1 << 23) ; Granularity
+%define SEG_DESC_DB         (1 << 22) ; Code: default size, Data: big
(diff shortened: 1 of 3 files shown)


src/x86_64/crt0.s
```diff
@@ -299,46 +299,52 @@ install_gdt64_and_tss:
...
+        dw 0         ; Base (low).
+        db 0         ; Base (middle)
+        db 10011010b ; Access (exec/read).
+        db 00100000b ; Granularity.
+        db 0         ; Base (high).
+        .Data: equ $ - GDT64       ; data - 0x10
         dw 0         ; Limit (low).
         dw 0         ; Base (low).
         db 0         ; Base (middle)
         db 10010010b ; Access (read/write).
```
convolvatron: these could be called out by the or of symbolically defined bits describing their function...but...

wjhun: thanks, fixed

PR #1068, opened 2020-01-24
little thread test using a silly little dataflow program

convolvatron:
this passes at least on my Mac with qemu...what other platforms do we need to test on? does circle-ci take of that?

it could definitely use a timeout. I don't think its the _best_ threading test imaginable, but
at least it has to go through a lot of synchronization to produce the single verifiable result


convolvatron:
not clear..this is in the right place. where do the syscall unit tests go?

wjhun:
test/unit is where the unit tests go for various runtime datastructures. These run on the local host, not in nanos. test/runtime is the place for the tests which run in nanos.

Granted, the naming is kind of unfortunate - as is the organization. Certainly we could be running all the unit tests under nanos as well as natively, and also the runtime tests could be run natively on the host as a parity check (and also to verify the tests themselves).

Should these all be under one dir and run both on host and in nanos?

convolvatron:
I'm not sure..I think having runtime units separate is a good idea. I was thrown off with the name 'runtime' since that's what I assumed was in there :)

it occurs to me to that I should really be looking at TLS variables given all that's going on down there

convolvatron:
it does actually admit some concurrency since the generator thread pushes several values at once ... the thread queue reaches some number between 12 and 20.

this is really just a minimal slice of a more general test which does general data flow graphs...so there can be multiple legs running concurrently. I'm not 100% happy with the breadth of this as a thread test, but I couldn't think of another general way to exercise threads and get out a result one could just check.

I have most of a translator for sha256, but there isn't really very much concurrency there...there is a new hash called blake3 that is supposed to be designed for parallelism.

two additions I think would be helpful in the short term would be to use TLS instead of stack variables (to test GS handling) and construction of processing nodes which use syscalls (for example to implement an edge using pipe())

the other thing that would seem important is to validate epoll()...maybe there is already a highly threaded test for that (?)

wjhun:
> the other thing that would seem important is to validate epoll()...maybe there is already a highly threaded test for that (?)

Sadly, there isn't, and it's kind of critical that we get one. The epoll test that is there is just a basic test of the interface from a contributor. The signal and time tests use epoll to test signalfd and timerfd, respectively, but only within a single thread.

PR #1069, opened 2020-01-25
some tuple parser cleanup

wjhun:
Create a separate value_parser to differentiate between tuple and value parsing. Spurred by a mkfs crash caused by a malformed manifest (an ops issue), this should help catch malformed tuples rather than parse them as buffers. Also includes recycling of closures in combinate() and closure deallocations where possible.

This is kind of an incremental improvement; more hardening and error catching is possible.

### DIFF
--- a/src/runtime/tuple_parser.c
+++ b/src/runtime/tuple_parser.c
@@ -62,6 +62,7 @@ closure_function(1, 2, parser, til_newline,
     return self;
 }
 
+// leaky; no finite lifespan
 closure_function(2, 2, parser, eat_whitespace,
                  heap, h, parser, finish,
                  parser, self, character, in)
@@ -72,10 +73,7 @@ closure_function(2, 2, parser, eat_whitespace,
     if (member(whitespace, in))
         return self;
 
-    parser f = bound(finish);
-    deallocate_closure(self);
-    closure_finish();
-    return apply(f, in);
+    return apply(bound(finish), in);
 }
 
 static parser ignore_whitespace(heap h, parser next)
@@ -166,6 +164,7 @@ closure_function(3, 1, parser, parse_value_string,
                  heap, h, completion, c, buffer, b,
                  character, in);
 
+// leaky; no finite lifespan
 closure_function(3, 1, parser, parse_value,
                  heap, h, completion, c, err_internal, err,
                  character, in)
@@ -189,10 +188,10 @@ closure_function(3, 1, parser, parse_value,
         q = ignore_whitespace(h, (parser)closure(h, parse_value_string, h, c, allocate_buffer(h, 8)));
         p = apply(q, in);
     }
-    closure_finish();
     return p;
 }
 
+// leaky; no finite lifespan
 closure_function(3, 1, parser, parse_tuple,
                  heap, h, completion, c, err_internal, err,
                  character, in)
@@ -206,7 +205,6 @@ closure_function(3, 1, parser, parse_tuple,
         p = apply(bound(err), aprintf(bound(h), "parse_tuple fail, leading char '%c'", in));
     else
         p = ignore_whitespace(h, combinate(h, closure(h, is_end_of_tuple, h, c, allocate_tuple(), err)));
-    closure_finish();
     return p;
 }
 
@@ -243,11 +241,10 @@ static parser is_end_of_tuple(struct _closure_is_end_of_tuple *__self, parser se
 {
     heap h = bound(h);
     if (in == ')') {
-        completion c = bound(c);
-        tuple t = bound(t);
+        parser p = apply(bound(c), bound(t));
         deallocate_closure(self);
         closure_finish();
-        return apply(c, t);
+        return p;
     }
 
     // XXX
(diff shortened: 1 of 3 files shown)


src/runtime/tuple_parser.c
```diff
@@ -35,15 +35,12 @@ static charset charset_from_string(heap h, char *elements)
     return t;
 }
 
-// kind of a special case y combinator, try to fix the
-// allocation issues
-// XXX - can recycle closure?
+// kind of a special case y combinator
 closure_function(2, 1, parser, selfinate,
```
convolvatron: oh good. do we still really need selfinate, maybe you're just trying to minimize disruption here

src/runtime/tuple_parser.c
```diff
@@ -52,49 +49,59 @@ static parser combinate(heap h, selfparser p)
...
 
-
 closure_function(1, 2, parser, til_newline,
                  parser, finish,
                  parser, self, character, in)
 {
-    if (in == '\n') return bound(finish);
+    if (in == '\n') {
+        parser f = bound(finish);
+        deallocate_closure(self);
```
convolvatron: I thought closure_finish does deallocate the closure (?)

src/runtime/tuple_parser.c
```diff
@@ -52,49 +49,59 @@ static parser combinate(heap h, selfparser p)
...
-    if (in == '#') {
+    if (in == '#')
         return combinate(bound(h), closure(bound(h), til_newline, self));
-    }
 
-    if (member(whitespace, in)) return self;
+    if (member(whitespace, in))
+        return self;
+
     return apply(bound(finish), in);
```
convolvatron: doesn't this conditional leg mean that this closure is done and can be deallocated?

src/runtime/tuple_parser.c
```diff
@@ -108,6 +115,8 @@ closure_function(3, 2, parser, terminal,
     if (member(bound(final), in)) {
         // apply(apply(x)) calls x twice
         parser p = apply(bound(c), bound(b));
+        deallocate_closure(self);
+        closure_finish();
```
convolvatron: ditto on the closure_finish

src/runtime/tuple_parser.c
```diff
@@ -153,29 +164,48 @@ closure_function(3, 1, parser, parse_value_string,
...
+{
+    heap h = bound(h);
+    completion c = bound(c);
+    err_internal err = bound(err);
+    parser p;
+
+    if (in != '(')
+        p = apply(bound(err), aprintf(bound(h), "parse_tuple fail, leading char '%c'", in));
+    else
+        p = ignore_whitespace(h, combinate(h, closure(h, is_end_of_tuple, h, c, allocate_tuple(), err)));
```
convolvatron: is_end_of_tuple can free this, we'd have to pass self when taking this closure

src/runtime/tuple_parser.c
```diff
@@ -185,34 +215,43 @@ closure_function(4, 1, parser, name_complete,
...
     heap h = bound(h);
     if (in == ')') {
-        return apply(bound(c), bound(t));
+        parser p = apply(bound(c), bound(t));
+        deallocate_closure(self);
+        closure_finish();
+        return p;
     }
 
+    // XXX
```
convolvatron: what is the xx here ? (closure finish above)

src/runtime/tuple_parser.c
```diff
@@ -224,30 +263,30 @@ static parser is_end_of_vector(struct _closure_is_end_of_vector *__self, parser
...
         return combinate(h, closure(h, quoted_string, h, bound(c), bound(b)));
 
-    parser p = combinate(h, closure(h, terminal, bound(c), value_terminal, bound(b)));
-    return apply(p, in);
+    return apply(combinate(h, closure(h, terminal, bound(c), value_terminal, bound(b))), in);
 }
 
-
+// leaks; no end condition to free on
 closure_function(0, 2, parser, kill,
```
convolvatron: yeah, this has to live for the parser lifetime...so It could go there...I think this is just a drain on a syntax error?

src/runtime/tuple_parser.c
wjhun: Ok, I wasn't sure how to simplify this, but on further inspection - we shouldn't need the self_parser type at all, right? So maybe there's no need for combinate or selfinate if a parser can readily refer to itself. I'll give it a shot.

src/runtime/tuple_parser.c
wjhun: Right, but this was deallocating the closure from selfinate - but I think this can go away.

src/runtime/tuple_parser.c
wjhun: Yes, should go away as above...same for below

src/runtime/tuple_parser.c
convolvatron: I think so..but this is just based on my memory of saying 'well, we don't have self, so I guess we're going to have to do this ugly thing'...so I'm just assuming that since now we do, we don't

wjhun:
> this is all great cleanup..wondering a little what the high level reason for having two entry pmts (tuple and value) since tuples are also values

Really just error checking: A malformed tuple, if parsed as a value, would just be parsed as a long buffer value (til eof); explicitly parsing tuples as opposed to values will at least give some chance to reject bad input.

That said, there's a way to go to make this parser hardened in any practical sense (as in, not crash on bad input). I did some of this to debug what was ultimately an ops bug, and figured it was at least moving in the forward direction (with the closure reuse and alloc cleanup going along for the ride).

convolvatron:
absolutely

src/runtime/tuple_parser.c
wjhun: just cruft placed on first pass while I was trying to understand the code...no issue there that I see now

src/runtime/tuple_parser.c
wjhun: Perhaps building out an encompassing parser type would be order...though I think maybe for a future pass.

Issue #1072, opened 2020-02-07
investigate eval'ng graalvm


eyberg:
was just asked about it on drift - uses AOT to produce a 'native image' to run jvm programs w/out boot time and also runs many other languages in a one runtime to rule them all env

https://www.graalvm.org/docs/getting-started/

eyberg:
```
export PATH=/home/eyberg/graal/graalvm-ee-java11-19.3.1/bin:$PATH
gu -L install native-image-installable-svm-svmee-java11-linux-amd64-19.3.1.jar
```

```
eyberg@box:~/c$ cat Bob.java
public class Bob {

    public static void main(String[] args) {
      System.out.println("yo");
    }

}
```

```
javac Bob.java
native-image --no-server --no-fallback Bob
```

```
1 mmap
1 mmap: target 0x0000000000000000, size 0x800000000, len 0x0, prot 0x0, flags 0x4022, fd -1, offset 0x0
mmap error: failed to allocate  virtual memory, size 0x0
1 direct return: -12, rsp 0x76ed1cf8
1 fcntl
1 fcntl: fd 2, cmd 3, arg -128
1 direct return: 2, rsp 0x76ed1cc0
1 fstat
1 fd 2, stat 0x0000000076ed15e0
1 st_ino 0, st_mode 0x1000, st_size 0
1 direct return: 0, rsp 0x76ed15d8
1 write
Fatal error: Failed to create a new Isolate. (code 8)
1 direct return: 54, rsp 0x76ed1bb8
1 exit_group
exit status 17
eyberg@box:~/c$
```

eyberg:
mmap works w/ https://github.com/nanovms/nanos/commit/52374af66410032a2b768151c1bca582abe857ab although it fails when trying to read /proc/self/maps

```
1 mmap: target 0x0000000000000000, size 0x800000000, len 0x800000000, prot 0x0, flags 0x4022, fd -1, offset 0x0
1    anon target: 0x1000000000, len: 0x800000000 (given size: 0x800000000)
1 direct return: 68719476736, rsp 0x76c47cf8
1 openat
1 "/proc/self/maps" - not found
1 direct return: -2, rsp 0x76c47b20
1 munmap
1 munmap: addr 0x0000001000000000, size 0x800000000
1 direct return: 0, rsp 0x76c47d28
1 fcntl
1 fcntl: fd 2, cmd 3, arg -128
1 direct return: 2, rsp 0x76c47cc0
1 fstat
[...]

PR #1073, opened 2020-02-07
changes to allow larger anonymous mmaps

wjhun:
Fix a few things preventing large (e.g. 32gb) mmaps:

- unnecessary truncation of size in mmap
- allow for using alternate heap for bitmap data (e.g. backed_heap for larger allocations, used for process virtual page heap)
- extend user mmap area down to 0x800000000

Resolves this mmap failure: https://github.com/nanovms/nanos/issues/1072#issuecomment-583477615


wjhun:
> the only question I had was moving the user space virtual heap base...just curious

Was just trying to expand the amount of address space available for mmap. Not strictly necessary, but I figured we may as well if we have already had a few programs try and request gobs of address space. It limits the amount available to the kernel, but only for general allocations - not tagged ones.

PR #1086, opened 2020-03-06
Fix %x for 32-bit integers with MSB bit set (negative 32-bit integers)

mkhon:


src/runtime/runtime_init.c
```diff
@@ -40,8 +40,12 @@ static void format_number(buffer dest, struct formatter_state *s, vlist *a)
     s64 x;
     if (s->modifier == 'l')
         x = varg(*a, s64);
```
wjhun: Maybe support for '%u' and '%lu' could be added trivially here?

PR #1088, opened 2020-03-18
Filesystem: add support for symlinks

francescolavra:
A symlink is represented by a tuple with a symbol called "linktarget", which contains the path of the filesystem entry being linked to.
A new runtime test is also added to exercise symlink-related code.

### DIFF
--- a/src/unix/filesystem.c
+++ b/src/unix/filesystem.c
@@ -1,0 +1,53 @@
+#include <unix_internal.h>
+#include <filesystem.h>
+
+// fused buffer wrap, split, and resolve
+int resolve_cstring(tuple cwd, const char *f, tuple *entry, tuple *parent)
+{
+    if (!f)
+        return -EFAULT;
+
+    tuple t = *f == '/' ? filesystem_getroot(current->p->fs) : cwd;
+    tuple p = t;
+    buffer a = little_stack_buffer(NAME_MAX);
+    char y;
+    int nbytes;
+
+    while ((y = *f)) {
+        if (y == '/') {
+            if (buffer_length(a)) {
+                p = t;
+                t = lookup(t, intern(a));
+                if (!t)
+                    goto done;
+                if (!children(t))
+                    return -ENOTDIR;
+                buffer_clear(a);
+            }
+            f++;
+        } else {
+            nbytes = push_utf8_character(a, f);
+            if (!nbytes) {
+                thread_log(current, "Invalid UTF-8 sequence.\n");
+                p = false;
+                goto done;
+            }
+            f += nbytes;
+        }
+    }
+
+    if (buffer_length(a)) {
+        p = t;
+        t = lookup(t, intern(a));
+    }
+done:
+    if (!t && (*f == '/') && (*(f + 1)))
+        /* The path being resolved contains entries under a non-existent
+         * directory. */
+        p = false;
+    if (parent)
+        *parent = p;
+    if (entry)
+        *entry = t;
+    return (t ? 0 : -ENOENT);
+}
--- a/src/unix/filesystem.c
+++ b/src/unix/filesystem.c
@@ -1,6 +1,8 @@
 #include <unix_internal.h>
 #include <filesystem.h>
 
+#define SYMLINK_HOPS_MAX    8
+
 // fused buffer wrap, split, and resolve
 int resolve_cstring(tuple cwd, const char *f, tuple *entry, tuple *parent)
 {
@@ -12,14 +14,22 @@ int resolve_cstring(tuple cwd, const char *f, tuple *entry, tuple *parent)
     buffer a = little_stack_buffer(NAME_MAX);
     char y;
     int nbytes;
+    int err;
 
     while ((y = *f)) {
         if (y == '/') {
             if (buffer_length(a)) {
                 p = t;
                 t = lookup(t, intern(a));
-                if (!t)
+                if (!t) {
+                    err = -ENOENT;
                     goto done;
+                }
+                err = filesystem_follow_links(t, p, &t);
+                if (err) {
+                    t = false;
+                    goto done;
+                }
                 if (!children(t))
                     return -ENOTDIR;
                 buffer_clear(a);
@@ -29,6 +39,7 @@ int resolve_cstring(tuple cwd, const char *f, tuple *entry, tuple *parent)
             nbytes = push_utf8_character(a, f);
             if (!nbytes) {
                 thread_log(current, "Invalid UTF-8 sequence.\n");
+                err = -ENOENT;
                 p = false;
                 goto done;
             }
@@ -40,6 +51,7 @@ int resolve_cstring(tuple cwd, const char *f, tuple *entry, tuple *parent)
         p = t;
         t = lookup(t, intern(a));
     }
+    err = -ENOENT;
 done:
     if (!t && (*f == '/') && (*(f + 1)))
         /* The path being resolved contains entries under a non-existent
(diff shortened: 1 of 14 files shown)


src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+    int nbytes;
+    int err;
+
+    while ((y = *f)) {
+        if (y == '/') {
+            if (buffer_length(a)) {
+                p = t;
+                t = lookup(t, intern(a));
+                if (!t)
+                {
```
wjhun: inconsistent formatting

src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+                    err = -ENOENT;
+                    goto done;
+                }
+                err = filesystem_follow_links(t, p, &t);
+                if (err) {
+                    t = false;
+                    goto done;
+                }
+                if (!children(t))
+                    return -ENOTDIR;
```
wjhun: Why are some of these direct returns as opposed to return via done?

src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+            }
+            f += nbytes;
+        }
+    }
+
+    if (buffer_length(a)) {
+        p = t;
+        t = lookup(t, intern(a));
+    }
+    err = -ENOENT;
```
wjhun: Initialize err to -ENOENT as default?

src/unix/syscall.c
```diff
@@ -748,36 +646,54 @@ static int file_type_from_tuple(tuple n)
...
-        } else if (!n) {
-            fs_status fs = filesystem_creat(current->p->fs, cwd, name, mode);
-            if (fs != FS_STATUS_OK)
-                return sysreturn_from_fs_status(fs);
-
-            /* XXX We could rearrange calls to return tuple instead of
-               status; though this serves as a sanity check. */
-            n = resolve_cstring(cwd, name);
+        } else if ((ret == -ENOENT) && parent) {
+            n = filesystem_creat(current->p->fs, parent,
```
wjhun: Error check here?

src/unix/syscall.c
```diff
@@ -1076,11 +1006,16 @@ sysreturn getdents64(int fd, struct linux_dirent64 *dirp, unsigned int count)
 
 sysreturn chdir(const char *path)
 {
+    int ret;
     tuple n;
     if (path == 0)
         return set_syscall_error(current, EINVAL);
```
wjhun: Should be EFAULT, I think...can actually remove this clause as resolve_cstring_follow should check anyway, right?

src/unix/syscall.c
```diff
@@ -997,7 +926,7 @@ sysreturn getdents(int fd, struct linux_dirent *dirp, unsigned int count)
         char *p = cstring(symbol_string(k), tmpbuf);
         r = try_write_dirent(f->n, dirp, p,
```
wjhun: null check for dirp

src/unix/syscall.c
```diff
@@ -1026,7 +955,8 @@ static int try_write_dirent64(tuple root, struct linux_dirent64 *dirp, char *p,
             *read_sofar -= len;
             return -1;
         } else {
-            tuple n = resolve_cstring(root, p);
+            tuple n;
+            resolve_cstring(root, p, &n, 0);
```
wjhun: check for error?

src/unix/syscall.c
```diff
@@ -964,7 +892,8 @@ static int try_write_dirent(tuple root, struct linux_dirent *dirp, char *p,
...
```
wjhun: error check?

test/runtime/symlink.c
```diff
@@ -0,0 +1,121 @@
...
+    } \
+} while (0)
+
+int main(int argc, char **argv)
+{
+    int fd;
+    char buf[8];
+    struct stat s;
+    char *cwd;
+
```
wjhun: May want to add setbuf(stdout, NULL) so we don't lose any error output on a crash.

src/unix/filesystem.h
```diff
@@ -0,0 +1,44 @@
...
+    else if (__dirfd == AT_FDCWD) cwd = current->p->cwd; \
+    else { \
+        file f = resolve_fd(current->p, __dirfd); \
+        if (!is_dir(f->n)) return set_syscall_error(current, ENOTDIR); \
+        cwd = f->n; \
+    } \
+    cwd; \
+})
+
+static inline boolean is_dir(tuple n)
```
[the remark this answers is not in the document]
francescolavra: I'm going to move linktarget() and lookup_sym() to filesystem.h, the other filesystem-related definitions as of now cannot be moved because they are used by code (tfs.c and stage3.c) that doesn't include unix internals.

src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+
+int resolve_cstring_follow(tuple cwd, const char *f, tuple *entry,
+        tuple *parent)
+{
+    tuple t, p;
+    int ret = resolve_cstring(cwd, f, &t, &p);
+    if (!ret) {
+        ret = filesystem_follow_links(t, p, &t);
+    }
+    if (entry) {
```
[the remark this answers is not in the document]
francescolavra: Indeed setting *entry doesn't make much sense, because if ret != 0 it means the target tuple hasn't been found. But setting *parent is necessary for those cases (e.g. file and dir creation) where we need to know the parent of a filesystem entry that doesn't exist.

src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+    file_op_begin(current);
+    filesystem_symlink(current->p->fs, parent, filename_from_path(path), target,
+            closure(heap_general(get_kernel_heaps()), symlink_complete,
+            current));
+    return file_op_maybe_sleep(current);
+}
+
+sysreturn symlink(const char *target, const char *linkpath)
+{
+    thread_log(current, "symlink %s -> %s", linkpath, target);
```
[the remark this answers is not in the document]
francescolavra: This is already being done for linkpath in resolve_cstring(). I will add the check for target in symlink_internal().

src/unix/filesystem.c
```diff
@@ -0,0 +1,150 @@
...
+    return symlink_internal(current->p->cwd, linkpath, target);
+}
+
+sysreturn symlinkat(const char *target, int dirfd, const char *linkpath)
+{
+    thread_log(current, "symlinkat %d %s -> %s", dirfd, linkpath, target);
+    tuple cwd = resolve_dir(dirfd, linkpath);
```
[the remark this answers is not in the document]
francescolavra: Yes, the resolve_dir macro returns on error.

src/unix/filesystem.c
francescolavra: I used this formatting to minimize the diff lines of the commit that adds these parentheses, but OK, will fix it.

src/unix/filesystem.c
francescolavra: It wouldn't serve any purpose, because err could be overwritten within the preceding while() loop, so we would need to re-set it anyway after the loop.

src/unix/syscall.c
francescolavra: I will add a check for n != 0; checking for I/O errors during file creation (which wasn't being done even before this commit) would require more work, since it's an async operation.

src/unix/syscall.c
francescolavra: Right, will remove this clause since resolve_cstring_follow() does the check for us.

src/unix/filesystem.c
francescolavra: Direct returns are supposed to be used when the return code is different from -ENOENT, in which case the caller shouldn't bother checking what values have been put into *entry and *parent. (The only error case where *parent is set is for -ENOENT.) But now I see that there is a direct return -ENOENT in case of invalid UTF-8 sequence, will fix that.

src/unix/syscall.c
francescolavra: No error can occur, since p is a string retrieved from the children of root; and even if there was an error it wouldn't cause any crash because n is only used in a call to u64_from_pointer().

src/unix/syscall.c
francescolavra: Same as above.

Issue #1099, opened 2020-03-29
Runtime test errors do not propagate as Make errors


francescolavra:
Since commit ead28633c5aaf0075e3a66aa6c6cae49a35d0806, when running a runtime test that is supposed to fail, the failure does not propagate to the Makefile, thus running `make test` doesn't shown test failures, nor does running `make run`. Also, output written by programs via printf() doesn't show up in the terminal.

francescolavra:
Fixed by #1103.

PR #1103, opened 2020-04-02
pci_bar_init: fix mapping when offset exceeds PAGESIZE

wjhun:
A change introduced in #1098 was causing nanos to crash if a call to pci_bar_init was made with an offset > PAGESIZE. Assuming that the correct fix is to map starting the first page boundary at or after the offset (as opposed to further padding the map length).

May have triggered (but not be the root cause of) #1099.

### DIFF
--- a/src/x86_64/pci.c
+++ b/src/x86_64/pci.c
@@ -133,8 +133,9 @@ void pci_bar_init(pci_dev dev, struct pci_bar *b, int bar, bytes offset, bytes l
         bytes len = pad(length, PAGESIZE);
         b->vaddr = allocate(virtual_page, len);
         pci_debug("%s: %p[0x%x] -> 0x%lx[0x%lx]+0x%x\n", __func__, b->vaddr, len, b->addr, b->size, offset);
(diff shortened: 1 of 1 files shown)


oaktree11:
This fixed doing "make run" with qemu default options in the Makefile. Qemu would just drop out without any error.

eyberg:
huh, @oaktree11 would you mind doing an ```ops profile``` so we can have a better idea of end users environments?

oaktree11:
I am building nanos from source on Ubunu 18.04 - gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0. 
git log shows the commit I am using
commit 8c4cbae14ff8f8d13be1b023ff71873b30c3e55c
I had been running the webg app with the options from here (without the -s ) https://github.com/nanovms/nanos/wiki/using-the-qemu-remote-gdb-interface
I also tried qemu 4.2 and same issue
When I applied the fix in this commit to pci.c the default "make run" worked. 

ops is functioning fine - ops profile here:
Ops version: 0.1.9
Nanos version: 0.1.24
Qemu version: 2.11.1
Arch: linux

PR #1106, opened 2020-04-07
Fix boot on VMware ESXi

mkhon:
- allocate first available (lowest) physical region instead of highest:
VMware ESXi has second high memory region that intersects with the kernel map
- BIOS E820 memory map: type 3 and 4 are ACPI reclaimable memory and
ACPI NVS memory but they were used for IDENTITY and IDENTITY_RESERVED
region types in nanos
- create_region() and BIOS e820: make sure last region (list terminator) has type == 0
VMware has more regions initially so regions created by boot2 spill over boot0 code
- halt early if bios_read_sectors() failed

This allows nanos on VMware ESXi boot to succeed.

ESXi VIOS E820 memory regions are:

Screen Shot 2020-04-07 at 18 05 46

### DIFF
--- a/src/x86_64/region.h
+++ b/src/x86_64/region.h
@@ -12,11 +12,11 @@ typedef struct region *region;
 
 #define REGION_PHYSICAL          1 /* available physical memory */
 #define REGION_DEVICE            2 /* e820 physical region configured for i/o */
-#define REGION_IDENTITY          3 /* for page table allocations in stage2 and stage3 */
-#define REGION_IDENTITY_RESERVED 4 /* entire identity area which must be preserved in stage3 */
-#define REGION_FILESYSTEM        5 /* offset on disk for the filesystem, see if we can get disk info from the bios */
-#define REGION_KERNIMAGE         6 /* location of kernel elf image loaded by stage2 */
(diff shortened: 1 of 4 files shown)


src/x86_64/region.h
```diff
@@ -26,6 +26,7 @@ static inline region create_region(u64 base, u64 length, int type)
     r->base = base;
     r->length = length;
     r->type = type;
+    (--r)->type = 0;
```
wjhun, line 29: Will actually return the next region...but I guess nothing uses the return val?

src/x86_64/region.h
```diff
@@ -41,16 +42,16 @@ static inline u64 allocate_region(heap h, bytes size)
     u64 base = 0;
     region r;
 
-    /* Select the highest physical region that's within 32-bit space. */
```
wjhun, line 44: I suppose the original intent here was to try and keep the identity heap out of the way of low address space that the process might be using. As long as this doesn't break our other platforms, it should be ok for now. We really ought to move all kernel mappings to high (> 2^47) address space.

src/x86_64/region.h
mkhon: Oops. Fixed!

Issue #1075, opened 2020-02-14
mono wanting /proc/self/maps


eyberg:
here's another vote for proc/self/maps ->

```
eyberg@box:~/mono$ ops run -c config.json  mono
[mono hi.exe]
booting /home/eyberg/.ops/images/mono.img ...
qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
assigned: 10.0.2.15
mono_threads_platform_get_stack_bounds: pthread_getattr_np failed with "No such file or directory" (2)

=================================================================
        Native Crash Reporting
=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================
/proc/self/maps:

=================================================================
        Basic Fault Adddress Reporting
=================================================================
Memory around native instruction pointer (0x85483ee97):0x85483ee87  d2 4c 89 ce bf 02 00 00 00 b8 0e 00 00 00 0f 05  .L..............
0x85483ee97  48 8b 8c 24 08 01 00 00 64 48 33 0c 25 28 00 00  H..$....dH3.%(..
0x85483eea7  00 44 89 c0 75 1f 48 81 c4 18 01 00 00 c3 0f 1f  .D..u.H.........
0x85483eeb7  00 48 8b 15 a9 bf 3a 00 f7 d8 41 b8 ff ff ff ff  .H....:...A.....

=================================================================
        Native stacktrace:
=================================================================
        0x5208b6 - mono : (null)
        0x4b44c1 - mono : (null)
        0x83a412890 - /lib/x86_64-linux-gnu/libpthread.so.0 : (null)
        0x85483ee97 - /lib/x86_64-linux-gnu/libc.so.6 : gsignal
        0x854840801 - /lib/x86_64-linux-gnu/libc.so.6 : abort
        0x70a554 - mono : (null)
        0x6f0d21 - mono : (null)
        0x70a43b - mono : (null)
        0x70a863 - mono : monoeg_g_logv
        0x70a91f - mono : monoeg_g_log
        0x7025b9 - mono : (null)
```

```
eyberg@s1:~/mono$ cat hi.cs
using System;

public class HelloWorld
{

    public static void Main(string[] args)
    {
        Console.WriteLine ("Hello Mono World");
    }
}
```

```
 csc hello.cs
```

eyberg:
[...]

PR #1111, opened 2020-04-11
Special files: add /proc/self/maps

francescolavra:
This file contains the currently mapped memory regions and their access permissions.

#1075

### DIFF
--- a/src/unix/special.c
+++ b/src/unix/special.c
@@ -37,6 +37,50 @@ static u32 null_events(file f)
     return EPOLLOUT;
 }
 
+closure_function(1, 1, void, maps_handler,
+                 buffer, b,
+                 vmap, map)
+{
+    buffer b = bound(b);
+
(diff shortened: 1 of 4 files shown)


src/unix/special.c
```diff
@@ -37,6 +37,50 @@ static u32 null_events(file f)
...
+}
+
+static sysreturn maps_read(file f, void *dest, u64 length, u64 offset)
+{
+    heap h = heap_general(get_kernel_heaps());
+    buffer b = allocate_buffer(h, 512);
+    if (b == INVALID_ADDRESS) {
+        return -ENOMEM;
+    }
+    vmap_iterator(current->p, closure(h, maps_handler, b));
```
wjhun, line 69: Looks like this closure gets leaked...you can just use a stack_closure here as a local allocation is fine (not an async operation).

francescolavra: Right. Fixed by using stack_closure() as you suggest.

PR #1125, opened 2020-04-15
page cache implementation and generic support for scatter-gather I/O

wjhun:
This PR proposes to merge an implementation of a page cache as well as a scatter-gather list type (sg_list) that can be used to support vectorized I/O operations. To support zero-copy operations, the page cache exposes a scatter-gather (sg) read interface.

In this phase, a page cache is instantiated as a transparent cache proxy for block storage. TFS has been modified to support only an sg read interface to the backing storage. In the case of stage3, this is the page cache used to create the filesystem. In stage2 and dump, the page cache is replaced with a wrapper that interfaces directly to the block device. Write interfaces are left as buffer-based for the time being, and writes to the page cache are currently write-through only.

While these changes will absorb redundant read accesses to the storage block device, there are a number of improvements to be made:

near-term:
- the cache presently isn't releasing pages, which limits the amount of storage that can be accessed at the moment ... implement a drain function as well as some sort of high watermark / heuristic to determine maximum cache size system-wide (highest priority)
- write methods for the page cache and tfs can be transformed to sg-only (linear writes map trivially to sg as a single fragment)
- the filesystem interface used in the syscall layer can become sg based
- sendfile(2) can take advantage of the sg interface to allow zero-copy
- readv implementation can be simplified using tfs sg read interface
- writev can be made to eliminate copies by interfacing directly with a tfs sg write
- sequential page accesses can be optimized by following links rather than doing a rangemap lookup per page - implementation TBD
- tfs writes can eliminate the intermediate head/tail sector read step by writing directly into the page cache (basically just moving those prerequisite reads into the page cache - which is partially there already to accommodate partial page writes) - this eliminates a copy as well as obviates the use of "dma bufs" in tfs
- allow pagecache writes to complete immediately - but create a facility for catching and reporting I/O errors from (deferred) block device operations
- possibly replace rangemap page lookup with e.g. hash table, r-b tree or skip list lookup (although the pages still need to be threaded in order for easy traversal)

future:
- implement true file-backed mmaps using pagecache buffers
- using a block mapper (table of offset -> sector number), make caches linked to fsfiles (our inode equivalent) instead of (or in addition to) the entire block storage device - this eliminates the need for tfs to do extent lookups on cached accesses
- write-back support to aggregate storage writes

### DIFF
--- a/mkfs/dump.c
+++ b/mkfs/dump.c
@@ -145,7 +145,7 @@ int main(int argc, char **argv)
                       SECTOR_SIZE,
                       infinity,
                       h,
-                      closure(h, bread, fd, get_fs_offset(fd)),
+                      sg_wrapped_block_reader(closure(h, bread, fd, get_fs_offset(fd)), SECTOR_OFFSET, h),
                       closure(h, bwrite, fd),
                       root,
                       false,
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -516,9 +516,9 @@ static void vmap_paint(heap h, rangemap pvmap, vmap q)
     assert((rq.end & MASK(PAGELOG)) == 0);
     assert(range_span(rq) > 0);
 
-    rmnode_handler nh = stack_closure(vmap_paint_intersection, h, pvmap, q);
-    range_handler rh = stack_closure(vmap_paint_gap, h, pvmap, q);
-    rangemap_range_lookup_with_gaps(pvmap, rq, nh, rh);
+    rangemap_range_lookup_with_gaps(pvmap, rq,
+                                    stack_closure(vmap_paint_intersection, h, pvmap, q),
+                                    stack_closure(vmap_paint_gap, h, pvmap, q));
 
     update_map_flags(rq.start, range_span(rq), page_map_flags(q->flags));
 }
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -517,10 +517,8 @@ static void vmap_paint(heap h, rangemap pvmap, vmap q)
     assert(range_span(rq) > 0);
 
     rmnode_handler nh = stack_closure(vmap_paint_intersection, h, pvmap, q);
-    rangemap_range_lookup(pvmap, rq, nh);
-
     range_handler rh = stack_closure(vmap_paint_gap, h, pvmap, q);
-    rangemap_range_find_gaps(pvmap, rq, rh);
+    rangemap_range_lookup_with_gaps(pvmap, rq, nh, rh);
 
     update_map_flags(rq.start, range_span(rq), page_map_flags(q->flags));
 }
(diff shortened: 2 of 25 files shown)


src/unix/mmap.c
```diff
@@ -516,11 +516,9 @@ static void vmap_paint(heap h, rangemap pvmap, vmap q)
...
     assert(range_span(rq) > 0);
 
-    rmnode_handler nh = stack_closure(vmap_paint_intersection, h, pvmap, q);
-    rangemap_range_lookup(pvmap, rq, nh);
-
-    range_handler rh = stack_closure(vmap_paint_gap, h, pvmap, q);
-    rangemap_range_find_gaps(pvmap, rq, rh);
+    rangemap_range_lookup_with_gaps(pvmap, rq,
+                                    stack_closure(vmap_paint_intersection, h, pvmap, q),
+                                    stack_closure(vmap_paint_gap, h, pvmap, q));
```
francescolavra, line 522: This doesn't work if the new mapping is a sub-range of an existing mapping and the existing mapping needs to be resized and moved at the end of the new mapping: in this case, the gap that is created by the resize is not caught by rangemap_range_lookup_with_gaps(), and as a result the gap handler is not called and the new mapping is not inserted in the vmap.

mkfs/dump.c
```diff
@@ -145,7 +145,7 @@ int main(int argc, char **argv)
                       SECTOR_SIZE,
                       infinity,
                       h,
-                      closure(h, bread, fd, get_fs_offset(fd)),
+                      sg_wrapped_block_reader(closure(h, bread, fd, get_fs_offset(fd)), SECTOR_OFFSET, h),
                       closure(h, bwrite, fd),
```
francescolavra, line 149: This needs a call to `init_sg(h)`, otherwise sg_heap will be NULL and sg_wrapped_block_reader() will crash when trying to allocate a closure.

src/unix/mmap.c
wjhun: Yes, thanks. Fixed.

mkfs/dump.c
wjhun: Fixed

wjhun:
> These changes seem to have a major impact on the process startup time: when I run (via Qemu) a program with a manifest containing just the program executable (which is around 160 MB), the time it takes for the "assigned: 10.0.2.15" message to show up on the console is more than 30 seconds. With the master branch, this startup time is around 1 second.

A filesystem_read_entire of a large executable like that exposed the poor scaling of the rangemap. While that should change to a tree or something more efficient, right now changing the storage pagecache to use 2M pages appears to resolve the performance issue - and is perhaps preferable anyhow as less pages need to be tracked for a given amount of cached data. The scaling issue will still need to be resolved before moving to 4K pages, which will be necessary to support file backed mmaps.

On a different note, the sendfile(2) implementation has been updated to use the scatter gather read interface to tfs, enabling zero-copy reads. On my setup, the downloading bulk data via a go webserver and wget improves from ~26MB/s to 135MB/s (~5x improvement - some of that is likely a result of fixing sendfile to cap read lengths).

PR #1133, opened 2020-04-17
Filesystem persistence fix

francescolavra:
This PR fixes the updating of tuples in the filesystem log, which was breaking filesystem persistence.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -533,7 +533,18 @@ static void remove_extent_from_file(fsfile f, extent ex, merge m)
     string allocated = table_find(e, sym(allocated));
     assert(allocated);
     deallocate_buffer(allocated);
-    deallocate_tuple(e);
+
+    /* This tuple is not deallocated because it is already referenced in the
+     * filesystem log and thus present in the dictionary. To avoid this leakage,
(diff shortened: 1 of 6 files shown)


src/tfs/tfs.c
```diff
@@ -533,7 +533,7 @@ static void remove_extent_from_file(fsfile f, extent ex, merge m)
     string allocated = table_find(e, sym(allocated));
     assert(allocated);
     deallocate_buffer(allocated);
-    deallocate_tuple(e);
+    clear_tuple(e);
```
wjhun, line 536: I guess e just sticks around in f->md? Do these ever get cleaned up on file deletion or are they just leaked?

francescolavra: `e` refers to an extent that is no longer associated to the file, so there are no references to it in the file extent map, and as such it's not cleaned up on file deletion.
To avoid this leakage, we need additional functionalities in the filesystem log, e.g.:
- being able to remove a dictionary entry without affecting the keys associated to the other (present and future) entries
- being able to prune the log (and clean up the dictionary accordingly), to remove any log entries that are no longer relevant in the current status of the filesystem

I added these considerations as a comment in the code.

PR #1135, opened 2020-04-20
Fixes for futex wait timeout handling

francescolavra:
Besides fixing wait timeout handling, this PR adds support for absolute futex timeout values.

### DIFF
--- a/src/unix/futex.c
+++ b/src/unix/futex.c
@@ -103,22 +103,19 @@ closure_function(2, 1, sysreturn, futex_bh,
     thread t = bound(t);
     sysreturn rv;
 
-    if (current == t)
-        rv = BLOCKQ_BLOCK_REQUIRED;
-    else if (flags & BLOCKQ_ACTION_NULLIFY)
+    if (flags & BLOCKQ_ACTION_NULLIFY)
         rv = -EINTR;
     else if (flags & BLOCKQ_ACTION_TIMEDOUT)
         rv = -ETIMEDOUT;
-    else
+    else if (current == t) {
+        return BLOCKQ_BLOCK_REQUIRED;
+        thread_log(t, "%s: struct futex: %p, blocking\n", __func__, bound(f));
+    } else
         rv = 0; /* no timer expire + not us --> actual wakeup */
 
     thread_log(t, "%s: struct futex: %p, flags 0x%lx, rv %ld\n", __func__, bound(f), flags, rv);
-
-    if (rv != BLOCKQ_BLOCK_REQUIRED) {
-        thread_wakeup(t);
-        closure_finish();
-    }
-
+    thread_wakeup(t);
+    closure_finish();
     return set_syscall_return(t, rv);
 }
 
--- a/src/unix/futex.c
+++ b/src/unix/futex.c
@@ -148,6 +148,8 @@ sysreturn futex(int *uaddr, int futex_op, int val,
     
     op = futex_op & 127; // chuck the private bit
     ts = get_timeout_timestamp(op, val2);
+    clock_id clkid = (futex_op & FUTEX_CLOCK_REALTIME) ? CLOCK_ID_REALTIME :
+            CLOCK_ID_MONOTONIC;
 
     switch (op) {
     case FUTEX_WAIT: {
@@ -163,7 +165,7 @@ sysreturn futex(int *uaddr, int futex_op, int val,
 
         return blockq_check_timeout(f->bq, current, 
                                     closure(f->h, futex_bh, f, current),
-                                    false, CLOCK_ID_MONOTONIC, ts, false);
+                                    false, clkid, ts, false);
     }
 
     case FUTEX_WAKE: {
@@ -254,10 +256,9 @@ sysreturn futex(int *uaddr, int futex_op, int val,
             return set_syscall_error(current, EAGAIN);
 
         set_syscall_return(current, 0);
-        // TODO: timeout should be absolute based on CLOCK_REALTIME
         return blockq_check_timeout(f->bq, current, 
                                     closure(f->h, futex_bh, f, current),
-                                    false, CLOCK_ID_MONOTONIC, ts, false);
+                                    false, clkid, ts, true);
     }
 
     case FUTEX_REQUEUE: rprintf("futex_requeue not implemented\n"); break;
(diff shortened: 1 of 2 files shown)


src/unix/futex.c
```diff
@@ -103,22 +103,19 @@ closure_function(2, 1, sysreturn, futex_bh,
...
 
-    if (current == t)
-        rv = BLOCKQ_BLOCK_REQUIRED;
-    else if (flags & BLOCKQ_ACTION_NULLIFY)
+    if (flags & BLOCKQ_ACTION_NULLIFY)
         rv = -EINTR;
     else if (flags & BLOCKQ_ACTION_TIMEDOUT)
         rv = -ETIMEDOUT;
+    else if (current == t)
+        return BLOCKQ_BLOCK_REQUIRED;
```
wjhun, line 111: Maybe add a thread_log here before return as a debug aid?

src/unix/futex.c
```diff
@@ -257,10 +254,11 @@ sysreturn futex(int *uaddr, int futex_op, int val,
             return set_syscall_error(current, EAGAIN);
 
         set_syscall_return(current, 0);
-        // TODO: timeout should be absolute based on CLOCK_REALTIME
+        boolean abs_time = !!(futex_op & FUTEX_CLOCK_REALTIME);
         return blockq_check_timeout(f->bq, current, 
                                     closure(f->h, futex_bh, f, current),
-                                    false, CLOCK_ID_MONOTONIC, ts, false);
+                                    false, abs_time ? CLOCK_ID_REALTIME :
+                                    CLOCK_ID_MONOTONIC, ts, abs_time);
```
wjhun, line 261: The futex(2) man page says for FUTEX_WAIT_BITSET: 

> If timeout is not NULL, the structure it points to specifies an absolute timeout for the wait operation.  If timeout is NULL, the operation can block indefinitely.

and earlier for FUTEX_WAIT:

> Note:  for  FUTEX_WAIT,  timeout is interpreted as a relative value.  This differs from other futex operations, where timeout is interpreted as an absolute value.  To obtain the equivalent of FUTEX_WAIT with an absolute timeout, employ FUTEX_WAIT_BITSET with val3 specified as FUTEX_BITSET_MATCH_ANY.

It would seem from these statements that the timeout should always be absolute for FUTEX_WAIT_BITSET.

wjhun: This is probably optional as it's only since Linux 4.5, but the man page also indicates that FUTEX_WAIT can support FUTEX_CLOCK_REALTIME as well. Under section FUTEX_WAIT:

>  The  timeout  is  by default measured according to the CLOCK_MONOTONIC clock, but, since Linux 4.5, the CLOCK_REALTIME clock can be selected by specifying FUTEX_CLOCK_REALTIME in futex_op.

src/unix/futex.c
francescolavra: OK, done

src/unix/futex.c
francescolavra: The man page is wrong where it says that the timeout is relative for FUTEX_WAIT and absolute for FUTEX_WAIT_BITSET: this doesn't match what the same man page says where it describes the FUTEX_CLOCK_REALTIME flag:

> This option bit can be employed only with the FUTEX_WAIT_BITSET, FUTEX_WAIT_REQUEUE_PI, and (since Linux .5) FUTEX_WAIT operations.
> 
> If this option is set, the kernel measures the timeout against the CLOCK_REALTIME clock.
> 
> If this option is not set, the kernel measures the timeout against the CLOCK_MONOTONIC clock.

I looked at the Linux kernel implementation at https://elixir.bootlin.com/linux/latest/source/kernel/futex.c#L3868, and indeed it honors the FUTEX_CLOCK_REALTIME flag for both FUTEX_WAIT and FUTEX_WAIT_BITSET.
I modified the commit to add support for real-time clock also to FUTEX_WAIT.

wjhun: > The man page is wrong where it says that the timeout is relative for FUTEX_WAIT and absolute for FUTEX_WAIT_BITSET: this doesn't match what the same man page says where it describes the FUTEX_CLOCK_REALTIME flag:
> 
> > This option bit can be employed only with the FUTEX_WAIT_BITSET, FUTEX_WAIT_REQUEUE_PI, and (since Linux 4.5) FUTEX_WAIT operations.
> > If this option is set, the kernel measures the timeout against the CLOCK_REALTIME clock.
> > If this option is not set, the kernel measures the timeout against the CLOCK_MONOTONIC clock.

I don't see where these lines say anything about absolute or relative time, only monotonic or realtime. Unless I am mistaken, the clock type (monotonic / realtime) and absolute / relative are indepedent. Are you saying that CLOCK_REALTIME suggests absolute?

See the following in Linux kernel/futex.c:

```
SYSCALL_DEFINE6(futex, u32 __user *, uaddr, int, op, u32, val,
                struct timespec __user *, utime, u32 __user *, uaddr2,
                u32, val3)
{
        struct timespec ts;
        ktime_t t, *tp = NULL;
        u32 val2 = 0;
        int cmd = op & FUTEX_CMD_MASK;

        if (utime && (cmd == FUTEX_WAIT || cmd == FUTEX_LOCK_PI ||
                      cmd == FUTEX_WAIT_BITSET ||
                      cmd == FUTEX_WAIT_REQUEUE_PI)) {
                if (unlikely(should_fail_futex(!(op & FUTEX_PRIVATE_FLAG))))
                        return -EFAULT;
                if (copy_from_user(&ts, utime, sizeof(ts)) != 0)
                        return -EFAULT;
                if (!timespec_valid(&ts))
                        return -EINVAL;

                t = timespec_to_ktime(ts);
>>>             if (cmd == FUTEX_WAIT)
>>>                     t = ktime_add_safe(ktime_get(), t);
                tp = &t;
        }
        /*
         * requeue parameter in 'utime' if cmd == FUTEX_*_REQUEUE_*.
         * number of waiters to wake in 'utime' if cmd == FUTEX_WAKE_OP.
         */
        if (cmd == FUTEX_REQUEUE || cmd == FUTEX_CMP_REQUEUE ||
            cmd == FUTEX_CMP_REQUEUE_PI || cmd == FUTEX_WAKE_OP)
                val2 = (u32) (unsigned long) utime;

        return do_futex(uaddr, op, val, tp, uaddr2, val2, val3);
}
```

do_futex and subsequent calls appear to require an absolute time (the argument is called abs_time in futex_wait()). See the lines above with emphasis: it is adding ktime_get to the specified time when cmd == FUTEX_WAIT. This would suggest that the timeout is relative for FUTEX_WAIT and absolute otherwise.

PR #1137, opened 2020-04-21
misc fixes

wjhun:
various bugs exposed by mac build


test/runtime/aio.c
```diff
@@ -165,9 +165,14 @@ static void aio_test_multiple()
...
-        test_assert(evts[i].res == 8);
+    for (int i = 0; i < 18; i++) {
+        if (i < 8) {
+            test_assert(evts[i].data == (__u64) write_buf);
+            test_assert(evts[i].res == 8);
+        } else {
+            iocb_ptrs[i] = &iocbs[i];
+        }
+
         iocb_setup_pread(&iocbs[i], fd, read_buf + i * 8, 8, i * 8);
```
francescolavra, line 176: Using more than 8 iocb structs overruns `read_buf`, which is 256 bytes in total. So we cannot use 18 iocb structs; the real error is at the line `test_assert(syscall(SYS_io_submit, ioc, 18, iocb_ptrs) == 8);`, where 18 should be 8 instead.

src/unix/aio.c
```diff
@@ -225,7 +225,7 @@ static sysreturn iocb_enqueue(struct aio *aio, struct iocb *iocb)
 sysreturn io_submit(aio_context_t ctx_id, long nr, struct iocb **iocbpp)
 {
     struct aio *aio;
-    if (!ctx_id) {
+    if (!ctx_id || !iocbpp) {
```
francescolavra, line 228: Right, I missed this one (although this is not what caused the breakage).

test/runtime/aio.c
wjhun: Got it, thanks!

PR #1139, opened 2020-04-22
mkfs: add commmand line option to set minimum file size

### LINKED ISSUE #1136: mkfs flag for initial image sz at creation
mkfs should take an optional flag to set the initial size of a volume

"A flag to mkfs could essentially do an ftruncate internally after the image has been written (as long as the compiled image isn't greater than the given value, which I guess would give an error). Right now the image length is just set implicitly by tfs writing blocks ... the size just happens to be the end of the last block written."

francescolavra:
The 's' command line option can be used to specify a file size for the disk image being created. The size can be expressed in bytes, KB (with k or K suffix), MB (with m or M suffix), and GB (with g or G suffix); if not multiple of the disk block size, it is increased to the next block size multiple. If the specified size is less than the minimum size required to create the file system, this command line option is ignored.
This change allows creating disk images with an arbitrarily large free disk space.

Closes #1136.

### DIFF
--- a/mkfs/mkfs.c
+++ b/mkfs/mkfs.c
@@ -302,10 +302,14 @@ static void usage(const char *program_name)
 {
     const char *p = strrchr(program_name, '/');
     p = p != NULL ? p + 1 : program_name;
-    printf("Usage: %s [-b boot-image] [-r target-root] image-file < manifest-file\n"
+    printf("Usage: %s [-b boot-image] [-r target-root] [-s image-size] "
+           "image-file < manifest-file\n"
            "\n"
            "-b	- specify boot image to prepend\n"
-           "-r	- specify target root\n",
+           "-r	- specify target root\n"
+           "-s	- specify minimum image file size; can be expressed in bytes, "
+           "KB (with k or K suffix), MB (with m or M suffix), and GB (with g or"
+           " G suffix)\n",
            p);
 }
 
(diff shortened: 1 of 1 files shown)


mkfs/mkfs.c
```diff
@@ -302,10 +302,14 @@ static void usage(const char *program_name)
...
     const char *p = strrchr(program_name, '/');
     p = p != NULL ? p + 1 : program_name;
-    printf("Usage: %s [-b boot-image] [-r target-root] image-file < manifest-file\n"
+    printf("Usage: %s [-b boot-image] [-r target-root] [-s image-size] "
+           "image-file < manifest-file\n"
            "\n"
            "-b	- specify boot image to prepend\n"
-           "-r	- specify target root\n",
+           "-r	- specify target root\n"
+           "-s	- specify image file size; can be expressed in bytes, KB (with "
```
wjhun, line 310: May want to specify "minimum image file size", as the actual size could potentially be larger than the given size.

mkfs/mkfs.c
```diff
@@ -382,6 +418,17 @@ int main(int argc, char **argv)
...
+        off_t current_size = lseek(out, 0, SEEK_END);
+        if (current_size < 0) {
+            halt("could not get image size: %s\n", strerror(errno));
+        }
+        if (current_size < img_size) {
+            if (ftruncate(out, img_size)) {
+                halt("could not set image size: %s\n", strerror(errno));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
wjhun: Yes, that sounds right.

mkfs/mkfs.c
francescolavra: OK, done

Issue #1130, opened 2020-04-15
low memory handling


wjhun:
We need a heuristic and global facility for detecting low memory situations and triggering drain functions around the various caches and free lists that exist in the kernel (page cache, object caches, etc). This is open to interpretation / design.

Related: #1125, #1117, #1079, #660

PR #1140, opened 2020-04-22
page eviction and simple mm service routine

### LINKED ISSUE #1117: future work page cache / page reclaim/drain /

wjhun:
Add drain function to pagecache to evict cached pages. The mm_service routine, called from runloop_internal, simply checks if free physical memory dips below a threshold (CACHE_DRAIN_CUTOFF) and calls the pagecache drain with the delta.

Clearly more can be done to address low memory situations (#1130). Perhaps we could chain drain functions for various caches in the system (of varying types - including object caches) and drain according to some priority scheme. And a low memory threshold could be something more intelligent - a proportion of total memory or something determined dynamically. But a fixed threshold is better than not draining the pagecache at all.

There isn't a runtime test to stress the cache or page eviction. To test these changes, I raised CACHE_DRAIN_CUTOFF to various levels (up to system memory) to force drainage while running 'make test' and a go webserver serving large files.

Also fixes a missing sg_buf release in sendfile(2).

Resolves #1117, partially addresses #1130

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -498,6 +498,7 @@ closure_function(9, 2, void, sendfile_bh,
                 thread_log(t, "   rewound %ld bytes to %ld", rewind, f_in->offset);
             }
             rv = bound(written) == 0 ? -EAGAIN : bound(written);
+            sg_buf_release(bound(cur_buf));
             thread_log(t, "   write would block, returning %ld", rv);
         } else {
             thread_log(t, "   zero or error, rv %ld", rv);
@@ -523,7 +524,6 @@ closure_function(9, 2, void, sendfile_bh,
     } else {
         bound(written) += rv;
         bound(cur_buf)->misc += rv;
-        assert(bound(cur_buf)->length >= rv);
         if (bound(cur_buf)->misc == bound(cur_buf)->length) {
             sg_buf_release(bound(cur_buf));
             if (bound(written) == bound(readlen)) {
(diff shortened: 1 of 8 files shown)


src/unix/syscall.c
```diff
@@ -527,6 +529,7 @@ closure_function(9, 2, void, sendfile_bh,
         if (bound(cur_buf)->misc == bound(cur_buf)->length) {
             sg_buf_release(bound(cur_buf));
             if (bound(written) == bound(readlen)) {
+                bound(cur_buf) = 0;
                 rv = bound(written);
                 goto out_complete;
```
francescolavra, line 534: This doesn't seem necessary, since at this point we are about to exit from the syscall, so `cur_buf` will never be used again.

src/x86_64/service.c
```diff
@@ -103,6 +107,33 @@ closure_function(1, 2, void, fsstarted,
...
+#endif
+void mm_service(void)
+{
+    if (!global_pagecache)
+        return;
+    heap p = (heap)heap_physical(&heaps);
+    u64 free = heap_total(p) - heap_allocated(p);
+    mm_debug("%s: total %ld, alloc %ld, free %ld\n", __func__, heap_total(p), heap_allocated(p), free);
+    if (free < CACHE_DRAIN_CUTOFF) {
+        u64 drain_bytes = CACHE_DRAIN_CUTOFF - free + PAGESIZE - 1;
```
francescolavra, line 130: Shouldn't it be PAGESIZE_2M instead of PAGESIZE? Incidentally, I think this math should be done internally by pagecache_drain(), which should ensure that the number of page it drains (or tries to drain) covers at least `drain_bytes`.

wjhun: Agreed on both comments. Thanks.

PR #1147, opened 2020-04-29
move kernel to high address space

wjhun:
These changes relocate the kernel image to near the top of 64-bit address space, freeing up the lowest 128TB of space for the user process. As is typical of other OS ports to amd64, the -mcmodel=kernel flag is used in compilation to allow signed 32-bit relocations (versus the more expensive mcmodel=large). This requires the kernel to exist in the negative (top) 2gb, so the kernel base is now 0xffffffff80000000.

Moving the kernel and associated kernel memory in the upper canonical portion of address space also provides a clean delineation to allow trivial validation of whether a buffer address is in user or kernel memory. Convoluted logic used to accommodate the kernel living in 32-bit space has been terminated.

In the process of cleaning up fixed mappings in low (user) address space, the identity heap used for page table mappings, shared between stage2 and stage3, has been replaced with a small scratch region used for creating initial mappings before jumping to stage3. This region is mapped at a relocated address, and upon entry into stage3, the region extent is recorded to allow translation of physical pte addresses to this mapping.The page table code in stage3 allocates further mappings from the backed heap, and maintains a table of physical->virtual mappings of these allocations (carved out of 2M pages to keep the table size low) to allow page table traversal. This removes the fixed upper limit of page table allocations as well as the large allotment of physical memory reserved for said allocations.

The tag encoding offset is changed to 39 due to loss of high order bit. It may at some point make sense to expand tag encodings, but this gives an 8-bit initial selector which can be used to select a type or used as part of an expanded encoding. Incidentally, it also means that this selector is aligned with the top level page directory, and that mappings for each selector will be contained together in second level directory entries.

Parameterization of the pages heap spread throughout the kernel have been removed in favor of specifying it once on page table initialization. I couldn't think of any reason for wanting to draw from multiple heaps for page table construction.


boot/service32.s
```diff
@@ -415,13 +415,16 @@ GDT64:  ; Global Descriptor Table (64-bit).
         dw $ - GDT64 - 1    ; Limit.
         dw GDT64, 0         ; 64 bit Base.
 
-        
+[BITS 64]
```
francescolavra, line 418: For the record, this requires the assembler to support 64-bit code with elf32 output format, which is not the case with NASM version 2.11.08 and older. If you get "error: elf32 output format does not support 64-bit code", update NASM to a newer version (2.13.02 has been verified to work).

src/x86_64/page.c
```diff
@@ -243,53 +294,65 @@ static inline boolean map_page(page base, u64 v, physical p, heap h,
...
-         next ## level = pt_level_end(addr ## level, PT ## level),      \
-             end ## level = MIN(next ## level, end),                    \
-             pte ## level = ((u64*)base) + pindex(addr ## level, PT ## level), \
-             addr ## level < levelend;                                  \
-         addr ## level = next ## level)
+#ifdef TRAVERSE_PTES_DEBUG
+    rprintf("   pbase 0x%lx, level %d, shift %d, lsize 0x%lx, laddr 0x%lx,\n"
+            "      start_idx %ld, end_idx %ld, x 0x%lx, offset 0x%lx\n",
+            __func__, pbase, level, shift, lsize, laddr,
+            start_idx, end_idx, offset);
```
francescolavra, line 313: The format string and the other arguments don't match: `__func__` shouldn't be there, and `x` is missing.

boot/service32.s
wjhun: Thanks, I will add this note to Build Envs in the wiki. nasm 2.11.08 appears to be over four years old, so I think we should be safe here.

PR #1154, opened 2020-05-02
validation of user address arguments in syscalls

wjhun:
Now that there is a clean delineation between user and kernel addresses, we can do trivial range checks to validate that no kernel memory is referenced in a syscall. This adds functions to validate user buffers, including null-terminated constant strings and iovec arrays. A first pass has been made through the syscalls to insert such checks, which typically return -EFAULT on failure.

Also fix mistaken mapping of interpreter ELF in virtual_huge rather than p->virtual.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -262,11 +262,14 @@ static sysreturn mincore(void *addr, u64 length, u8 *vec)
     length = pad(length, PAGESIZE);
     nr_pgs = length >> PAGELOG;
 
+    if (!validate_user_memory(vec, nr_pgs, true))
+        return -EFAULT;
+
     /* -ENOMEM if any unmapped gaps in range */
     process p = current->p;
     vmap_lock(p);
     boolean found = rangemap_range_find_gaps(p->vmaps,
-                                             (range){start, start + length},
+                                             irange(start, start + length),
                                              stack_closure(mincore_vmap_gap));
     vmap_unlock(p);
     if (found)
@@ -420,17 +423,31 @@ closure_function(3, 1, void, vmap_attribute_update_intersection,
     }
 }
 
-static void vmap_attribute_update(heap h, rangemap pvmap, vmap q)
+closure_function(0, 1, void, vmap_attribute_update_gap,
+                 range, r)
+{
+    thread_log(current, "   found gap [0x%lx, 0x%lx)", r.start, r.end);
+}
+
+static boolean vmap_attribute_update(heap h, rangemap pvmap, vmap q)
 {
     range rq = q->node.r;
     assert((rq.start & MASK(PAGELOG)) == 0);
     assert((rq.end & MASK(PAGELOG)) == 0);
     assert(range_span(rq) > 0);
 
+    thread_log(current, "%s: validate %R", __func__, rq);
+    if (!validate_user_memory(pointer_from_u64(rq.start), range_span(rq), false) ||
+        rangemap_range_find_gaps(pvmap, rq, stack_closure(vmap_attribute_update_gap))) {
+        return false;
+    }
+
     rmnode_handler nh = stack_closure(vmap_attribute_update_intersection, h, pvmap, q);
-    rangemap_range_lookup(pvmap, rq, nh);
+    if (!rangemap_range_lookup(pvmap, rq, nh))
+        return false;
 
     update_map_flags(rq.start, range_span(rq), page_map_flags(q->flags));
+    return true;
 }
 
 sysreturn mprotect(void * addr, u64 len, int prot)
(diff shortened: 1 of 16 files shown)


src/unix/syscall.c
```diff
@@ -1433,6 +1497,9 @@ sysreturn getrlimit(int resource, struct rlimit *rlim)
 {
     thread_log(current, "getrlimit: resource %d, rlim %p", resource, rlim);
 
+    if (rlim && !validate_user_memory(rlim, sizeof(struct rlimit), true))
+        return -EFAULT;
```
francescolavra, line 1501: A NULL rlim pointer is invalid and shouldn't be allowed, so I would call validate_user_memory() unconditionally, and remove `if (!rlim) return set_syscall_error(current, EINVAL);` from each of the `case` clauses below.

src/unix/syscall.c
```diff
@@ -1931,6 +2025,8 @@ sysreturn sched_getaffinity(int pid, u64 cpusetsize, cpu_set_t *mask)
 sysreturn capget(cap_user_header_t hdrp, cap_user_data_t datap)
 {
     if (datap) {
+        if (!validate_user_memory(datap, sizeof(cap_user_data_t), true))
```
francescolavra, line 2028: cap_user_data_t is a pointer type, we should use either `struct user_cap_data` or `sizeof(*datap)`.

src/unix/thread.c
```diff
@@ -10,13 +10,25 @@ sysreturn gettid()
 
 sysreturn set_tid_address(int *a)
 {
+    /* man page says this always succeeds, but... */
+    if (!validate_user_memory(a, sizeof(int), true))
+        return set_syscall_error(current, EFAULT);
```
francescolavra, line 15: A NULL value for `a` is valid and shouldn't trigger an EFAULT.

src/unix/thread.c
```diff
@@ -10,13 +10,25 @@ sysreturn gettid()
...
     current->clear_tid = a;
     return current->tid;
 }
 
 sysreturn arch_prctl(int code, unsigned long addr)
 {    
     thread_log(current, "arch_prctl: code 0x%x, addr 0x%lx", code, addr);
+    if ((code == ARCH_GET_FS || code == ARCH_GET_GS) &&
+        !validate_user_memory((void *)addr, sizeof(u64), true))
+        return set_syscall_error(current, EFAULT);
```
francescolavra, line 25: Now that we have these checks, the code `if (!addr) return set_syscall_error(current, EINVAL);` should be removed from the ARCH_GET_FS and ARCH_GET_GS cases below.

src/unix/thread.c
```diff
@@ -50,6 +62,16 @@ sysreturn clone(unsigned long flags, void *child_stack, int *ptid, int *ctid, un
         return set_syscall_error(current, ENOSYS);
     }
 
+    /* no stack size given, just validate the top word */
+    if (!validate_user_memory(child_stack, sizeof(u64), true))
+        return set_syscall_error(current, EFAULT);
```
francescolavra, line 67: As far as I can see, the kernel will never access this stack, so I think it shouldn't bother validating it.

src/unix/unix_clock.c
```diff
@@ -77,6 +85,8 @@ sysreturn clock_nanosleep(clockid_t _clock_id, int flags, const struct timespec
 
 sysreturn sys_time(time_t *tloc)
 {
+    if (!validate_user_memory(tloc, sizeof(time_t), true))
+        return -EFAULT;
```
francescolavra, line 89: A NULL tloc value is valid and shouldn't trigger an EFAULT.

src/unix/mmap.c
```diff
@@ -259,14 +259,17 @@ static sysreturn mincore(void *addr, u64 length, u8 *vec)
     if (start & MASK(PAGELOG))
         return -EINVAL;
 
+    if (!validate_user_memory(addr, (length + MASK(PAGELOG)) >> PAGELOG, true))
+        return -EFAULT;
+
     length = pad(length, PAGESIZE);
```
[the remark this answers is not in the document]
wjhun: Oops. That was meant to be a validate for vec. (and yes, nr_pgs is just below)

src/unix/thread.c
wjhun: I'm not sure if there's any reason for the kernel to access the user stack. I can't think of one either...but I'd still prefer a little paranoia to mask any unintended affects from a bad address getting loaded into %rsp and then a return to the frame. I can't see how the validate could hurt.

wjhun:
Changes applied. Thanks for the thorough review. I'll give it another pass before merging.

francescolavra:
LGTM, thanks.

PR #1160, opened 2020-05-04
fcntl: add support for F_SETPIPE_SZ and F_SETPIPE_SZ commands

francescolavra:
These commands allow to set and get the capacity of a pipe.

### DIFF
--- a/src/runtime/buffer.h
+++ b/src/runtime/buffer.h
@@ -63,20 +63,31 @@ static inline boolean buffer_is_wrapped(buffer b)
     return b->wrapped;
 }
 
+static inline bytes buffer_set_capacity(buffer b, bytes len)
+{
+    assert(!buffer_is_wrapped(b));  /* wrapped buffers can't be resized */
+    if (len < b->end - b->start)
+        len = b->end - b->start;
+    if (len != b->length) {
+        void *new = allocate(b->h, len);
+        if (new == INVALID_ADDRESS)
+            return b->length;
+        runtime_memcpy(new, b->contents + b->start, b->end - b->start);
+        deallocate(b->h, b->contents, b->length);
+        b->length = len;
+        b->end = b->end - b->start;
+        b->start = 0;
+        b->contents = new;
+    }
+    return len;
+}
+
 static inline void buffer_extend(buffer b, bytes len)
 {
     // xxx - pad to pagesize
     if (b->length < (b->end + len)) {
-        assert(!buffer_is_wrapped(b));    /* wrapped buffers can't be extended */
-        int oldlen = b->length;
-        b->length = 2*((b->end-b->start)+len);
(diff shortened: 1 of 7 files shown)


src/runtime/buffer.h
```diff
@@ -92,6 +92,25 @@ static inline void extend_total(buffer b, int offset)
     }
 }
 
+static inline bytes buffer_set_capacity(buffer b, bytes len)
+{
+    assert(!buffer_is_wrapped(b));
+    if (len < b->end - b->start)
+        len = b->end - b->start;
+    if (len != b->length) {
+        void *new = allocate(b->h, len);
```
wjhun, line 101: This block is nearly identical to one in buffer_extend...maybe this can be unified / refactored? Maybe buffer_extend and extend_total can use buffer_set_capacity internally. (That interface is confusing to begin with...extend_total should probably be renamed.)

src/unix/pipe.c
```diff
@@ -361,3 +361,21 @@ int do_pipe2(int fds[2], int flags)
     pipe_release(pipe);
     return -ENOMEM;
 }
+
+int pipe_set_capacity(fdesc f, int capacity)
+{
+    pipe_file pf = (pipe_file)f;
+    pipe p = pf->pipe;
+    if (capacity < INITIAL_PIPE_DATA_SIZE)
```
wjhun, line 369: Any idea where this INITIAL_PIPE_DATA_SIZE value of 100 comes from? Seems from pipe(7) that the lower limit should be PAGESIZE. Also, do we want to place an upper bound here? (man page claims default for /proc/sys/fs/pipe-max-size is 1MB).

francescolavra: The INITIAL_PIPE_DATA_SIZE value of 100 has been there since the initial pipe implementation, I would guess it's just a made up number. I'm OK with setting the lower limit to PAGESIZE, but as far as an upper limit is concerned, the man pages say that the /proc/sys/fs/pipe-max-size limit applies only to unprivileged processes, while a privileged process (CAP_SYS_RESOURCE) can override the limit. So there is no really upper bound for privileged processes, and I think the process running on a unikernel should be as privileged as it can get :-)

src/runtime/buffer.h
francescolavra: OK, I refactored buffer_extend() to use buffer_set_capacity() internally. I didn't change extend_total() because the most "logical" way for extend_total() to call buffer_set_capacity() would be as in `buffer_set_capacity(b, offset)`, but with that change the vector unit tests take forever to complete, because the buffer of a vector gets resized at each vector element insertion; and there might be places in the kernel as well that take advantage of the resize logic that is in buffer_extend(). So I think it's better for extend_total() to keep calling buffer_extend().

PR #1182, opened 2020-05-25
Initial implementation of VMware pvscsi driver

mkhon:


src/vmware/pvscsi.c
```diff
@@ -0,0 +1,664 @@
...
+
+    pci_bar_write_4(&dev->bar, PVSCSI_REG_OFFSET_COMMAND, cmd);
+    for (int i = 0; i < len; ++i) {
+        pci_bar_write_4(&dev->bar, PVSCSI_REG_OFFSET_COMMAND_DATA, data_ptr[i]);
+    }
+}
+
+static void pvscsi_hcb_dealloc(pvscsi dev, struct pvscsi_hcb *hcb)
+{
+    u64 flags = spin_lock_irq(&dev->mem_lock);
```
wjhun, line 85: Does this need to be irq-safe? It seems like the int handler wouldn't touch the hcb cache.

src/vmware/pvscsi.c
```diff
@@ -0,0 +1,664 @@
...
+            assert(i);
+            struct pvscsi_hcb *hcb = struct_from_list(i, struct pvscsi_hcb *, links);
+            assert(hcb);
+            list_delete(i);
+            apply(hcb->completion);
+            pvscsi_hcb_dealloc(dev, hcb);
+        }
+    }
+
+    u64 flags = spin_lock_irq(&dev->queue_lock);
```
wjhun, line 434: Same here...would anything at interrupt level be touching dev->hcb_queue?

PR #1187, opened 2020-05-26
Network interfaces: add loopback interface

francescolavra:
This is needed by the Erlang runtime system, which when started in "distributed" mode spawns the Erlang process mapper daemon and then communicates with it via the loopback interface. This also prevents a "Failed to find the loopback interface" NetUtil warning from OpenJDK.
The loopback interface is assigned index number 0 in the network interface list, thus the "real" network interface is assigned index number 1 and then named "en1".
In order to receive packets from the loopback interface, netif_poll() must be called by the kernel, and this is achieved via the netsock_poll() thunk, which is put in the runqueue and then invoked by the scheduler.

### DIFF
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -149,7 +149,7 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
 }
 
 void init_network_iface(tuple root) {
-    struct netif *n = netif_find("en0");
+    struct netif *n = netif_find("en1");
     if (!n) {
         rprintf("no network interface found\n");
         return;
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -148,6 +148,24 @@ static sysreturn netsock_sendto(struct sock *sock, void *buf, u64 len,
 static sysreturn netsock_recvfrom(struct sock *sock, void *buf, u64 len,
         int flags, struct sockaddr *src_addr, socklen_t *addrlen);
 
+static thunk net_loop_poll;
+static boolean net_loop_poll_queued;
+
+closure_function(0, 0, void, netsock_poll) {
+    net_loop_poll_queued = false;
+    netif_poll_all();
+}
+
+static void netsock_check_loop(void)
+{
+    /* Not race-free, but the worst that can happen is that the thunk is
+     * enqueued more than once. */
+    if (!net_loop_poll_queued) {
+        net_loop_poll_queued = true;
+        enqueue(runqueue, net_loop_poll);
+    }
+}
+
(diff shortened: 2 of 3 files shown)


src/net/netsyscall.c
```diff
@@ -1702,6 +1702,11 @@ void register_net_syscalls(struct syscall *map)
     register_syscall(map, shutdown, shutdown);
 }
 
+closure_function(0, 0, void, netsock_poll) {
+    netif_poll_all();
```
wjhun, line 1706: I know the comment above netif_poll states that it should be called in a main loop, but we may want to question if this is absolute necessary, as it means running this routine every time through runloop (when kernel lock is acquired). (And it's really a flaw in runloop that the CPU can go idle while something is sitting on a runqueue...if that was corrected this would run endlessly and never sleep.)

Is it only needed for loopback function and, if so, could it be triggered by net syscalls instead? Or perhaps run from a periodic timer if the loopback interface is actively in use (has buffers queued)?

src/net/net.c
```diff
@@ -149,7 +149,7 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
 }
 
 void init_network_iface(tuple root) {
-    struct netif *n = netif_find("en0");
+    struct netif *n = netif_find("en1");
```
wjhun, line 152: It doesn't matter much at this point, but could this be renamed to lo0, just to differentiate it from additional ethernet interfaces we might want to add in the future?

wjhun: Correction: I meant to ask if this could stay as en0. Isn't the loopback interface already lo0? (see netif_loopif_init() in lwip/src/core/netif.c)

francescolavra: Interface names in lwIP are composed of a two-letter identifier followed by a number, and the number corresponds to the 0-based index of the interface in the global interface list. The loopback interface, if present, is always added first (it is added from `netif_init()`), so the next interface must have index number 1 and so if its identifier is "en" it must be called "en1". Any additional ethernet interfaces we might add in the future will be named, according to the order in which they are added, "en2", "en3", etc.

src/net/netsyscall.c
francescolavra: I agree it's better not to rely on the fact that the scheduler doesn't immediately serve re-queued items in the runqueue. Yes, calling netif_poll() is only needed for loopback function, and as such it can be triggered by anything that sends a network packet. How about the updated code, where the thunk is put in the runqueue only when data is sent on a socket (and in case of TCP when a connection/disconnection is done)?

PR #1195, opened 2020-06-10
syscalls: add io_uring support

francescolavra:
Supported operations are nop, readv, writev, read_fixed, write_fixed, poll_add, poll_remove, timeout, timeout_remove, close, files_update, read, and write.
A new vmap flag, VMAP_FLAG_PREALLOC, is being defined in order to support mapping and unmapping pre-allocated memory regions.

### DIFF
--- a/src/runtime/list.h
+++ b/src/runtime/list.h
@@ -72,3 +72,26 @@ static inline struct list *list_pop_back(struct list *list)
     list_delete(back);
     return back;
 }
+
+static inline boolean list_find(struct list *head, struct list *elem)
+{
+    list_foreach(head, e) {
+        if (e == elem)
+            return true;
+    }
+    return false;
+}
+
+/* Move all elements from source list to destination list. */
+static inline void list_move(struct list *dest, struct list *src)
+{
+    if (list_empty(src)) {
+        list_init(dest);
+        return;
+    }
+    dest->next = src->next;
+    dest->next->prev = dest;
+    dest->prev = src->prev;
+    dest->prev->next = dest;
+    list_init(src); /* make it an empty list */
+}
--- a/src/unix/io_uring.c
+++ b/src/unix/io_uring.c
@@ -1,0 +1,1335 @@
+#include <unix_internal.h>
+#include <page.h>
+
+#define IORING_SETUP_CQSIZE     (1 << 3)
+
+#define IORING_FEAT_SINGLE_MMAP     (1 << 0)
+#define IORING_FEAT_RW_CUR_POS      (1 << 3)
+
+#define IORING_OFF_SQ_RING  0ULL
+#define IORING_OFF_CQ_RING  0x8000000ULL
+#define IORING_OFF_SQES     0x10000000ULL
+
+#define IORING_TIMEOUT_ABS  (1 << 0)
+
+#define IORING_ENTER_GETEVENTS  (1 << 0)
+
+#define IO_URING_OP_SUPPORTED   (1 << 0)
+
+#define IOUR_SQ_ENTRIES_MAX 0x40000000UL
+#define IOUR_CQ_ENTRIES_MAX (2 * IOUR_SQ_ENTRIES_MAX)
+#define IOUR_FILES_MAX      0x8000
+
+#define IOSQE_FIXED_FILE    (1 << 0)
+#define IOSQE_ASYNC         (1 << 4)
+
+//#define IOUR_DEBUG
+#ifdef IOUR_DEBUG
+#define iour_debug(x, ...) do { \
+    rprintf("%s: " x "\n", __func__, ##__VA_ARGS__); \
+} while(0)
+#else
+#define iour_debug(x, ...)
+#endif
+
+struct io_uring_sqe {
+    u8 opcode;
+    u8 flags;
+    u16 ioprio;
+    s32 fd;
+    u64 off;
+    u64 addr;
+    u32 len;
+    union {
+        u32 rw_flags;
+        u32 fsync_flags;
+        u16 poll_events;
+        u32 sync_range_flags;
+        u32 msg_flags;
+        u32 timeout_flags;
+    };
+    u64 user_data;
+    union{
+        u16 buf_index;
+        u64 __pad2[3];
+    };
+};
+
+struct io_uring_cqe {
+    u64 user_data;
+    s32 res;
+    u32 flags;
+};
+
+enum iour_enter_opcode {
+    IORING_OP_NOP,
+    IORING_OP_READV,
+    IORING_OP_WRITEV,
+    IORING_OP_FSYNC,
+    IORING_OP_READ_FIXED,
+    IORING_OP_WRITE_FIXED,
+    IORING_OP_POLL_ADD,
+    IORING_OP_POLL_REMOVE,
+    IORING_OP_SYNC_FILE_RANGE,
+    IORING_OP_SENDMSG,
+    IORING_OP_RECVMSG,
+    IORING_OP_TIMEOUT,
+    IORING_OP_TIMEOUT_REMOVE,
+    IORING_OP_ACCEPT,
+    IORING_OP_ASYNC_CANCEL,
+    IORING_OP_LINK_TIMEOUT,
+    IORING_OP_CONNECT,
+    IORING_OP_FALLOCATE,
+    IORING_OP_OPENAT,
+    IORING_OP_CLOSE,
+    IORING_OP_FILES_UPDATE,
+    IORING_OP_STATX,
+    IORING_OP_READ,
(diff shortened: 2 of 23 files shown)


src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+    map(virt, physical_from_virtual(iour->rings) + region_offset, len,
+        mapflags);
+    return virt;
+}
+
+simple_closure_function(1, 2, void, iour_efd_complete,
+                        u64, efd_val,
+                        thread, t, sysreturn, rv)
+{
+    closure_finish();
```
wjhun, line 475: Is this a placeholder function? Just trying to understand the closure_new / closure_get scheme ... where is the closure getting filled (and the function)?

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+        rings->cq_overflow++;
+    }
+    if (iour->eventfd && (async || !iour->eventfd_async)) {
+        closure_new(iour->h, iour_efd_complete, completion);
+        if (completion != INVALID_ADDRESS) {
+            completion->efd_val = 1;
+            apply(iour->eventfd->write, &completion->efd_val,
+                  sizeof(completion->efd_val), 0, current, true,
+                  closure_get(iour_efd_complete, completion));
+        }
```
wjhun, line 502: assert or otherwise handle alloc error here?

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+    case IORING_OP_READV:
+    case IORING_OP_WRITEV: {
+        if (sqe->buf_index) {
+            res = -EINVAL;
+            goto complete;
+        }
+        struct iovec *iov = pointer_from_u64(sqe->addr);
+        u32 len = sqe->len;
+        boolean write = (sqe->opcode == IORING_OP_WRITEV);
+        if (!validate_iovec(iov, len, write)) {
```
wjhun, line 870: should be !write

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+            void *buf = pointer_from_u64(sqe->addr);
+            u32 len = sqe->len;
+            boolean write = sqe->opcode == IORING_OP_WRITE_FIXED;
+
+            /* If we have the capability to pin memory pages, as done in Linux
+             * with pin_user_pages(), we will validate the memory at buffer
+             * registration time and avoid validating it here. */
+            if ((buf < iov->iov_base) || (u64_from_pointer(buf) + len >
+                    u64_from_pointer(iov->iov_base) + iov->iov_len) ||
+                    !validate_user_memory(buf, len, write)) {
```
wjhun, line 898: !write

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+    case IORING_OP_WRITE:
+        if (sqe->buf_index) {
+            res = -EINVAL;
+            goto complete;
+        } else {
+            void *buf = pointer_from_u64(sqe->addr);
+            u32 len = sqe->len;
+            boolean write = sqe->opcode == IORING_OP_WRITE;
+
+            if (!validate_user_memory(buf, len, write)) {
```
wjhun, line 989: !write

src/runtime/list.h
```diff
@@ -72,3 +72,26 @@ static inline struct list *list_pop_back(struct list *list)
...
+        if (e == elem)
+            return true;
+    }
+    return false;
+}
+
+/* Move all elements from source list to destination list. */
+static inline void list_move(struct list *dest, struct list *src)
+{
+    if (list_empty(src)) {
```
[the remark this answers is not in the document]
francescolavra: I would prefer this not to assert, I'm actually calling list_move() with an uninitialized dest, it seems a bit inconvenient to have to make the dest list empty only to have it initialized again in list_move().

src/unix/unix_internal.h
```diff
@@ -417,6 +422,23 @@ static inline int fdesc_type(fdesc f)
     return f->type;
 }
 
+static inline fdesc fdesc_get(process p, int fd)
+{
+    /* XXX To ensure atomicity, we need a mutex that protects against concurrent
```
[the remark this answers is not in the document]
francescolavra: Yes, I left this comment as a kind of reminder if we move to an implementation that allows concurrent execution of syscalls or stuff like that.

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+struct io_uring_probe {
+    u8 last_op;
+    u8 ops_len;
+    u16 resv;
+    u32 resv2[3];
+    struct io_uring_probe_op ops[0];
+};
+
+typedef struct io_rings {
+    unsigned int sq_head, sq_tail;
```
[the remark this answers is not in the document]
francescolavra: OK, I changed unsigned ints to u32.

src/unix/io_uring.c
```diff
@@ -0,0 +1,1340 @@
...
+                res = -EFAULT;
+        } else {
+            struct iovec *iov = &iour->bufs[buf_index];
+            void *buf = pointer_from_u64(sqe->addr);
+            u32 len = sqe->len;
+            boolean write = sqe->opcode == IORING_OP_WRITE_FIXED;
+
+            /* If we have the capability to pin memory pages, as done in Linux
+             * with pin_user_pages(), we will validate the memory at buffer
+             * registration time and avoid validating it here. */
```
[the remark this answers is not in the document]
francescolavra: If I understood it correctly, page pinning in general prevents a given memory range from being deallocated for a given process. If we don't have that, an application could register a set of buffers with io_uring, then free (unmap) that memory, then request I/O on that memory, at which point a page fault would happen (in kernel mode) and the relevant address wouldn't be found in the process vmap, i.e. the kernel wouldn't be able to handle the fault.

src/unix/io_uring.c
francescolavra: Here I'm just using the closure to piggyback the buffer (efd_val) passed to the eventfd write(). So instead of allocating the buffer and the completion closure separately, I'm doing everything with one allocation only. (Alternatively I could allocate just the buffer and use io_completion_ignore as completion.) The closure is being filled in the line `completion->efd_val = 1;`.

src/unix/io_uring.c
francescolavra: The allocation is actually done by `closure_new()` (after which I check for `completion != INVALID_ADDRESS`, i.e. check against allocation failure). `closure_get()` is just synctatic sugar to retrieve the apply() function from the closure name.

src/unix/io_uring.c
francescolavra: Right. Also found a mistake in `io_uring_enter()`, when validating `sig`, it should be `validate_user_memory(sig, sizeof(*sig), false)`.

src/runtime/list.h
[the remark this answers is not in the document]
wjhun: Right...this makes sense.

src/unix/io_uring.c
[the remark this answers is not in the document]
wjhun: I think the pinning is to take a reference for any physical pages that are behind the mapping at the time of the call, but to my knowledge it doesn't guarantee anything about the v->p mapping of those pages or whether a virtual address range remains mapped or consistent. But maybe it does prevent those pages from being drained from the page cache or swapped out (provided the process isn't doing something to remove a mapping). I'm not clear on that detail.

But it may still be relevant here for any calls which do zero-copy (or less-copy?) directly from user buffers - those indeed will need to have pages pinned. Our equivalent - at this time anyway - is to add a reference to the pagecache_page so that the (physical) page cannot be drained from the cache or otherwise freed. (We maintain those references in the cache, for we don't have anything like the struct page array in Linux.) But that still won't prevent the process from unmapping pages or otherwise doing something that might cause a fault on translation.

src/unix/io_uring.c
wjhun: Ok. I have at times thought it might be useful to be able to change the environment of a closure after it is allocated, prior to being applied, and these macros allow that to happen.

src/unix/io_uring.c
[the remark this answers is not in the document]
francescolavra: OK, got it, I changed the code so that it does iovec validation when buffers are registered, instead of validating the memory at each READ_FIXED/WRITE_FIXED operation.
The io_uring_register man page in the IORING_REGISTER_BUFFERS description says "The buffers associated with the iovecs will be locked in memory", so it looks like Linux does prevent swapping out relevant pages. But I don't see any I/O ops (among those that can be triggered via io_uring) that _require_ pages to be pinned, so I guess we can leave it as it is for now.

PR #1196, opened 2020-06-12
Filesystem dump: add option to display filesystem as a tree

### LINKED ISSUE #564: fs explorer
this really has nothing to do w/nanos proper but it'd be nice to have a simple little tool that you could see a filesystem tree after you've ran the image

we have the manifest of the filesystem *before* boot but after you run it and shut it down there could easily be new files, folders, etc. that the program created and there's not an easy way of seeing that information

I'm not looking for a ```mount -t tfs``` (although that'd be cool) but if it's easy enough to run like a ```tree``` on an image or something like that is what I'd be looking for - i suppose if we had support for multiple images - the ability to load one tfs on one drive and then you could load an explorer as the kernel to read what's in there but that prob is even more work than doing mount

francescolavra:
The mkfs/dump tool now accepts the '-t' command line option, which makes the tool display the filesystem from a given image as a tree. In the displayed tree, directory entries are blue, symbolic links are cyan (with indication of the link target path), and regular files are white.

The existing functionality to dump filesystem contents into a given directory can now be invoked via the '-d <target dir>' command line option.

Closes #564.

### DIFF
--- a/mkfs/dump.c
+++ b/mkfs/dump.c
@@ -11,6 +11,12 @@
 #include <string.h>
 #include <limits.h>
 
+#define DUMP_OPT_TREE  (1U << 0)
+
+#define TERM_COLOR_BLUE     94
+#define TERM_COLOR_CYAN     96
+#define TERM_COLOR_WHITE    97
+
 closure_function(1, 3, void, bwrite,
                  descriptor, d,
                  void *, s, range, blocks, status_handler, c)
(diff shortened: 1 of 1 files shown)


mkfs/dump.c
```diff
@@ -123,23 +170,54 @@ static u64 get_fs_offset(descriptor fd)
...
-        fprintf(stderr, "couldn't open file %s: %s\n", argv[1], strerror(errno));
+        fprintf(stderr, "couldn't open file %s: %s\n", argv[optind],
+            strerror(errno));
         exit(EXIT_FAILURE);
     }
 
     heap h = init_process_runtime();
     tuple root = allocate_tuple();
+    filesystem_complete completion;
+    if (target_dir)
```
wjhun, line 216: Do these options need to be mutually exclusive? If -d is specified, a -t will silently have no effect.

mkfs/dump.c
```diff
@@ -98,6 +102,49 @@ closure_function(3, 2, void, fsc,
...
+}
+
+static void dump_fsentry(int indent, symbol name, tuple t)
+{
+    table c;
+    buffer target_buf;
+    if ((c = children(t))) {
+        print_colored(indent, TERM_COLOR_BLUE, name, true);
+        table_foreach((tuple)c, k, vc) {
+            if (k == sym_this(".") || k == sym_this(".."))
```
wjhun, line 124: Not in the scope of this PR, but it would be really nice if we could do away with these links. I realize the ascending link obviates the need for a traversal, but it just seems icky to have these hanging around at every node - and persistently on storage.

mkfs/dump.c
francescolavra: Indeed the two options don't have to be mutually exclusive. I changed the code so that it always uses the fsc closure, which now takes an additional parameter `options`, and does everything that has been specified from the command line options.

mkfs/dump.c
francescolavra: Probably it won't make this any less icky, but the . and .. directories entries are not persisted on storage (see `fs_set_dir_entry()` in tfs.c), they are created at runtime when the filesystem is read, by the `fixup_directory()` function.

PR #1198, opened 2020-06-12
mkdir/mkdirat: fix directory creation when path ends with '/'

francescolavra:
mkdir() and mkdirat() should detect if the supplied path ends with the '/' delimiter, and strip the trailing delimiter if found. The newly defined dirname_from_path() function implements this.

The mkdir suite of runtime tests has been modified so that it would fail if the above was not implemented correctly.

### DIFF
--- a/src/runtime/text.h
+++ b/src/runtime/text.h
@@ -146,14 +146,17 @@ static inline const u8 *utf8_find(const u8 *x, character c)
     return false;
 }
 
-static inline const u8 *utf8_find_r(const u8 *x, character c)
+static inline const u8 *utf8_findn_r(const u8 *x, bytes n, character c)
 {
+    if (!x)
+        return false;
     int nbytes;
+    bytes offset = 0;
     const u8 *found = false;
 
-    while (x && *x) {
-        if (utf8_decode(x, &nbytes) == c) found = x;
-        x += nbytes;
+    while ((offset < n) && *(x + offset)) {
+        if (utf8_decode(x + offset, &nbytes) == c) found = x + offset;
+        offset += nbytes;
     }
     return found;
 }
--- a/src/unix/filesystem.h
+++ b/src/unix/filesystem.h
@@ -57,6 +57,36 @@ static inline const char *filename_from_path(const char *path)
     return filename;
 }
 
+/* Expects an empty buffer, and never resizes the buffer. */
+static inline boolean dirname_from_path(buffer dest, const char *path)
+{
+    int pathlen = runtime_strlen(path);
+    const char *last_delim = path_find_last_delim(path, PATH_MAX);
+    const char *dirname;
+    int len;
+    if (!last_delim) {
+        dirname = path;
+        len = pathlen;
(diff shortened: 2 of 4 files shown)


src/runtime/text.h
```diff
@@ -157,3 +157,18 @@ static inline const u8 *utf8_find_r(const u8 *x, character c)
     }
     return found;
 }
+
+static inline const u8 *utf8_findn_r(const u8 *x, bytes n, character c)
```
wjhun, line 161: It would be nice to just get rid of the unbounded version of this. PATH_MAX could be used as a limit in absence of an actual string length.

src/unix/filesystem.h
```diff
@@ -52,6 +52,35 @@ static inline const char *filename_from_path(const char *path)
...
+    int pathlen = runtime_strlen(path);
+    const char *last_delim = (char *)utf8_find_r((u8 *)path, '/');
+    const char *dirname;
+    int len;
+    if (!last_delim) {
+        dirname = path;
+        len = pathlen;
+    } else if (last_delim < path + pathlen - 1) {
+        dirname = last_delim + 1;
+        len = pathlen - (last_delim + 1 - path);
```
wjhun, line 67: A total nit, but "len = pathlen - (dirname - path)" seems clearer.

src/unix/filesystem.h
```diff
@@ -52,6 +52,35 @@ static inline const char *filename_from_path(const char *path)
...
+    } else {    /* The path ends with '/'. */
+        const char *delim = (char *)utf8_findn_r((u8 *)path, last_delim - path,
+            '/');
+        if (!delim) {
```
wjhun, line 71: Another nit: this block is a little repetitious and could be refactored with a static inline.

src/unix/filesystem.h
```diff
@@ -52,6 +52,35 @@ static inline const char *filename_from_path(const char *path)
...
+            dirname = path;
+            len = pathlen - 1;
+        } else {
+            dirname = delim + 1;
+            len = last_delim - (delim + 1);
+        }
+    }
+    if (len < dest->length)
```
wjhun, line 79: Nit # 3: Maybe just return false if the dirname won't fit instead of doing the peek_char?

src/runtime/text.h
francescolavra: OK, I split this PR into 2 commits, and the first commits gets rid of the unbounded function.

src/unix/filesystem.h
francescolavra: Agreed. Also changed `last_delim - (delim + 1)` to `last_delim - dirname` a few lines below.

src/unix/filesystem.h
francescolavra: I created the `path_find_last_delim()` static inline, also used in `filename_from_path()`.

src/unix/filesystem.h
francescolavra: Yes, it looks better.

PR #1203, opened 2020-06-17
Kernel ELF file: move from root filesystem to boot filesystem

francescolavra:
This PR adds a new partition, containing the "boot filesystem", and moves the kernel file from the existing root filesystem to the new boot filesystem. This prevents the user process from accessing (and possibly overwriting) the kernel file.
As a side effect, stage2, which now loads the boot filesystem, no longer needs to load metadata for the entire root filesystem, which was unnecessary overhead at boot time.
The location of the root filesystem is no longer derived from the REGION_FILESYSTEM memory region, but is found from the MBR partition table.
The contents of the boot filesystem are specified in the manifest file via the "boot" tuple. If such tuple is not found in a given manifest, for backward compatibility the mkfs tool retrieves the kernel from the root tuple.

### DIFF
--- a/mkfs/mkfs.c
+++ b/mkfs/mkfs.c
@@ -12,6 +12,7 @@
 #include <limits.h>
 
 #include <region.h>
+#include <storage.h>
 
 static buffer read_stdin(heap h)
 {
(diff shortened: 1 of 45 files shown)


mkfs/mkfs.c
```diff
@@ -12,6 +12,9 @@
 #include <limits.h>
 
 #include <region.h>
+#include <storage.h>
+
+#define BOOTFS_SIZE (8 * MB)
```
wjhun, line 17: Minor: parameters like these are typically stored in src/runtime/uniboot.h.

francescolavra: OK, I moved it there.

Issue #1165, opened 2020-05-05
tfs: inflate existing file extent when appending, merge adjacent extents where possible, eliminate extent size maximum


wjhun:
Right now we're allocating tfs extents more often than we need to. Explore the possibility of extending an existing file extent on append rather than allocating a new one. To accommodate the extension, a larger default storage allocation size could be used for each new extent, or id_heap_set_area could be used to try and extend an existing allocation (failure would just mean creating a new extent).

We could also look at merging extents that are contiguous on disk, but such conditions should become much less frequent if the above is implemented, so it's not clear how valuable such a function would be.

Once we complete the removal of all temporary buffers for retaining extent data, we can also remove the 1MB maximum extent size.

wjhun:
Extension of extents to fill allocated space was merged with #1207. We can still choose to:

- join extents that are adjacent on storage (keeps the amount of file metadata down)
- remove (or raise) the extent size maximum

  This was required a long time ago when we needed to allocate contiguous memory to hold an entire extent, which is no longer the case. (Note: Given that id heap allocations are aligned (but not padded) to size, this may result in greater use of storage space, which in cases like the boot partition would just be wasted space. There are some applications within the kernel that currently rely on this property of the id heap, and it does allow for a faster bitmap search, but we could explore a revisitation or parameterizing behavior on heap creation.)

francescolavra:
#1765 implements extension of the allocated space of a given extent and removes the 1-MB upper limit on extent size.

Issue #1127, opened 2020-04-15
support writing of unaligned byte extents to pagecache


wjhun:
As an optimization, writing byte-based extents directly to pagecache buffers can eliminate the head/tail intermediate reads in tfs code. The tfs "dma buf" code can then be eliminated there as well.

see #1125

wjhun:
Resolved in #1207

PR #1207, opened 2020-06-29
add per-file nodes to pagecache and related tfs work

### LINKED ISSUE #1126: complete conversion to scatter gather I/O at tfs and syscall levels
- add sg write interfaces to tfs and pagecache
- add sg read/write methods to struct fdesc for use with iovec (#1115) and sendfile (#1116)

wjhun:
This work prepares for file-backed demand paging by caching file data in pages indexed by file offset. Per-file read and write methods are now closures directly into the cache, which in turn fills page data from and/or writes updated pages into tfs. The pagecache no longer directly calls block device I/O, instead leaving it to tfs to both manage filesystem extent and meta updates as well as initiate I/O requests. tfs duties are now confined to purely managing metadata and handling these operations; the dma heap has been removed, and there is no direct managing of file data other than to apply I/O requests from the cache. Additionally, tfs metadata such as extent ranges and allocated sizes are now kept in block units (aside from filelength). Updates to metadata are posted immediately to the tlog tuple staging buffer, but no longer kick off an instantaneous log flush. Rather, a delay timer is maintained for log flushing in order to amortize flush overhead across more updates. Merges for metadata update completions (and in turn completions for operations like filesystem_mkdir, rename, delete, etc.) have also been obviated.

Note: Revamping the locking scheme in the pagecache to be safe for page eviction didn't make the cut, so sadly that will be piled on to the other SMP bitrot.

Resolves #1126, #1127, #1165


wjhun:
> I tested this with a couple of database applications, and there is an unhandled page fault.

Thank you. This helped me find two serious bugs: The tlog completion vector wasn't being cleared after service (causing completions to be applied repeatedly - the cause of the crash), and write_sg_finish in the pagecache was applying the write completion a second time if the write into tfs gave an error.

To address the second issue, and the general lack of handling fs errors, I decided to store the handled error as a pending error in the pagecache volume. Because completions for writes to the pagecache are applied immediately after writing into the cache buffers, and not after I/O completes, there isn't a way to precisely report fs or storage errors. Storing the pending error at least allows deferred error reporting on a subsequent write so that a program (such as the test you provided) won't run indefinitely. Also, I decided to include more context for tfs errors by encoding an fs_status into error values. These errors eventually propagate back to the syscall, which in turn can unpack them and extract an errno using sysreturn_from_fs_status_value. The test program will now exit after filling storage, with write returning -ENOSPC.

One issue with this change is that the cache does not yet clear pending errors. I think the "right" way to do this would be to test for and clear specific conditions on related events, e.g. clearing a pending error with FS_STATUS_NOSPACE after an extent has been deleted. Until then, such an error is unrecoverable. This could be an issue if a program depends on (recoverable) errors to e.g. probe available size. I don't know if that's something that really happens in the wild - if it does with anything we're running now, this will have to be revised before merging. (I thought about associating the error with a node/file and clearing after reporting once - but that doesn't work with the test case you provided as multiple files are opened and written to. Then I tried doing so with the volume itself, but the error was just getting reported to tlog on the flush ... it just didn't seem like a reliable way to report a pending error condition.)

Clearly there's more work to be done here - namely coming up with a robust policy for error reporting and clearing pending conditions - but at least there is some degree of reporting fs and storage errors with context now.

Review / trial of the last commit would be appreciated.

src/tfs/tfs.c
```diff
@@ -576,335 +460,245 @@ static void remove_extent_from_file(fsfile f, extent ex, merge m)
...
-                        msg_err("failed: can't find extent tuple\n");
-                        goto fail;
-                    }
-                    if (table_find(extent_tuple, sym(uninited))) {
-                        tfs_debug("   removing uninited flag\n");
-                        table_set(extent_tuple, sym(uninited), 0);
-                        filesystem_write_eav(f->fs, extent_tuple, sym(uninited), 0,
-                                apply_merge(m_meta));
-                    }
-                    e->uninited = false;
```
francescolavra, line 870: This block of code, whose purpose is to remove the "uninited" flag from an extent being filled with user data, seems to be missing in the new code. This causes file data to always read as zeros for files that have been `fallocate()`d, even after the files have been written to.

wjhun:
> The latest commits fix the issues I was seeing with database applications, but now I noticed another issue regarding the "uninited" flag.

I just pushed a commit which restores this function (see write_extent). As a slight optimization, each extent meta is stashed in the extent struct to avoid multiple table lookups with each extent update.

A simple persistence test has been added to the write runtime test which also tests fallocate and reversal of an uninited extent. The test must be run manually (not part of runtime-tests) by using the '-p' flag: once to create the file, do the fallocate and write, and a second time (with the same image, of course) to verify the change. I was able to verify that the test fails prior to the uninited fix and passes with the fix.

francescolavra:
> > The latest commits fix the issues I was seeing with database applications, but now I noticed another issue regarding the "uninited" flag.
> 
> I just pushed a commit which restores this function (see write_extent).

Looks good to me, and solves the last remaining issue I observed when testing different applications.

PR #1212, opened 2020-07-09
support for MAP_SHARED and MAP_PRIVATE mappings, msync(2)

### LINKED ISSUE #1113: file backed demand paging

wjhun:
This is a rider to the file-backed demand paging PR (#1211 - please compare against the file-pf branch when reviewing, at least until that is merged and I can rebase) to implement propagation of writes to a MAP_SHARED mapping to underlying storage and implement copy-on-write for MAP_PRIVATE pages. msync(2) is also implemented, and a per-node sync has been added to the pagecache (with an update for fsync(2)).

On an mmap with MAP_SHARED, the pagecache is made aware of the new shared mapping, whose page table entries are periodically scanned for dirty bits. For each dirty page that is found in the page table, the associated pagecache_page is placed on the cache's "dirty" list and the pte's dirty bit is cleared. When scanning is complete, dirty pages are committed to storage. This process is triggered by a periodic timer (PAGECACHE_SCAN_PERIOD_SECONDS in uniboot.h), whenever shared mappings are present, or immediately on a map close, msync, fsync or other cache-wide sync.

MAP_PRIVATE for file-backed mappings is implemented by marking pages as read-only. On a page fault for a write, handle_protection_fault checks if a write protection violation occurred on behalf of a vmap that is set as writable. Using this as an indicator of COW, it directs the pagecache to allocate a new (physical) page, copy the cached file content into it, and replace the cached page with the copy, set as writable.

Closes #1113


wjhun:
Ok, do_demand_page will now deliver SIGBUS to the offending thread if a file-backed access extends beyond the file length. Also, truncate_internal now calls "truncate_file_maps" to commit and unmap any mapped pages that extend beyond a newly-truncated length. I added a test to induce and catch SIGBUS to the mmap test and also fixed a few issues that were preventing the test from being run on linux directly. (Incidentally, we should aim to parity-check all of our runtime tests on linux.)

src/unix/mmap.c
```diff
@@ -681,6 +681,56 @@ static boolean mmap_reserve_range(process p, range q)
...
+
+    if (flags & MS_ASYNC)
+        return 0;               /* nop / default behavior */
+
+    /* TODO: Linux appears to only use this flag to test whether a map
+       is locked (and return -EBUSY if it is). We don't swap out
+       pages, so mlock is a stub for us. It's not clear if we will
+       later need to track locked status and test for this condition. */
+    if (flags & MS_INVALIDATE)
+        return 0;
```
francescolavra, line 747: When Linux returns 0, it means the sync has been done, while if the sync is not being done (e.g. because of memory locks) an error is returned. Here we are returning 0 without actually doing the sync, which seems inconsistent, we should either ignore the MS_INVALIDATE flag, or return error (EBUSY) when the flag is specified. My preference would be to return EBUSY, since as you noted all of our mappings currently are resident in memory.

test/runtime/mmap.c
```diff
@@ -669,6 +676,279 @@ void mremap_test(void)
...
+    p = mmap(NULL, PAGESIZE, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (p == (void *)-1ull)
+        handle_err("mmap mapfile, first page");
+
+    /* induce kernel page fault by writing from mmaped area */
+    rv = write(out, p, PAGESIZE);
+    if (rv < 0)
+        handle_err("write");
+    if (rv < PAGESIZE)
+        printf("   short read: %d\n", rv);
```
francescolavra, line 734: "short write"

src/unix/mmap.c
wjhun: > When Linux returns 0, it means the sync has been done, while if the sync is not being done (e.g. because of memory locks) an error is returned.

This is true if MS_SYNC is specified (not MS_ASYNC). The error here was in thinking that MS_INVALIDATE was mutually exclusive with the other flags. But the msync man page says for EBUSY: "MS_INVALIDATE was specified in flags, and a memory lock exists for the specified address range." File locking isn't applicable for us ... so the question is whether to default to returning 0 or treat all files as locked and return -EBUSY as you suggest. I'm going to assume the latter is the safest option for now. (Updated, and also added a default to ASYNC if neither sync flag is specified to match Linux behavior.)

There seems to be quite a bit of confusion as to the purpose of this flag. I wasn't able to find a use case for it from a cursory search. And Linux does not invalidate cache pages when this flag is specified; it only returns -EBUSY for locked pages.

https://groups.google.com/g/comp.unix.programmer/c/reGmzSHKBHE?pli=1
https://stackoverflow.com/questions/38689221/historical-current-use-of-ms-invalidate

PR #1224, opened 2020-07-16
x86 HPET: add support for standard (I/O APIC) interrupt mapping

francescolavra:
The current HPET driver configures the HPET interrupt mapping using the FSB option, without checking whether this option is actually supported. This prevents the HPET from generating timer interrupts in platforms (such as VMWare ESXi) where FSB mapping is not supported.
This PR adds a check for the FSB capability in the HPET timer configuration register, and sets up interrupt mapping based on this
capability: if FSB mapping is not supported, standard mapping (which routes interrupts to the I/O APIC) is set up instead.

Note: QEMU by default reports its HPET peripheral as not supporting FSB mapping, even though this feature is actually supported; for this reason, after this commit, when running on QEMU without KVM the kernel sets up the HPET using standard mapping instead of FSB mapping (when KVM is enabled this doesn't apply, since with KVM the kernel uses the TSC deadline timer instead of the HPET). In order to make QEMU report the FSB capability, the '-global hpet.msi=true' command line option needs to be passed to QEMU.

### DIFF
--- a/src/x86_64/hpet.c
+++ b/src/x86_64/hpet.c
@@ -2,6 +2,8 @@
 #include <pci.h>
 #include <page.h>
 
+#include "apic.h"
+
 #define HPET_TABLE_ADDRESS 0xfed00000ull
 #define HPET_MAXIMUM_INCREMENT_PERIOD 0x05F5E100ul /* 100ns */
 
@@ -93,7 +95,10 @@ struct HPETMemoryMap {
 
 static volatile struct HPETMemoryMap* hpet;
 static timestamp hpet_period_scaled_32;
-static int hpet_interrupts[4];
+static struct hpet_timer_cfg {
+    int interrupt;
+    u64 config;
+} hpet_timers[4];
 
 static u64 hpet_config_get() __attribute__((noinline));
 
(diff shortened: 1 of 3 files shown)


src/x86_64/hpet.c
```diff
@@ -117,22 +119,45 @@ static u64 hpet_main_counter()
...
         hpet_interrupts[timer] = allocate_interrupt();
-        msi_format(&a, &d, hpet_interrupts[timer]);
-        hpet->timers[timer].fsb_int = ((u64)a << 32) | d;
+        if (hpet->timers[timer].config & TCONF(FSB_INT_DEL_CAP)) {
+            u32 a, d;
+            msi_format(&a, &d, hpet_interrupts[timer]);
+            hpet->timers[timer].fsb_int = ((u64)a << 32) | d;
+            c |= TCONF(FSB_EN_CNF);
+        } else {
+            u32 route_cap = (hpet->timers[timer].config >>
```
wjhun, line 130: Nit: the macro "field_from_u64" can be used here

src/x86_64/hpet.c
```diff
@@ -117,22 +119,45 @@ static u64 hpet_main_counter()
...
+                    HPET_TIMER_CONFIG_INT_ROUTE_CAP_SHIFT) &
+                    MASK(HPET_TIMER_CONFIG_INT_ROUTE_CAP_BITS);
+            int gsi = -1;
+            for (int i = 0; i < 64; i++) {
```
wjhun, line 134: Should this be i < 32? route_cap is a u32.

src/x86_64/hpet.c
```diff
@@ -117,22 +119,45 @@ static u64 hpet_main_counter()
...
     }
 
-    u64 c = TCONF(FSB_EN_CNF) | TCONF(32MODE_CNF) | TCONF(INT_ENB_CNF);
     if (periodic) {
         if ((hpet->timers[timer].config & TCONF(PER_INT_CAP)) == 0) {
 	    console("HPET timer not capable of periodic interrupts.\n");
 	    return;
 	}
-	c |= TCONF(VAL_SET_CNF) | TCONF(TYPE_CNF);
+        hpet->timers[timer].config |= TCONF(VAL_SET_CNF) | TCONF(TYPE_CNF);
```
wjhun, line 157: Couldn't the two writes to this register be merged into one, as it was before?

francescolavra: Before, the bits of the register that control interrupt mapping were set up statically, so we could always re-write them with fixed values, but now their contents may depend on a value (the GSI number) calculated at initialization, so any time we select between periodic and non-periodic mode we have to read first the current contents of the register. Do you prefer something like below?
```
u64 c;
if (!hpet_interrupts[timer]) {
  c = ...;
} else {
  c = hpet->timers[timer].config;
}
if (periodic) {
  ...
  c |= TCONF(VAL_SET_CNF) | TCONF(TYPE_CNF);
} else {
  c &= ~(TCONF(VAL_SET_CNF) | TCONF(TYPE_CNF));
}
hpet->timers[timer].config = c;
```

wjhun: Yes, or maybe stash a copy of the GSI? I'm assuming accesses to HPET registers cause a vm exit and are expensive, but I don't know how ESXi handles them.

src/x86_64/hpet.c
francescolavra: Yes, I changed it to i < 32

src/x86_64/hpet.c
francescolavra: OK, I created an array of struct hpet_timer_cfg that contains both the interrupt number (which replaces the hpet_interrupts static array) and the timer configuration, so now timer_config() only does one write (and no reads) to the HPET timer config register (except for the initial interrupt configuration).

PR #1229, opened 2020-07-25
iov_op(): add support for scatter gather I/O operations

### LINKED ISSUE #1115: future work page cache / readv/writev hookup to page-cache

francescolavra:
If the file descriptor on which a vector I/O operation is being invoked implements scatter/gather I/O, the kernel now uses this implementation, instead of invoking the linear read/write closures for each element in the I/O vector.
This PR also adds a scatter/gather write implementation for regular files.

Closes #1115.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -312,7 +312,18 @@ boolean validate_iovec(struct iovec *iov, u64 len, boolean write)
     return true;
 }
 
+declare_closure_struct(2, 2, void, iov_op_each_complete,
+                       int, iovcnt, struct iov_progress *, progress,
+                       thread, t, sysreturn, rv);
+
+declare_closure_struct(2, 0, void, iov_bh,
+                       struct iov_progress *, p, thread, t);
+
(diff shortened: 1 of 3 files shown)


francescolavra:
> All looks good. The non-sg / each_complete case remains a serial issue (vs parallel with merge) - which might account for some latency depending on whether or not the operation is a direct return. But maybe it's better to leave it and instead work on adding sg r/w methods to all descriptor types.

In general we cannot start N parallel I/O ops, because there are cases where only a subset of those operations can complete immediately, while the remaining ones would block (possibly indefinitely). E.g. the case where a single packet is available to be read from a socket but the I/O vector has room for more than one packet: the correct thing to do is to read the existing packet and then return immediately, without waiting for other packets. This can be done reliably either by doing the transfers serially (and checking for EPOLLIN/EPOLLOUT after each transfer, as iov_op_each_complete does), or by passing the entire I/O vector (or rather an sg list) to the low-level I/O implementation and let it manage the transfer(s).

PR #1236, opened 2020-08-04
fix mac ci build

### LINKED ISSUE #1227: un-supported version of nasm comment killing nightly mac build
https://github.com/nanovms/nanos/blob/master/src/x86_64/crt0.s#L127 - although if you git blame this line is really damn old

https://stackoverflow.com/questions/48148597/can-a-hashtag-be-used-as-a-comment-in-assembly-nasm

this might be correlated w/the fact that the nightly build is taking 27min? to install wget && nasm on mac?

wjhun:
fix various bugs uncovered by mac nightly build:

- hyperv atomics used asm flags constraints (`@CC<cond>`)without checking for support; use alternate versions when `__GCC_ASM_FLAG_OUTPUTS__` is not defined
- missing include path for runtime tests
- use absolute path for NANOS_TARGET_ROOT
- wrong type in writev test caused always-true condition
- incorrect use of buf->length in tuple_test
- unsupported comment syntax in crt0.s

resolves #1227

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
       - run: 
           name: macbuild
           environment:
-            NANOS_TARGET_ROOT: target-root
+            NANOS_TARGET_ROOT: $HOME/project/target-root
           command: |
             OPS_DIR=$HOME/.ops
             PATH=$HOME/.ops/bin:$PATH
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - run: 
           name: macbuild
           environment:
-            NANOS_TARGET_ROOT: target-root
+            NANOS_TARGET_ROOT: /Users/distiller/project/target-root
```
mkhon, line 65: Have you checked if ~ can be used here?

.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
...
```
mkhon, line 84: Ditto

.circleci/config.yml
wjhun: I changed them to use $HOME.

PR #1241, opened 2020-08-10
reorganization of build to support multiple architectures and platforms

wjhun:
These changes allow specification of processor architecture, target platform and cross toolchain to the build. The default intel build is now under the 'pc' platform, and stub definitions and a hello world are included for the aarch64 (amd64) platform 'virt', which corresponds to the virt generic arm machine type in qemu.

For example, to cross-compile the arm stub:

```make CROSS_COMPILE=aarch64-none-elf- ARCH=aarch64 PLATFORM=virt TARGET=dummy run-noaccel```

A '-k' option has been added to mkfs to specify which kernel binary to use, eliminating the need to specify it in the manifest.

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -39,40 +30,12 @@ AWS_AMI_IMAGE=	nanos-$(TARGET)
...
-	$(Q) $(RM) -r release
-	$(Q) $(MKDIR) release
-	$(CP) $(MKFS) release
-	$(CP) $(BOOTIMG) release
-	$(CP) $(STAGE3) release
-	cd release && $(TAR) -czvf nanos-release-$(REL_OS)-${version}.tar.gz *
-
-contgen:
-	$(Q) $(MAKE) -C $@
+.PHONY: image release target distclean
```
francescolavra, line 33: The `release` and `target` Makefile targets don't do anything now in the top-level Makefile, so either they should be removed from the above line, or there should be a redirection to the appropriate platform directory, otherwise when running e.g. `make release` one gets the misleading message `make: Nothing to be done for 'release'`.

rules.mk
```diff
@@ -10,19 +10,24 @@ TOOLDIR=	$(OUTDIR)/tools/bin
...
 CAT=		cat
-CC=		$(CROSS_COMPILE)gcc
+# XXX llvm for darwin
 CP=		cp
 DD=		dd
 ifeq ($(UNAME_s),Darwin)
 GNUTAR=		gnutar
 else
 GNUTAR=		tar
+NATIVE_CC=	gcc
```
francescolavra, line 30: It appears this is not used anywhere?

src/aarch64/machine.h
```diff
@@ -0,0 +1,177 @@
...
+static inline void memory_barrier(void)
+{
+}
+
+#if 0
+static inline word fetch_and_add(word *variable, word value)
+{
+    asm volatile("lock; xadd %0, %1" : "+r" (value), "+m" (*variable) :: "memory", "cc");
+    return value;
+}
```
francescolavra, line 151: I guess this should be removed, since there is a real implementation a few lines below.

src/unix_process/unix_process_runtime.c
```diff
@@ -72,13 +85,69 @@ void halt(char *format, ...)
...
+struct tagheap {
+    struct heap h;
+    heap mh;
+    u64 vtag;
+};
+
+static void tag_dealloc(heap h, u64 a, bytes s)
+{
+    struct tagheap *th = (struct tagheap *)h;
+    tag_debug("%s: tag %d, a 0x%lx, s 0x%lx\n", __func__, th->vtag >> VA_TAG_OFFSET, a, s);
```
francescolavra, line 106: tag_debug() already adds `__func__` at the beginning of the output, so another `__func__` here is redundant. The same applies to the other uses of tag_debug() in this file.

wjhun:
> Regarding moving the stage3.c file from stage3/ to platform/pc/, it seems to me that this file contains platform-independent code only, so I think it should go to src/kernel/ instead.

Yes, this will move to src/kernel along with the init parts of service.c There's a bunch of platform-specific stuff in service.c that needs to move to the platform/pc.

There's definitely more work to do to split machine-dependent stuff out of the common areas. This PR is just a checkpoint.

wjhun:
I addressed the above issues and swapped src/kernel/service.c and platform/pc/stage3.c, as service is largely platform-specific and, as noted, stage3 is all platform and architecture independent. The machine independent initialization in service.c will be split out in an upcoming PR.

PR #1242, opened 2020-08-11
TFS: implement log compaction

### LINKED ISSUE #21: tfs should compress log
if the number of removals gets to be a sufficiently large ratio of the log space, the 
kernel should start a background process to write the in-memory version
(where the creates and removal have already been reconciled) to new segments.

the root of the log should be (atomically) replaced with a single sector containg
a next record pointing to the new log. this clearly depends on https://github.com/deferpanic/uniboot/issues/20

if the disk is exhausted, then it may be impossible to regain this space because of the
lack of working area.

francescolavra:
Any filesystem operation done by a running application causes one or more entries to be appended to the TFS log, and these new entries may obsolete existing log entries. As a result, after many filesystem operations, the TFS log may contain many obsolete entries, which wastes storage space and makes filesystem loading at boot inefficient.
This PR implements log compaction, which eliminates obsolete entries and recreates a "clean" log. Log compaction is triggered when the number of obsolete and total entries in the current log exceed certain thresholds.
In order to account for any filesystem operation done while log compaction is in progress, log entries created during this time are written to both the current and the new log, so that the filesystem remains in a consistent state regardless of whether the log compaction completes successfully.
After a successful log compaction, any tuples that are no longer associated to existing filesystem entries are destroyed.
In order to prevent log compaction from adding to the log non-persistent filesystem entries, any tuple corresponding to a non-persistent entry is marked with a "no_encode" symbol, which is checked against in the encode_tuple() function.
In order to avoid use-after-free issues if an existing log is being flushed during log compaction, a reference count with an associated deallocation closure is being added to the TFS log structure: flush_log_extension() increments the refcount, and both flush_log_extension_complete() and log_switch_complete() decrement the refcount.

Closes #21.


src/tfs/tlog.c
```diff
@@ -29,6 +30,13 @@ static const char *tfs_magic = "NVMTFS";
 #define TFS_EXTENSION_LINK_BYTES (1 + 2 * MAX_VARINT_SIZE)
 #define TFS_LOG_RESERVED_BYTES (TFS_EXTENSION_HEADER_BYTES + TFS_EXTENSION_LINK_BYTES)
 
+/* Minimum number of obsolete log entries needed to trigger a log compaction. */
+#define TFS_LOG_COMPACT_OBSOLETE   8192
```
wjhun, line 34: move these tunables to uniboot.h?

src/tfs/tlog.c
```diff
@@ -349,8 +541,12 @@ closure_function(1, 1, void, log_flush_timer_expired,
 
 static void log_set_dirty(log tl)
 {
-    if (tl->dirty)
+    if (tl->dirty) {
+        if (buffer_length(tl->tuple_staging) >= bytes_from_sectors(tl->fs,
+                range_span(tl->current->sectors)) / 2)
+            log_flush(tl, 0);
```
wjhun, line 547: can cancel timer here

src/tfs/tlog.c
```diff
@@ -365,60 +561,73 @@ static void log_set_dirty(log tl)
...
 #endif
 
-void log_write_eav(log tl, tuple e, symbol a, value v)
+boolean log_write_eav(log tl, tuple e, symbol a, value v)
 {
     tlog_debug("log_write_eav: tl %p, e %p, a %b, v %p\n", tl, e, symbol_string(a), v);
     u64 len = buffer_length(tl->tuple_staging);
-    encode_eav(tl->tuple_staging, tl->dictionary, e, a, v);
+    if (len >= bytes_from_sectors(tl->fs, range_span(tl->current->sectors)))
+        return false;
```
wjhun, line 569: It's not clear to me why this would fail (same for log_write). tl->tuple_staging can store up an arbitrary amount of encoded tuple data which can span multiple extensions; it should not be limited to the size of the current extension. A very large tuple can even span an extension boundary (this is why we have TUPLE_EXTENDED).

The only problem with this scheme is that an error on log extension creation (out of storage) would happen asynchronously, as a result of a log flush. But, much like reporting write errors in tfs, perhaps it would suffice to record such an error in the log and report it on a subsequent log_write_* (imprecise error reporting).

francescolavra: Both log_write() and log_write_eav() call log_set_dirty(), which has now been modified in order to call log_flush() directly if the staging buffer is "full enough". Because of that, the staging buffer (which is consumed when log_flush() is called) should never grow more than the "full enough" threshold used in log_set_dirty(); and log_write() and log_write_eav() check against a threshold double that in log_set_dirty(). So an overly full buffer indicates that a previous call to log_flush() was not successful.
But if you prefer I can replace those checks in log_write() and log_write_eav() with the error reporting scheme you mentioned (we would need to add an error flag in the tlog structure, which will be set if a log_flush cannot complete successfully, and log_write() and log_write_eav() would return false if this error flag is set).

src/tfs/tlog.c
francescolavra: The timer is already being canceled in log_flush()

src/tfs/tfs.c
```diff
@@ -429,64 +448,32 @@ static void add_extent_to_file(fsfile f, extent ex)
     table_set(e, sym(length), value_from_u64(h, range_span(ex->node.r)));
     table_set(e, sym(allocated), value_from_u64(h, ex->allocated));
     if (ex->uninited)
-        table_set(e, sym(uninited), null_value());
+        table_set(e, sym(uninited), null_value);
+    symbol offs = intern_u64(ex->node.r.start);
+    fs_status s = filesystem_write_eav(f->fs, extents, offs, e);
+    if (s != FS_STATUS_OK) {
+        destruct_tuple(e, true);
```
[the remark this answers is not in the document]
francescolavra: At this point we cannot destroy the extents tuple (even if empty) because it has been encoded in the log already. We might remove a reference to it from the file metadata (as in `table_set(f->md, sym(extents), 0);`), but I think it's better to keep a reference to it, otherwise if later a new attempt is made to add an extent to the file, the extents tuple would be wastefully re-allocated.

src/runtime/tuple.c
```diff
@@ -143,8 +143,18 @@ value decode_value(heap h, tuple dictionary, buffer source)
...
                 if (!s) halt("indirect symbol not found: 0x%lx, offset %d\n", nlen, source->start);
             }
-            value nv = decode_value(h, dictionary, source);
+            value nv = decode_value(h, dictionary, source, total, obsolete);
+            if (obsolete) {
+                value old_v = table_find(t, s);
+                if (old_v) {
+                    (*obsolete)++;
+                    if (!nv)
+                        (*obsolete)++;
```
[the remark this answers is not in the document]
francescolavra: If the value of a tuple is being deleted at a given point, it means there are (at least) 2 unnecessary log entries: the one that sets the initial value, and the one that removes the value. So if both entries were removed, the log would still be consistent with the current state. So "obsolete" is meant to count the number of log entries that would not be there if the log were to be recreated "cleanly". I know "obsolete" is not exactly the right word here, because in the above scenario the latter log entry is needed to undo what the former entry did, but I couldn't come up with a better naming.

src/tfs/tlog.c
francescolavra: OK, I moved the TFS_LOG_COMPACT_OBSOLETE and TFS_LOG_COMPACT_RATIO tunables to src/config.h.

francescolavra:
Also, in log_complete() in tfs.c, I added a check `if (is_ok(s))` before calling `fixup_directory(fs->root, fs->root)`, because now that the root tuple is allocated internally in the TFS code, if a filesytem log cannot be read correctly the root tuple is not allocated and cannot be touched.

PR #1247, opened 2020-08-20
Storage: add support for multiple filesystem mounts

francescolavra:
Multiple filesystems can now be mounted by specifying a "mounts" entry in the root tuple of the manifest file. This entry is a tuple containing one entry for each mount, with format <uuid>:<mountpoint>, where <uuid> is a volume identifier (which corresponds to the UUID of the filesystem contained in the volume) and <mountpoint> is the path of the directory in which the filesystem is to be mounted (this directory must exist in the root filesystem). Example to mount 2 filesystems (one at /mnt and the other at /mnt2):
mounts:(
  2b8750f4-07f2-2584-7dcc-892b3b52bd07:/mnt
  f6f1d240-6b80-2f01-3a87-186b64b7caa6:/mnt2
)
The UUID is assigned to a given filesystem when the filesystem is created by the mkfs tool, and its value is displayed in the terminal output by this tool (as well as by the dump tool).
In order to create a (non-bootable) filesystem to be used for a volume, the mkfs tool must be invoked without passing a boot image or a kernel image. To create an empty filesystem:
mkfs -e my_volume.raw
To create a non-empty filesystem:
mkfs my_volume.raw < my_volume.manifest
The manifest file in the above command must contain all the files and folders that need to go in the filesystem.

The kernel probes any detected storage devices at boot time: if a device contains an MBR with a partition table, then it is assumed to be the root volume, otherwise if a filesystem is detected in the first disk sector and its UUID corresponds to one of the mount entries, it is mounted; if no MBR and no filesystem is found, then the storage device is ignored.
If one or more filesystems are detected and need to be mounted, the kernel waits for the mount operations to complete before starting the user process.
At runtime, the filesystem corresponding to each open file is referenced in the new "fs" field of struct file.
The virtIO SCSI driver has been modified to be able to handle multiple SCSI targets (which are used on a Google Cloud instance
when multiple disks are attached).

Usage notes:

- Qemu: to specify additional volumes, for each volume add the following command line options:
`-drive if=none,id=hd<N>,format=raw,file=<volumeN.raw> -device scsi-hd,bus=scsi0.0,drive=hd<N>`
where `<N>` = 1,2,..., and `<volumeN.raw>` is the name of the file created via mkfs for the N-th volume

- Google Cloud Platform: change the volume.raw file name to disk.raw, compress it into a tar.gz file, then upload the tar.gz file to Google Cloud Storage; then, create a Compute Engine image by specifying "Cloud storage file" as image source, and selecting the tar.gz file; then, create a Compute Engine disk by specifying "Image" as source type, and selecting the image created above; then, to attach a disk to a VM instance, select the instance, select "Edit", go to the "Additional disks" section, select "Attach existing disk" and then select the disk created above.

- Firecracker: for each additional volume, add a JSON object to the "drives" array in the Firecracker configuration file, as shown in the example below:
```
    {
      "drive_id": "mntN",
      "path_on_host": "volumeN.raw",
      "is_root_device": false,
      "is_read_only": false
    }
```

- AWS: not supported yet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ MKFS=		$(TOOLDIR)/mkfs
 BOOTIMG=	$(PLATFORMOBJDIR)/boot/boot.img
 KERNEL=		$(PLATFORMOBJDIR)/bin/kernel.img
 
-all: image
+all: image tools
 
-.PHONY: image release target distclean
+.PHONY: image release target tools distclean
 
 include rules.mk
 
@@ -53,6 +53,9 @@ release: mkfs
 
 target: contgen
 	$(Q) $(MAKE) -C test/runtime $(TARGET)
+
+tools:
+	$(Q) $(MAKE) -C $@
 
 distclean: clean
 	$(Q) $(RM) -rf $(VENDORDIR)
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -23,6 +23,7 @@ SRCS-kernel.img= \
 	$(SRCDIR)/kernel/pvclock.c \
 	$(SRCDIR)/kernel/schedule.c \
 	$(SRCDIR)/kernel/stage3.c \
+	$(SRCDIR)/kernel/storage.c \
 	$(SRCDIR)/kernel/symtab.c \
 	$(SRCDIR)/kernel/vdso-now.c \
 	$(SRCDIR)/net/direct.c \
--- a/platform/pc/boot/stage2.c
+++ b/platform/pc/boot/stage2.c
@@ -300,14 +300,12 @@ void newstack()
 
     setup_page_tables();
 
-    pagecache pc = allocate_pagecache(h, h, 0, PAGESIZE);
-    assert(pc != INVALID_ADDRESS);
+    init_pagecache(h, h, 0, PAGESIZE);
     create_filesystem(h,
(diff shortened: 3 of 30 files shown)


ronaudinho:
not sure if this is the best place to ask, but having difficulties with this implementation on https://github.com/nanovms/ops/issues/587. to validate what i'm trying to implement, i created a Go executable that:
1. reads the root dir of the unikernel
2. calls `hello` (taken from https://github.com/nanovms/ops-examples/tree/master/golang/01-hello-world), which would be mounted at `/mnt`

my `hello.manifest` looks like this
```
(
  children: (
    hello: (contents:(/path/to/hello))
  )
  environment: (USER: root PWD:/)
)
```
which i then run with `mkfs hello.raw < hello.manifest` to get the raw file and the UUID, and i add those to the main manifest
```
(
  children: (
    ... // the rest of the directories
    mnt: (children: ())
  )
  mount: (
    "UUID-for-hello.raw:/mnt"
  )
  environment: (USER: root PWD:/)
)
```
but when i do this, my Go executable cannot stat the root dir of the unikernel (without `mount:(...)`, it runs fine)

francescolavra:
The problem is that the `mount` entry in your main manifest is named incorrectly, it should be `mounts` instead.

PR #1251, opened 2020-08-24
Implement protection against arbitrary code execution

francescolavra:
This PR implements an "exec protection" feature that, if enabled, prevents the kernel from executing any code outside the main program executable and other "trusted" files explicitly marked as executable. This feature is enabled by adding an "exec_protection" symbol to the root tuple in the manifest file; then, any files containing executable code that need to be executed by the application (usually, shared libraries) need to marked in the manifest file with an "exec" symbol. It is not necessary to add the "exec" symbol to the main program, since this is always considered as a "trusted" executable file.
Example of a manifest file with exec protection enabled and with one shared library:
(
  children:(
    my_app:(contents:(host:bin/my_app))
    lib64:(children:(ld-linux-x86-64.so.2:(contents:(host:/lib64/ld-linux-x86-64.so.2))))
    lib:(children:(x86_64-linux-gnu:(children:(libc.so.6:(exec:t contents:(host:/lib/x86_64-linux-gnu/libc.so.6))))))
  )
  program:/my_app
  environment:(USER:bobby PWD:/)
  exec_protection:t
)

(Note the exec symbol in the libc.so.6 shared library, without which the application would fail to start, with an error such as "error while loading shared libraries: libc.so.6: failed to map segment from shared object".)

The application cannot modify the executable files and cannot create new executable files.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -196,6 +196,7 @@ vmap allocate_vmap(rangemap rm, range q, struct vmap k)
         return vm;
     rmnode_init(&vm->node, q);
     vm->flags = k.flags;
+    vm->allowed_flags = k.allowed_flags;
     vm->node_offset = k.node_offset;
     vm->cache_node = k.cache_node;
     if (!rangemap_insert(rm, &vm->node)) {
@@ -399,6 +400,17 @@ static sysreturn mincore(void *addr, u64 length, u8 *vec)
     return 0;
 }
 
+closure_function(2, 1, void, vmap_update_protections_validate,
+                 u32, newflags, boolean *, prot_violation,
+                 rmnode, node)
+{
+    if (*bound(prot_violation))
+        return;
+    vmap match = (vmap)node;
+    if (bound(newflags) & ~match->allowed_flags)
+        *bound(prot_violation) = true;
+}
+
 /*
    case 1: !head && !tail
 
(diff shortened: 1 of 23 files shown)


src/unix/unix_internal.h
```diff
@@ -423,6 +441,44 @@ static inline tuple file_get_meta(file f)
     return f->f.type == FDESC_TYPE_REGULAR ? fsfile_get_meta(f->fsf) : f->meta;
 }
 
+static inline boolean file_is_readable(fdesc f)
```
wjhun, line 444: fdesc_is_{readable,writable}

src/unix/mmap.c
```diff
@@ -492,15 +507,25 @@ static boolean vmap_update_protections(heap h, rangemap pvmap, range q, u32 newf
...
-        return false;
+        return -ENOMEM;
+    }
+
+    rmnode_handler v_handler = stack_closure(vmap_update_protections_validate,
+        newflags, false);
+    if (!rangemap_range_lookup(pvmap, q, v_handler)) {
+        return -ENOMEM;
+    } else {
+        closure_ref(vmap_update_protections_validate, v);
```
wjhun, line 518: Why not just enclose a pointer to a local boolean rather than do this ref / cast to access prot_violation?

wjhun (approved):
Looks great!

src/unix/unix_internal.h
francescolavra: OK, changed the naming to fdesc_is_{readable,writable}.

src/unix/mmap.c
francescolavra: Indeed I made it more complicated than it needs to be. Now I replaced this with a local boolean and let go of the closure_cast() thing.

PR #1252, opened 2020-08-24
Hyper-V: pv timer

mkhon:
- Reference TSC page support
- VMBUS ET timer

### DIFF
--- a/src/hyperv/vmbus/hyperv.c
+++ b/src/hyperv/vmbus/hyperv.c
@@ -35,13 +35,21 @@
      MSR_HV_GUESTID_OSID_FREEBSD |  \
      MSR_HV_GUESTID_OSTYPE_FREEBSD)
 
+#define CPUID_SSE2  0x04000000
+
 struct hypercall_ctx {
     void            *hc_addr;
     u64             hc_paddr;
 };
 
+struct hyperv_reftsc_ctx {
+    struct hyperv_reftsc    *tsc_ref;
+    struct hyperv_dma   tsc_ref_dma;
+};
+
 typedef struct hyperv_platform_info {
     heap general;                  /* general heap for internal use */
+    heap contiguous;               /* physically */
 
     u32 features;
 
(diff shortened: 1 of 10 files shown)


src/hyperv/vmbus/vmbus_et.c
```diff
@@ -0,0 +1,117 @@
...
+
+typedef struct vmbus_timer *vmbus_timer_t;
+
+static vmbus_timer_t vmbus_et;
+
+static __inline u64
+hyperv_sbintime2count(sbintime_t time)
+{
+    struct timespec ts;
+    timespec_from_time(&ts, time);
```
wjhun, line 62: This is pulling in a unix type just to extract seconds and nanoseconds from a timestamp. You can just use sec_from_timestamp(time) and nsec_from_timestamp(truncate_seconds(time)) instead.

src/hyperv/vmbus/vmbus_et.c
```diff
@@ -0,0 +1,117 @@
...
+
+struct vmbus_timer {
+} vmbus_timer;
+
+typedef struct vmbus_timer *vmbus_timer_t;
+
+static vmbus_timer_t vmbus_et;
+
+static __inline u64
+hyperv_sbintime2count(sbintime_t time)
```
wjhun, line 59: Shouldn't this be timestamp instead of sbintime2?

src/hyperv/vmbus/hyperv.c
```diff
@@ -131,22 +139,80 @@ hypercall_signal_event(bus_addr_t monprm_paddr)
...
+    struct hyperv_reftsc *tsc_ref = hyperv_info.hyperv_ref_tsc.tsc_ref;
+    u32 seq;
+
+    while ((seq = atomic_load_acq32(&tsc_ref->tsc_seq)) != 0) {
+        u64 disc, ret, tsc;
+        u64 scale = tsc_ref->tsc_scale;
+        s64 ofs = tsc_ref->tsc_ofs;
+
+        /* rdtsc_precise contains a load fence */
+        tsc = rdtsc_precise();
```
wjhun, line 154: Is rdtsc_precise absolutely necessary? A while back we had this barrier in our tsc read routine, and in our first few ftrace runs we discovered that it was causing a huge performance hit, apparently due to cpuid causing a full vm exit. I don't know if this is the case for hyperv, I guess it would be trivial to measure (we were just running webg / ab tests).

(Also, so I've read anyway, cpuid doesn't just act as a load fence, but it serializes all instruction execution around it.)

PR #1265, opened 2020-08-29
vm_exit(): optionally dump kernel and Unix heap statistics

### LINKED ISSUE #272: heap tracking
Don't do anything fancy or design anything at this time. Just enumerate the various heaps used within the kernel as well as unix world and, with a compile-time debug flag, just dump out the "allocated" field of each in a user-readable table - for now just on halt() or a fault, or when qemu is killed. In the next phase we can discuss doing proper stats/counter collection using backed tuples, but right now a crude report would suffice.

francescolavra:
If the DUMP_MEM_STATS preprocessor define is present, just before shutting down the VM the kernel dumps on the serial terminal a usage summary for the kernel and Unix heaps.

Closes #272.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -10,6 +10,7 @@
 #include <page.h>
 #include <storage.h>
 #include <symtab.h>
+#include <unix.h>
 #include <virtio/virtio.h>
 #include <vmware/vmxnet3.h>
 #include <drivers/storage.h>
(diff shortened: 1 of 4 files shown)


platform/pc/service.c
```diff
@@ -10,6 +10,7 @@
 #include <page.h>
 #include <storage.h>
 #include <symtab.h>
+#include <unix.h>
```
wjhun, line 13: It's not critical, but it seems like this could be more of a generic kernel thing rather than unix-specific.

Issue #1081, opened 2020-02-28
Use xsave (or xsavec, xsaves) when available


wjhun:
A thread's FP register context is saved and restored using fxsave and fxrstor - a safe common denominator for the platforms we support. We should also detect for newer (and faster) variants, such as xsave, xsavec or xsaves, and use them.

We are currently saving and restoring this state unconditionally. There is some question as to whether we should bother with trying to do a lazy fp save or just keep it within the general context save. From Intel SDM, 3a, section 13.4:

> The use of lazy restore mechanism in context switches is not recommended when XSAVE feature set is used to save/restore states for the following reasons.
> — With XSAVE feature set, Intel processors have optimizations in place to avoid saving the state components that are in their initial configurations or when they have not been modified since they were restored last. These optimizations eliminate the need for lazy restore. See section 13.5.4 in Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 1.
> — Intel processors have power optimizations when state components are in their initial configurations. Use of lazy restore retains the non-initial configuration of the last thread and is not power efficient.
> — Not all extended states support lazy restore mechanisms. As such, when one or more such states are enabled it becomes very inefficient to use lazy restore as it results in two separate state restore, one in context switch for the states that does not support lazy restore and one in the #NM handler for states that support lazy restore.
>

convolvatron:
just noting that the current implementation causes a reproducible fault in the go runtime strongly correlated with the use of futexes. its not clear whether this is a direct causal result of using xsave or just an accidental surfacing of an independent issue

francescolavra:
Support for xsave has been added in #1506. xsavec and xsaves are still not being used.

PR #1277, opened 2020-09-09
Fix booting on Xeon 5520

### LINKED ISSUE #1225: openstack rollup
not sure what all is necessary to support - it uses kvm underneath w/virtio-net && virtio-block (and also virtio-rng)

image

however currently it is just rebooting (uploaded as a raw disk - maybe there are different options to specify when provisioning

this was done using devstack under fusion so it could be the nested virt as well - idk

https://docs.openstack.org/devstack/latest/

mkhon:
Fixes #1225


wjhun (approved):
This is fine for the time being. Related to #1081

PR #1281, opened 2020-09-14
Fixed mincore handling of offsets into fat pages

sanderssj:
During my testing for the eviction changes while I was looping the mmap test, I discovered that the mincore portion of the mmap test would fail infrequently. After some debugging I discovered that this happened when mincore was called on an address offset inside a fat page. The code handling the population of the mincore vector data didn't handle that case properly, so this change fixes that.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -345,13 +345,13 @@ closure_function(3, 3, boolean, mincore_fill_vec,
     u64 pgoff, i;
 
     if (pt_entry_is_present(e)) {
-        if (addr < bound(base))
+        if (addr <= bound(base))
             pgoff = 0;
         else
             pgoff = ((addr - bound(base)) >> PAGELOG);
         if (pt_entry_is_fat(level, e)) {
-            u64 foff = (addr & PAGEMASK_2M) >> PAGELOG;
-            for (i = 0; (i < 512 - (pgoff ? 0 : foff)) && (pgoff + i < bound(nr_pgs)); i++)
+            u64 foff = pgoff ? 0 : (addr & PAGEMASK_2M) >> PAGELOG;
+            for (i = 0; (i < 512 - foff) && (pgoff + i < bound(nr_pgs)); i++)
                 bound(vec)[pgoff + i] = 1;
         } else if (pt_entry_is_pte(level, e)) {
             bound(vec)[pgoff] = 1;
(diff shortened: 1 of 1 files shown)


src/unix/mmap.c
```diff
@@ -345,14 +345,22 @@ closure_function(3, 3, boolean, mincore_fill_vec,
...
+            u64 off, coff;
+            off = (addr & PAGEMASK_2M) >> PAGELOG;
+            if (addr > bound(base)) {
+                pgoff = (addr - bound(base)) >> PAGELOG;
+                coff = 0;
+            } else {
+                pgoff = 0;
+                coff = off;
+            }
+            for (i = 0; (i + coff < 512) && (pgoff + off + i < bound(nr_pgs)); i++) {
```
wjhun, line 359: Just a nit, but I think some expressions here could be simplified if i started at pgoff.

wjhun (approved):
From inspection it looks like the corner cases are covered here, but would it be trivial to add checks to the mincore test? (btw, where did it fail on a 2M mapping - in the stack? program segments? heap? I don't think there would be 2M mappings in any mmapped area as far as I can recall...)

src/unix/mmap.c
sanderssj: It always fails in the stack. I'm not sure about making a test for it, because it's not always that the stack is in a 2M page.

src/unix/mmap.c
```diff
@@ -345,13 +345,14 @@ closure_function(3, 3, boolean, mincore_fill_vec,
     u64 pgoff, i;
 
     if (pt_entry_is_present(e)) {
-        pgoff = (addr - bound(base)) >> PAGELOG;
-
+        if (addr < bound(base))
```
wjhun, line 348: <=

src/unix/mmap.c
```diff
@@ -345,13 +345,14 @@ closure_function(3, 3, boolean, mincore_fill_vec,
...
+            pgoff = 0;
+        else
+            pgoff = ((addr - bound(base)) >> PAGELOG);
         if (pt_entry_is_fat(level, e)) {
-            /* whole level is mapped */
-            for (i = 0; (i < 512) && (pgoff + i < bound(nr_pgs)); i++) {
+            u64 foff = (addr & PAGEMASK_2M) >> PAGELOG;
+            for (i = 0; (i < 512 - (pgoff ? 0 : foff)) && (pgoff + i < bound(nr_pgs)); i++)
```
wjhun, line 354: the pgoff ? conditional could be in foff assignment

wjhun (approved):
A few aesthetic nits but otherwise looks good.

Issue #1253, opened 2020-08-25
fix pagecache locking


wjhun:
The pagecache was originally written with a spinlock protecting accesses to the page tree (held during traversal of pages - during which pages may be added or removed) and a spinlock protecting state transitions of the pages themselves (including removal from previous state list to new state list). When nodes were added, the pages lock changed to a per-node pages tree lock but with the same semantics.

For operations on a node (file), this model seems simple enough: hold the node's pages lock during the operation (usually a traversal, but could also be walking page tables) and only take the state lock when making state transitions or touching a completions vector. (State transitions as a result of asynchronous I/O completions can occur outside of a node lock being held. Needless to say, no node lock acquisition should be attempted while holding the state lock...)

For performing page evictions across the entire cache, however, the pages lock isn't currently being taken - making evictions unsafe on SMP. This lock acquire can be avoided here by instead moving evicted pages to state PAGECACHE_PAGESTATE_FREE - keeping meta for the evicted page resident and relinquishing memory (really this interface a future free physical page / epoch list #1079).

Incidentally, this change also allows for future optimizations: The freed page metas remain in the pages trees and remain associated with the same virtual address. A "refault" - faulting on a page which was previously evicted - causes a page fill and return to the new list. Refault statistics are kept persistently across eviction / fault cycles, and can be used to improve cache performance by reducing the number of redundant faults for a frequently-used page.

sanderssj:
Closed by #1282

PR #1282, opened 2020-09-14
Refactor pagecache evict

sanderssj:
This change addresses issue #1253 by moving evicted pages to a free list when no longer used, but leaving their meta in the node's tree. When the page is accessed again, memory is reallocated. This change also fixes a missing reserve_refcount in map_pages.

### DIFF
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -112,22 +112,13 @@ static inline void change_page_state_locked(pagecache pc, pagecache_page pp, int
 {
     int old_state = page_state(pp);
     switch (state) {
-#if 0
-    /* Temporarily disabling use of free until we have a scheme to
-       keep and act on "refault" data */
     case PAGECACHE_PAGESTATE_FREE:
-        assert(old_state == PAGECACHE_PAGESTATE_EVICTED);
-        pagelist_enqueue(&pc->free, pp);
-        break;
-#endif
-    case PAGECACHE_PAGESTATE_EVICTED:
         if (old_state == PAGECACHE_PAGESTATE_NEW) {
-            pagelist_remove(&pc->new, pp);
+            pagelist_move(&pc->free, &pc->new, pp);
         } else {
             assert(old_state == PAGECACHE_PAGESTATE_ACTIVE);
-            pagelist_remove(&pc->active, pp);
+            pagelist_move(&pc->free, &pc->active, pp);
         }
-        /* caller must do release following state change to evicted */
         break;
     case PAGECACHE_PAGESTATE_ALLOC:
         assert(old_state == PAGECACHE_PAGESTATE_FREE);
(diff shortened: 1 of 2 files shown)


sanderssj:
The io_uring test sometimes fails with a page fault, suggesting there's still some refcount mishandling. I need to find out what causing that problem before merging to master.

wjhun (commented):
I thought that we should be able to re-instate the refcount_release in pagecache_write_sg_finish (line 518). But when I tried it, I consistently get a failure in the io_uring test:

```
assigned: 10.0.2.15                                                                                                                                                                                                                            
ffffffff8000a281        (pagecache_page_free + 0000000000000151/00000000000001ab)                                                                                                                                                              
ffffffff8000a993        (pagecache_write_sg_finish + 00000000000001d3/0000000000000c26)                                                                                                                                                        
ffffffff8001f980        (merge_join + 0000000000000040/0000000000000098)                                                                                                                                                                       
ffffffff8001f980        (merge_join + 0000000000000040/0000000000000098)                                                                                                                                                                       
ffffffff80066dc5        (virtio_scsi_io_done + 0000000000000025/00000000000000a2)                                                                                                                                                              
ffffffff80066577        (virtio_scsi_request_complete + 0000000000000027/000000000000008b)                                                                                                                                                     
ffffffff80068151        (virtqueue_service_vqmsgs + 00000000000001a1/00000000000001b8)                                                                                                                                                         
ffffffff80010362        (runloop_internal + 0000000000000292/000000000000032b)                                                                                                                                                                 
assertion pp->write_count == 0 failed in /share/src/nanovms/justin-nanos/src/kernel/pagecache.c: pagecache_page_free() on line 367; halt
```

Is this the same failure that you saw occur sporadically last week? (Edit: the remaining tests appear to pass without issue.)

src/kernel/pagecache.c
```diff
@@ -616,49 +614,86 @@ closure_function(1, 3, void, pagecache_write_sg,
...
+    vector v;
+    u64 evicted = 0;
+    u64 lastevicted = 0;
 
-    /* We could avoid taking both locks here if we keep drained page
-       objects around (which incidentally could be useful to keep
-       refault data). */
+    if ((v = allocate_vector(pc->h, DRAIN_ITER_MAX)) == INVALID_ADDRESS)
+        return 0;
+    while (evicted < pages) {
```
wjhun, line 680: Pages is being decremented while evicted is accumulating, so this should either change to pages > 0, or perhaps better yet, remove the pages decrement below and use MIN(pages - evicted, DRAIN_ITER_MAX). Also, keeping lastevicted isn't really necessary - could just check the return value of evict_pages_locked, break if 0, else add to evicted.

src/kernel/pagecache.c
```diff
@@ -453,7 +478,7 @@ closure_function(5, 1, void, pagecache_write_sg_finish,
 
     pagecache_lock_node(pn);
     pagecache_page pp = page_lookup_nodelocked(pn, pi);
-    if (bound(complete)) {
+    if (!is_ok(s) || bound(complete)) {
```
wjhun, line 481: Really minor, but I would update the debug below to indicate whether the error was on write (complete true) or read. Yet I would still record a read error in pv->write_error, for it is an error for the cache write operation as a whole.

wjhun (approved):
Just one issue with the loop in pagecache_drain, but otherwise looks good.

src/kernel/pagecache.c
sanderssj: Nice catch! I corrected and cleaned it up.

PR #1284, opened 2020-09-15
sigaction: add support for SA_RESTART flag

### LINKED ISSUE #930: implement support for SA_RESTART
System calls interrupted by signals cause immediate termination of the bottom half with a "nullify" flag, typically resulting in returning from the syscall -EINTR.

It's not clear to me whether this truly means such system calls should be restarted from the top half or whether the bottom half can just be delayed until after returning from signal handling. If the latter is the case, add support for retaining such bottom halves while temporarily suspending their execution, allowing any signal handler to be dispatched before finally finishing the system call.

#925

francescolavra:
When this flag is specified in a signal action, if the signal is raised while the recipient thread is blocked in a restartable syscall, the syscall is restarted internally by the kernel instead of returning -EINTR to userspace.
Restartable syscalls are those that don't have timeout values in their argument list.

Closes #930.

### DIFF
--- a/src/unix/unix_internal.h
+++ b/src/unix/unix_internal.h
@@ -249,8 +249,8 @@ typedef struct thread {
     /* blockq thread is waiting on, INVALID_ADDRESS for uninterruptible */
     blockq blocked_on;
 
-    /* set by file op completion; used to detect if blocking is necessary */
-    boolean file_op_is_complete;
+    /* set by syscall_return(); used to detect if blocking is necessary */
+    boolean syscall_complete;
 
     /* for waiting on thread-specific conditions rather than a resource */
     blockq thread_bq;
@@ -759,15 +759,10 @@ static inline sysreturn sysreturn_value(thread t)
     return (sysreturn)thread_frame(t)[FRAME_RAX];
 }
 
-static inline void file_op_begin(thread t)
-{
-    t->file_op_is_complete = false;
-}
-
-static inline sysreturn file_op_maybe_sleep(thread t)
+static inline sysreturn thread_maybe_sleep_uninterruptible(thread t)
 {
     u64 flags = irq_disable_save(); /* XXX mutex / spinlock */
-    if (!t->file_op_is_complete) {
+    if (!t->syscall_complete) {
         /* leave ints disabled... */
         thread_sleep_uninterruptible();
     }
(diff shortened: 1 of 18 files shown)


sanderssj (commented):
This looks okay to me! The only question I have is whether or not some of the other places that check for BLOCKQ_ACTION_NULLIFY should return ERESTARTSYS instead of EINTR or EAGAIN (for non-timeout cases), e.g. epoll or futex.

francescolavra:
OK @sanderssj, I added support for restarting syscalls that have a timeout parameter but are called with an infinite timeout value. This allowed me to spot a bug in io_getevents(), and to fix the wrong -EINVAL value that was being returned by poll() and friends.

sanderssj (commented):
Looks good, thanks!

src/unix/unix_internal.h
```diff
@@ -784,6 +784,13 @@ static inline void file_op_maybe_wake(thread t)
     irq_restore(flags);
 }
 
+static inline sysreturn syscall_return(thread t, sysreturn val)
+{
+    set_syscall_return(t, val);
+    file_op_maybe_wake(t);
```
wjhun, line 790: I guess you could just merge these two functions and nuke file_op_maybe_wake (there's only one use left in sendfile_bh, and that could be changed to if (bound(bh)) syscall_return(...)). Then really "file_op_is_complete" could be something more generic like "syscall_complete", and file_op_maybe_sleep could be something like "thread_maybe_sleep_uninterruptible". Then nuke file_op_begin and universally set syscall_complete to false in the syscall_debug.

In general I think it would be helpful / less confusing to unify the syscall exit points ... maybe all calls to set_syscall_return should change to syscall_return...

wjhun (approved):
Looks good. I guess there's no use for BLOCKQ_ACTION_BLOCKED anymore?

francescolavra:
> Looks good. I guess there's no use for BLOCKQ_ACTION_BLOCKED anymore?

There is still a potentially good use for it, and that is in blockq_handle_completion(), which has been written to deal with I/O completion handlers that trigger another I/O operation that in turn uses the blockq API. Maybe this currently is not strictly necessary, but may become necessary in case we want the blockq API to be concurrency-safe (and so protected by a lock).

src/unix/unix_internal.h
francescolavra: OK, I nuked the file_op_* stuff and now things look more intuitive and consistent.
As for changing all calls to set_syscall_return() to calls to syscall_return(), maybe that's not ideal from an efficiency perspective, because in most cases set_syscall_return() is called by code that does a direct (non-blocking) return from a syscall, and we can avoid the overhead of disabling the irqs and all the rest. In those cases I think we should just return the syscall return value and be done with it, because set_syscall_return() is already called by syscall_debug()

PR #1286, opened 2020-09-17
spool net direct output using a periodic timer

wjhun:
This fixes a regression in trace downloading via HTTP by using a periodic runloop timer to send spooled data over a direct TCP connection.

### DIFF
--- a/src/net/direct.c
+++ b/src/net/direct.c
@@ -18,7 +18,7 @@ typedef struct direct {
 
 typedef struct direct_conn {
     direct d;
-    boolean running;            /* background processing started */
+    timer send_timer;
     struct list l;              /* direct list */
     struct tcp_pcb *p;
     struct list sendq_head;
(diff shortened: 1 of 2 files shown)


src/net/direct.c
```diff
@@ -118,12 +120,12 @@ closure_function(1, 1, status, direct_conn_send,
...
-                thunk t = closure(dc->d->h, direct_conn_send_bh, dc);
-                dc->running = true;
-                assert(enqueue(runqueue, t));
-            }
+        if (!dc->send_timer && !direct_conn_send_internal(dc)) {
+            timer_handler th = closure(dc->d->h, direct_conn_send_timer, dc);
+            assert(th != INVALID_ADDRESS);
+            dc->send_timer = register_timer(runloop_timers, CLOCK_ID_MONOTONIC, DIRECT_CONN_SEND_INTERVAL,
+                                            false, true, th);
+            assert(dc->send_timer != INVALID_ADDRESS);
```
sanderssj, line 128: Is there a reason we can't just use lwip's tcp_sent on the direct connection to specify a callback when send buffer frees up instead of using the timer and retry the send there?

wjhun: You're absolutely right. Going to close this and re-open another PR...

wjhun:
closing in favor of #1287

PR #1287, opened 2020-09-17
net direct: implement tcp_sent callback

wjhun:
Fix regression in net direct output by using a tcp_sent callback to send spooled data.

supersedes #1286

### DIFF
--- a/src/net/direct.c
+++ b/src/net/direct.c
@@ -18,7 +18,7 @@ typedef struct direct {
 
 typedef struct direct_conn {
     direct d;
-    boolean running;            /* background processing started */
+    struct spinlock send_lock;
     struct list l;              /* direct list */
     struct tcp_pcb *p;
     struct list sendq_head;
@@ -33,11 +33,16 @@ typedef struct qbuf {
 } *qbuf;
 
 /* return true if sendq empty */
-static boolean direct_conn_send_internal(direct_conn dc)
+static void direct_conn_send_internal(direct_conn dc)
 {
     direct_debug("dc %p\n", dc);
     list next;
 
+    /* It appears TCP_EVENT_SENT is only called from tcp_input. If in
+       the future it could ever be invoked as a result of a call to
+       tcp_write or tcp_output, this will need to be revised to avoid
+       deadlock. */
+    spin_lock(&dc->send_lock);
     while ((next = list_get_next(&dc->sendq_head))) {
         qbuf q = struct_from_list(next, qbuf, l);
         if (!q->b) {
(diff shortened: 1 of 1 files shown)


src/net/direct.c
```diff
@@ -33,11 +33,16 @@ typedef struct qbuf {
...
 {
     direct_debug("dc %p\n", dc);
     list next;
 
+    /* It appears TCP_EVENT_SENT is only called from tcp_input. If in
+       the future it could ever be invoked as a result of a call to
+       tcp_write or tcp_output, this will need to be revised to avoid
+       deadlock. */
+    spin_lock(&dc->send_lock);
     while ((next = list_get_next(&dc->sendq_head))) {
```
sanderssj, line 46: Locking on sendq_head seems inconsistent--direct_conn_send doesn't use the spinlock, but does disable interrupts, and here it's vice versa

src/net/direct.c
```diff
@@ -33,11 +33,16 @@ typedef struct qbuf {
 } *qbuf;
 
 /* return true if sendq empty */
```
sanderssj, line 35: Comment no longer valid

src/net/direct.c
wjhun: Thanks for catching that. I moved the qbuf enqueue under the locked region inside direct_conn_send_internal.

src/net/direct.c
```diff
@@ -43,6 +42,8 @@ static void direct_conn_send_internal(direct_conn dc)
        tcp_write or tcp_output, this will need to be revised to avoid
        deadlock. */
     spin_lock(&dc->send_lock);
+    if (q)
+        list_insert_before(&dc->sendq_head, &q->l);
     while ((next = list_get_next(&dc->sendq_head))) {
         qbuf q = struct_from_list(next, qbuf, l);
```
sanderssj, line 48: It doesn't matter really, but shadow assignment not really needed.

src/net/direct.c
```diff
@@ -43,6 +42,8 @@ static void direct_conn_send_internal(direct_conn dc)
        tcp_write or tcp_output, this will need to be revised to avoid
        deadlock. */
     spin_lock(&dc->send_lock);
+    if (q)
+        list_insert_before(&dc->sendq_head, &q->l);
```
sanderssj, line 46: I might be missing something obvious, but why isn't this an append?

src/net/direct.c
wjhun: Sorry, I'm not sure what you mean here...that q is defined again within this scope? I guess I didn't realize I was reusing that name for the arg...but it shouldn't hurt.

src/net/direct.c
wjhun: This does add to the list, as the element is being placed "before" the list head (pointed to by head->prev). I am in favor of revising the list interface though...it could be much clearer.

sanderssj: That's my bad! I didn't look closely enough at what a list head actually represents.

sanderssj:
Looks good to me!

Issue #164, opened 2018-10-24
Lwip NOSYS 0 should be used


tijoytom-zz:
Right now we use Lwip with no OS mode, which is fine for now since we run only on single core.But eventually we want to use LWip with OS mode

francescolavra:
As of now, all lwIP calls are made from code (syscalls, bottom halves, timers) under the kernel lock, and thus we are MP-safe.
If we want to take (some of) this code out of the kernel lock, I see two alternatives:
- switch to NO_SYS=0: this requires 1) adding support for kernel threads (because the lwIP core would be run from the `tcpip_thread`); 2) implementing synchronization primitives such as `sys_mutex_lock()` that would be called from lwIP; 3) wrap all lwIP calls in `LOCK_TCPIP_CORE()`/`UNLOCK_TCPIP_CORE()`, or alternatively modify them to use `tcpip_api_call()` (for example, instead of calling tcp_write(), we need to set up a struct holding function call argument values, implement a helper function that takes these values from the struct and calls tcp_write(), and call tcpip_api_call() by passing it the helper function pointer and the struct pointer)
- keep using NO_SYS=1, and implement our own locking (e.g. define a global lwIP spinlock, and use it at each lwIP call).

To me, the second alternative seems simpler and more efficient.

PR #1289, opened 2020-09-21
runloop: service the bhqueue before taking kernel lock

wjhun:
These changes move bhqueue service outside of the locked region in runloop_internal, necessary for storage completions to execute while a syscall continuation, holding the kernel lock, awaits a file-backed page to be faulted in. This divides the various I/O completions around the system between those that queue to bhqueue (storage completions) - which run without the kernel lock held - and ones that queue to runqueue (e.g. net / lwIP processing) and execute with the kernel lock held.

Also:
- spin locks on a non-SMP build are changed from nops to sanity checks
- address a deadlock issue when attempting to deallocate a physical page in pagecache_unmap_page_nodelocked

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -178,6 +178,7 @@ SRCS-kernel.img+= \
 	$(ARCHDIR)/ftrace.s
 endif
 
+CFLAGS+=	-DSPIN_LOCK_DEBUG_NOSMP
 #CFLAGS+=	-DLWIPDIR_DEBUG -DEPOLL_DEBUG -DNETSYSCALL_DEBUG -DKERNEL_DEBUG
 AFLAGS+=	-felf64 -I$(OBJDIR)/
 LDFLAGS+=	$(KERNLDFLAGS) --undefined=_start -T linker_script
--- a/src/hyperv/include/vmbus.h
+++ b/src/hyperv/include/vmbus.h
@@ -141,11 +141,11 @@ typedef void	(*vmbus_chan_callback_t)(struct vmbus_channel *, void *);
  *			memory passed through 'br'.
  */
 void		vmbus_chan_open(struct vmbus_channel *chan,
-		    int txbr_size, int rxbr_size, const void *udata, int udlen,
-		    vmbus_chan_callback_t cb, void *cbarg);
+                                int txbr_size, int rxbr_size, const void *udata, int udlen,
+                                vmbus_chan_callback_t cb, void *cbarg, queue sched_queue);
 int		vmbus_chan_open_br(struct vmbus_channel *chan,
-		    const struct vmbus_chan_br *cbr, const void *udata,
-		    int udlen, vmbus_chan_callback_t cb, void *cbarg);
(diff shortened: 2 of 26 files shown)


sanderssj:
This looks fine to me! I looked for things that might be mp unsafe if no longer protected by the kernel lock and while most things were locked or used lockless queues, it looked like ata_pci_service_req in ata-pci.c might do some queue and pci writes outside of a lock. You might want to double check that.

francescolavra (commented):
As far as I can see, the ATA PCI code is MP-safe, because its internal logic ensures that ata_pci_service_reqs() is only called when the ATA controller is idle (unless the controller generates more than one interrupt after one I/O request, which should never happen).
But I see another problem with servicing the bhqueue without the kernel lock, and that is related to lwIP, see https://github.com/nanovms/nanos/issues/164.
Where is it that we wait for a file-backed page to be faulted in while holding the kernel lock? From what I can see, the scheduler always drops the lock (in `kernel_sleep()`) before putting a CPU to sleep.

wjhun:
> internal logic ensures that ata_pci_service_reqs() is only called when the ATA controller is idle (unless the controller generates more than one interrupt after one I/O request, which should never happen)

I see that it is only invoked after a r/w operation if going from idle -> busy, but it is also invoked by the service routine which is enqueued to bhqueue by an interrupt - which runs the routine without checking for idle. So I think it might be possible for that queued service to run on one CPU (outside of the kernel lock) while another CPU is issuing a request. From inspection I can't really be certain that there isn't a race there. It would be much clearer if there was a single lock covering the whole service - right now the lock seems to just be protecting the reqs list.

> But I see another problem with servicing the bhqueue without the kernel lock, and that is related to lwIP, see #164.
Where is it that we wait for a file-backed page to be faulted in while holding the kernel lock? From what I can see, the scheduler always drops the lock (in kernel_sleep()) before putting a CPU to sleep.

Ok, there's a nuance here that is actually missing from this PR because I thought it had already been merged with the pagecache work.

The intention here is to have storage completions go to the bhqueue only, for they need to complete outside of the scope of the kernel lock. Completions for network devices, on the other hand, should go to the runqueue instead. Thus, lwIP code should be running off of the runqueue _only_, being protected by the kernel lock.

I will update the PR with the necessary changes.

> From what I can see, the scheduler always drops the lock (in kernel_sleep()) before putting a CPU to sleep.

As for holding the kernel lock, the path is that some syscall continuation touches a user buffer, causes a page fault for a non-present page and then suspension of the running kernel context. There is no return to the runloop, but rather a call to runloop, so the lock should not be dropped from within runloop_internal.

But the lock drop in kernel_sleep() should be removed, though. Thanks for pointing that out.

francescolavra:
> So I think it might be possible for that queued service to run on one CPU (outside of the kernel lock) while another CPU is issuing a request.

Whenever a request is issued, ata_pci_io() calls ata_pci_service_reqs() only if the reqs list was previously empty, and the list being previously empty guarantees that the queued service is not running.

francescolavra:
Looks good to me.

eyberg:
this works on azure for me

PR #1292, opened 2020-09-22
Misc fixes in preparation for enabling SMP support

francescolavra:
Since the kernel lives at the top of the 64-bit address space, the GDT address used at runtime does not fit in 32 bits, so a "temporary" GDT is needed in the AP boot page for secondary CPUs to use when switching to 64-bit mode.
Handling of the /sys/devices/system/cpu/online special file has been updated to report the correct number of available CPUs (its
events callback has been changed so that the EPOLLIN flag is always returned, which is the correct behavior for files that do not block when read).

### DIFF
--- a/src/x86_64/lock.h
+++ b/src/x86_64/lock.h
@@ -2,7 +2,7 @@ typedef struct spinlock {
     word w;
 } *spinlock;
 
-#ifdef SMP_ENABLE
+#if defined(KERNEL) && defined(SMP_ENABLE)
 static inline boolean spin_try(spinlock l) {
     u64 tmp = 1;
     /* Is it worth the compare and branch to skip a pause? */
--- a/src/unix/special.c
+++ b/src/unix/special.c
@@ -82,34 +82,18 @@ static u32 maps_events(file f)
     return EPOLLIN;
 }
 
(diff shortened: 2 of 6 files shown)


src/x86_64/lock.h
```diff
@@ -17,7 +17,8 @@ static inline void spin_lock(spinlock l) {
 
 static inline void spin_unlock(spinlock l) {
     compiler_barrier();
-    *(volatile u64 *)&l->w = 0;
+    volatile word *w = &l->w;
```
wjhun, line 20: I found the type aliasing error was being caused by stage2 getting compiled with these - but stage2 doesn't need spinlocks. Were you also getting those errors during stage2 build?

I suggest changing #ifdef SMP_ENABLE to #if defined(KERNEL) && defined(SMP_ENABLE).

wjhun (approved):
Looks good...see note about spinlocks getting built for stage2.

src/x86_64/lock.h
francescolavra: Indeed the error only occurs when building stage2. I reverted the change and used `#if defined(KERNEL) && defined(SMP_ENABLE)` instead.

PR #1293, opened 2020-09-23
Fixed xennet driver handling of multipage packets

sanderssj:
I found that the xennet driver was sending out corrupted/incomplete
packets in the case where the packet consists of multiple pages.
There were two problems: first, the end flag was incorrectly set for
each tx page. Second, xen wants the length of the first tx descriptor
for the packet to be the length of the entire packet.


francescolavra (commented):
Looks good to me

wjhun (approved):
Looks right to me.

PR #1297, opened 2020-09-27
SMP fixes

### LINKED ISSUE #1271: accessing current thread in syscall continuations
Continuations for syscalls typically enclose the invoking thread, referring to it in place of 'current' (usually, 't'). While the thread pointer is accessible from within the closure body of the continuation, there are other contexts in which it is needed. For example, if a syscall continuation touching some user buffer causes a protection fault, the fault handler will attempt to send a signal to current. If current was not restored before continuation, it will end up delivering the fault to the thread last scheduled to run on the current cpu.

With #1269, an active thread has a pause method invoked to settle up system time accounting. Yet no resumption of accounting occurs on a continuation - only the next return to userspace. It's not clear where such a resume method would be called, since syscall continuations can be served from the scheduling queues just as well as one of the several "completions vector" service routines or some other structure like a vqmsg - there isn't a single point where these continuations (which could be thunks, status_handlers ... anything) are dispatched that can also be used to restore current and apply a resume method.

A brute force approach is to just manually restore current to the enclosed thread at the head of each continuation _within unix_. There could still be an issue if some syscall - e.g. sendfile - leads to a chain of asynchronous operations outside the scope of unix, where no thread context is being maintained. A fault on a user buffer anywhere in that sequence could not be reliably associated with the offending thread, for another thread may have been scheduled between a request and a completion.

Another idea is to enqueue each completion as two words -  together as a completion type, perhaps, which includes the completion as well as a thread address (or invalid). (The lock free queue is currently only word-sized elements, but variants with a size parameter can be added.) Since completions go into a variety of containers - queue, vector, vqmsq, etc. - the extra housekeeping can be done by helpers: save saves current into the thread field and calls its pause method. Restore restores current from the thread field and calls its resume method.

#1254, #1269

francescolavra:
This PR does a couple more changes needed for the kernel to run in SMP mode.

Closes #1271.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -340,7 +340,6 @@ extern boolean shutting_down;
 void kernel_shutdown(int status)
 {
     shutting_down = true;
-    apic_ipi(TARGET_EXCLUSIVE_BROADCAST, 0, shutdown_vector);
     if (root_fs) {
         storage_sync(closure(heap_general(&heaps), sync_complete, status));
         runloop();
@@ -351,7 +350,6 @@ void kernel_shutdown(int status)
 void kernel_shutdown_ex(status_handler completion)
 {
     shutting_down = true;
-    apic_ipi(TARGET_EXCLUSIVE_BROADCAST, 0, shutdown_vector);
     if (root_fs) {
         storage_sync(completion);
         runloop();
--- a/src/kernel/kernel.h
+++ b/src/kernel/kernel.h
@@ -52,9 +52,6 @@ typedef struct cpuinfo {
 
     /* Stack for interrupts */
     void *int_stack;
-
-    /* pointer to nanos_thread (and target OS dependent stuff) */
-    nanos_thread current_thread;
 } *cpuinfo;
 
 #define cpu_not_present 0
@@ -98,6 +95,18 @@ static inline context get_running_frame(void)
 static inline void set_running_frame(context f)
 {
     current_cpu()->running_frame = f;
+}
+
+static inline nanos_thread get_current_thread()
+{
+    context f = current_cpu()->kernel_context->frame;
+    return pointer_from_u64(f[FRAME_THREAD]);
+}
+
+static inline void set_current_thread(nanos_thread t)
+{
+    context f = current_cpu()->kernel_context->frame;
+    f[FRAME_THREAD] = u64_from_pointer(t);
 }
 
 static inline void *stack_from_kernel_context(kernel_context c)
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -13,6 +13,9 @@
 #include <page.h>
 #else
 #include <runtime.h>
+typedef void *nanos_thread;
+#define get_current_thread()    0
+#define set_current_thread(t)
 #endif
 
 #include <pagecache.h>
@@ -460,8 +463,8 @@ static void touch_or_fill_page_by_num_nodelocked(pagecache_node pn, u64 n, merge
         touch_or_fill_page_nodelocked(pn, pp, m);
 }
 
-closure_function(5, 1, void, pagecache_write_sg_finish,
-                 pagecache_node, pn, range, q, sg_list, sg, status_handler, completion, boolean, complete,
+closure_function(6, 1, void, pagecache_write_sg_finish,
+                 nanos_thread, t, pagecache_node, pn, range, q, sg_list, sg, status_handler, completion, boolean, complete,
                  status, s)
 {
     pagecache_node pn = bound(pn);
@@ -536,6 +539,7 @@ closure_function(5, 1, void, pagecache_write_sg_finish,
     } else {
         write_sg = 0;
     }
+    set_current_thread(bound(t));
     do {
         assert(pp != INVALID_ADDRESS && page_offset(pp) == pi);
         u64 copy_len = MIN(q.end - (pi << page_order), cache_pagesize(pc)) - offset;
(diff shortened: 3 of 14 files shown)


src/kernel/kernel.h
```diff
@@ -100,6 +97,18 @@ static inline void set_running_frame(context f)
     current_cpu()->running_frame = f;
 }
 
+static inline nanos_thread get_current_thread()
+{
+    context f = current_cpu()->kernel_context->frame;
+    return pointer_from_u64(f[FRAME_THREAD]);
```
wjhun, line 103: Just curious as to your reasoning for moving these back to the kernel context frame. I went back and forth between the two. It does seem that a kernel page fault on a user page in a syscall continuation could result in the continuation being resumed on another core...which would indeed advocate placing the thread in the frame.

src/kernel/pagecache.c
```diff
@@ -596,7 +600,8 @@ closure_function(1, 3, void, pagecache_write_sg,
         pn->length = q.end;
 
     /* prepare pages for writing */
-    merge m = allocate_merge(pc->h, closure(pc->h, pagecache_write_sg_finish, pn, q, sg, completion, false));
+    merge m = allocate_merge(pc->h, closure(pc->h, pagecache_write_sg_finish,
+        get_current_thread(), pn, q, sg, completion, false));
```
wjhun, line 604: This seems reasonable, but shouldn't such restoration happen on the read path as well?

wjhun (commented):
Would not most instances of thread_resume in syscall continuations be better served within blockq.c, e.g. in blockq_apply_bi_locked?

platform/pc/service.c
```diff
@@ -340,7 +340,6 @@ extern boolean shutting_down;
 void kernel_shutdown(int status)
 {
     shutting_down = true;
-    apic_ipi(TARGET_EXCLUSIVE_BROADCAST, 0, shutdown_vector);
```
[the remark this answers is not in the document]
francescolavra: It's explained in the commit message, basically if there is some data to be flushed to storage and the relevant I/O completion interrupt is delivered to a halted core, the kernel stalls.
Why would we need to halt the cores on shutdown? The VM is going to be stopped anyway.

src/kernel/kernel.h
francescolavra: Yes, as I wrote in the commit message, a kernel context can be moved between cores, so the current thread should be stored in the frame of the kernel context.

src/kernel/pagecache.c
francescolavra: In the read path, SG buffers are allocated by the kernel, and the pagecache code never touches user buffers. It's the completion routine that should restore the thread before copying data to user buffers (as done e.g. by iov_read_complete() and file_read_complete()).

francescolavra:
Indeed restoring the current thread in blockq.c would cover most instances (the thread_resume() calls in poll.c would still be needed, as they use notify entries instead of block queues).
Do you prefer moving the thread_resume() calls (where applicable) to blockq_apply_bi_locked()?

platform/pc/service.c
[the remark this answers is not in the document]
wjhun: Sorry, I didn't see your commit description as it was folded. That makes sense about not stopping cpu 0 before a flush as interrupts may arrive on it, and the case of a fault/panic is covered. I question not leaving the cores in a quiescent state, but I guess I don't have a good argument for doing so (shutdown set to soft reboot?) ... so go ahead.

src/kernel/pagecache.c
wjhun: Right, my mistake.

wjhun:
> Indeed restoring the current thread in blockq.c would cover most instances (the thread_resume() calls in poll.c would still be needed, as they use notify entries instead of block queues).
> Do you prefer moving the thread_resume() calls (where applicable) to blockq_apply_bi_locked()?

Is there any reason not to? It would eliminate most instances of thread_resume here.

francescolavra:
> > Indeed restoring the current thread in blockq.c would cover most instances (the thread_resume() calls in poll.c would still be needed, as they use notify entries instead of block queues).
> > Do you prefer moving the thread_resume() calls (where applicable) to blockq_apply_bi_locked()?
> 
> Is there any reason not to? It would eliminate most instances of thread_resume here.

The only reason I can see is that individual blockq actions can resume the thread only when necessary (i.e. when they are going to access a user buffer). But I don't have a strong opinion as to which way is better.

wjhun:
> The only reason I can see is that individual blockq actions can resume the thread only when necessary (i.e. when they are going to access a user buffer). But I don't have a strong opinion as to which way is better.

But thread resumption should also be resuming system time accounting on behalf of the thread, so would we not want to always resume when possible once a continuation chain moves back into the realm of a syscall? It may also be less error-prone to resume in the blockq, as it would eliminate any possibility of a call to thread_resume being omitted in an action.

francescolavra:
> But thread resumption should also be resuming system time accounting on behalf of the thread, so would we not want to always resume when possible once a continuation chain moves back into the realm of a syscall? It may also be less error-prone to resume in the blockq, as it would eliminate any possibility of a call to thread_resume being omitted in an action.

OK, agreed, I'm going to make this change.

sanderssj:
This change does break ftrace, where current is now null inside the ftrace_return_to_handler. That might imply a missing thread_resume someplace. I'm just starting to look at it, so I'll get some more info.

sanderssj:
I'm still working on getting ftrace going again with this change, but I don't want to hold it up any longer and I think most of the changes will be limited to ftrace.c. There may be a need to copy the FRAME_THREAD value to the spare_kernel_context in suspend_kernel_context, but I'm not totally sure it's necessary yet. Otherwise looks good to me!

francescolavra:
OK, I made the change suggested by @wjhun, now the current thread is restored whenever a blockq action is called.
As a result of this change, the current thread always coincides with the thread associated to a blockq action being called: this required a small change in the implementation of the futex BH handler, which otherwise would never unblock a blocked thread unless a timeout condition occurs.

wjhun (approved):
Looks good

PR #1298, opened 2020-09-27
Remove unnecessary lock in xenblk_get_rreq

sanderssj:
I ran across this testing on AWS. The lock was getting reacquired causing an assertion failure. The lock is unnecessary here, as it is already acquired in the grandparent caller.


francescolavra (approved):
Right, this was my fault.

PR #1300, opened 2020-09-30
uintptr_t type for future Mingw64 tools build

mkhon:


wjhun (changes requested):
We almost entirely use our own types; these instances of unsigned long and long just slipped through. Rather than add yet a new type, I would prefer that these instances be changed to u64 and s64, respectively.

eyberg:
i chatted w/ max & will on this and some end-users, plus I've already found out how to deal with this in wsl - https://github.com/nanovms/ops/issues/282 - I think we should prob. just not support this

PR #1301, opened 2020-09-30
Rename io.h -> inout.h in order to avoid header conflicts under mingw64

mkhon:


wjhun (approved):

eyberg (changes requested):
not having used windows forever, I'm kinda wondering if we should explore wsl2 first, if that gets us hyper-v orchestration and it is the preferred dev env then there's prob. no reason for this

eyberg:
i chatted w/ max & will on this and some end-users, plus I've already found out how to deal with this in wsl - https://github.com/nanovms/ops/issues/282 - I think we should prob. just not support this

PR #1303, opened 2020-09-30
Volume mounting: add support for volumes identified by a label

francescolavra:
In volume mount directives included in a manifest file, it is now possible to specify a volume by its filesystem label (alternatively
to using the filesystem UUID). This allows to easily change volumes to be mounted by a VM instance without re-creating the image for the instance.
Example of a mount directive using a label: `mounts:(my_label:/mnt)`.
The mkfs tool has an additional command line option (-l) that allows to specify a label (if not specified, the default label is an emtpy
string), and the dump tool displays the label retrieved from an existing filesystem.
Example to create an empty filesystem with label "my_label": `mkfs -l my_label -e volume.raw`.

### DIFF
--- a/src/runtime/text.h
+++ b/src/runtime/text.h
@@ -136,6 +136,21 @@ static inline character pop_character(buffer b)
     return c;
 }
 
+static inline boolean read_cstring(buffer b, char *dest, bytes maxlen)
+{
+    bytes count = 0;
+    u8 c;
+    do {
+        if (buffer_length(b) == 0)
+            return false;
+        c = *(u8 *)buffer_ref(b, 0);
+        bytes len = utf8_length(c);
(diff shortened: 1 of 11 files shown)


src/runtime/text.h
```diff
@@ -136,6 +136,19 @@ static inline character pop_character(buffer b)
     return c;
 }
 
+static inline boolean read_cstring(buffer b, char *dest, bytes maxlen)
+{
+    bytes count = 0;
+    u8 c;
+    do {
+        c = *(u8 *)buffer_ref(b, 0);
```
wjhun, line 144: check buffer_length(b) before dereference

wjhun (approved):
Looks good

Issue #1302, opened 2020-09-30
track safety of interrupt handler and bhqueue processing paths


wjhun:
These notes were made in an attempt to audit all execution paths that stem from interrupt handlers through enqueues to bhqueue and subsequent bhqueue service, as well as the request paths originating under the kernel lock that might possibly contend with these paths that run outside of the lock.

Most issues found here concern a lack of safety with heap allocation, primarily from heap (de)allocations in bh service as well as deallocation of objects upstream as a result of closure_finish() or a refcount release. There are a handful of allocations at interrupt level as well. The proposed solution here is to add spinlocks to lower-level heaps (physical - which is already protected by a lock in page.c, virtual_huge, virtual_page, and backed) and add a lock-protected mcache for memory that might be (de)allocated in bh service (or in interrupt handlers, though I would personally prefer to see these removed altogether if possible). The 'general' mcache will remain unlocked for use under the kernel lock only (for which we can add debug assertions that the kernel lock is in fact held during (de)allocs). The places that require use of a locked heap are indicated by the TODO items below (as well as some that otherwise require additional locking, or might just need verification).

BUG items indicate general bugs that were found during the audit, and COMPLETION items highlight the areas where a status handler completion is invoked (no action required for these - added for reference).

While there are many TODO items listed here, most would be cleared out by the use of locked heaps.

Note that completions up into the syscall level are not covered here; I think it will be possible to invoke these from the runqueue with a minor change to the per-page completion handling. Ideally, we can isolate all these safety issues to the pagecache level and lower.

drivers
-------
- ata-pci: apci->service (ata_pci_service)
  - ata_pci_service_reqs()
    - ata_pci_service_req()
      - ~~debugs msgs could alloc, so maybe rprintf & friends should
        use locked heap (not performance sensitive anyway)~~ actually they use stack allocations
      - ata_io_cmd_dma()
        - ata_set_lba()
          - kernel_delay
            - [x] no allocs, but calls now() and kern_pause - must be
              safe outside kern lock - TODO
      - [x] may call timm() on error - so errheap in runtime_init.c
        should use locked heap - TODO
      - ata_io_cmd() (for pio fallback)
        - also may call timm()
        - ata_wait()
          - ata pio and kernel_delay, covered above
        - status handler invoked from here - COMPLETION
      - status handler invoked on error or request completion - COMPLETION
    - deallocates ata_pci_req
      - [x] dev->h needs to be locked heap - TODO

- storvsc: vmbus_chan_open
[...]

PR #1304, opened 2020-10-01
more smp fixes

wjhun:
- remove sloppy uses of ci->have_kernel_lock on lock acquisition; flag should instead be used for debug assertions
- explicitly release kernel lock on calls to runloop (aside from kernel page fault)
- clear ci->have_kernel_lock before suspending kernel context on page fault, reset on context resume
- allow pagecache I/O completions to be served on either bhqueue or runqueue depending on origin of request
- change vq_interrupt to enqueue to vq->sched_queue instead of bhqueue
- fix invalid dereference of current in syscall_schedule

While stability issues remain when running with SMP (see #1302 for known issues), webg and other tests progress much further now.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -57,7 +57,8 @@ define_closure_function(5, 0, void, thread_demand_file_page,
     vmap vm = bound(vm);
     pagecache_node pn = vm->cache_node;
     pagecache_map_page(pn, bound(node_offset), bound(page_addr), bound(flags),
-                       (status_handler)&bound(t)->demand_file_page_complete);
+                       (status_handler)&bound(t)->demand_file_page_complete,
+                       false /* complete on runqueue */);
     range ra = irange(bound(node_offset) + PAGESIZE,
         vm->node_offset + range_span(vm->node.r));
     if (range_valid(ra)) {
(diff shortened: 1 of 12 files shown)


src/unix/mmap.c
```diff
@@ -121,14 +124,17 @@ boolean do_demand_page(u64 vaddr, vmap vm, context frame)
...
             pagecache_map_page(vm->cache_node, node_offset, page_addr, flags,
-                               (status_handler)&do_kernel_demand_pf_complete);
+                               (status_handler)&do_kernel_demand_pf_complete,
+                               true /* complete on bhqueue */);
             if (kernel_demand_page_completed) {
                 pf_debug("   immediate completion\n");
                 return true;
             }
             faulting_kernel_context = suspend_kernel_context();
+            current_cpu()->have_kernel_lock = false;
```
sanderssj, line 137: This is redundant with suspend_kernel_context, which also clears have_kernel_lock.

src/unix/syscall.c
```diff
@@ -2476,6 +2477,8 @@ void syscall_debug(context f)
     // i dont know that we actually want to defer the syscall return...its just easier for the moment to hew
     // to the general model and make exceptions later
     schedule_frame(f);
+  out:
+    kern_unlock();
```
sanderssj, line 2481: Okay, this might be a bit too nitpicky, but it might be nice to keep lock and unlock together, either move kern_try_lock in here, or else move unlock/runloop in syscall_schedule, and let it call runloop exclusively.

francescolavra:
Looks good to me.

src/unix/syscall.c
wjhun: In general I would agree about keeping lock acquire/release paired, but this isn't really possible when a scheduled thunk is invoked with the kernel lock already acquired (e.g. resume_syscall). For that matter, anything that is served off of the runqueue (or syscall handler) is invoked with the lock held, so a subsequent call to runloop (excepting kernel fault case) will require an (unpaired) unlock.

src/unix/mmap.c
wjhun: Ok, I removed the one in suspend_kernel_context for consistency (as kernel_frame_return sets it above).

PR #1307, opened 2020-10-09
build fixes

wjhun:
- fix parallel build breakage on fresh tree caused by redundant tools build
- maintain runtime source files in a single list (from https://github.com/convolvatron/nanos)


sanderssj (commented):
Looks good to me!

PR #1308, opened 2020-10-09
Use Xen event upcall mask vs individual hypercalls

sanderssj:
By using the xen event upcall mask we can eliminate the unmask hypercalls
in each interrupt handler, improving performance and latency. This changeset
continues to set the second level mask for spurious interrupts so those will
only interrupt once. My tests with webg and ab show around a 7.5% improvement
in requests per second and request latency for Xen.

### DIFF
--- a/src/xen/xen.c
+++ b/src/xen/xen.c
@@ -115,7 +115,7 @@ closure_function(0, 0, void, xen_interrupt)
                 } else {
                     /* XXX we have an issue with seemingly spurious interrupts at evtchn >= 2048... */
                     xenint_debug("  evtchn %d: spurious interrupt", i2);
-                    si->evtchn_mask[bit1] |= 1ULL<<bit2;
+                    si->evtchn_mask[bit1] |= U64_FROM_BIT(bit2);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/xen/xen.c
```diff
@@ -115,9 +115,11 @@ closure_function(0, 0, void, xen_interrupt)
                 } else {
                     /* XXX we have an issue with seemingly spurious interrupts at evtchn >= 2048... */
                     xenint_debug("  evtchn %d: spurious interrupt", i2);
+                    si->evtchn_mask[bit1] |= 1ULL<<bit2;
```
wjhun, line 118: minor: can use U64_FROM_BIT(bit2) here

wjhun (approved):
Looks good, and nice performance boost. If I understand correctly, as it is a per-vcpu flag, evtchn_upcall_mask will mask upcalls only on the current CPU, so this could pose an issue if we ever handle Xen events on other CPUs - in which case the shared info evtchn_mask might still need to be set. But that's not the case for us now, and I think the significant performance boost justifies making this change for now.

PR #1309, opened 2020-10-15
PCI: add support for a driver to operate more than one device

francescolavra:
With this change, the core PCI code allocates a struct pci_dev for each discovered device, and keeps track of whether a driver has
been attached to it; in this way, the probe function is not called more than once for the same device, while allowing a given driver
to control more than one device. This is needed to support multiple storage devices instantiated as separate PCI devices.
Since now each PCI device has an allocated structure in the core code, there is no need for drivers to keep a local struct pci_dev.

### DIFF
--- a/src/kernel/pci.c
+++ b/src/kernel/pci.c
@@ -41,13 +41,9 @@
 #define PCICAP_NEXTPTR  0x1
 
 // use the global nodespace
+static vector devices;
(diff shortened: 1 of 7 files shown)


sanderssj (commented):
Looks good to me!

wjhun (approved):

PR #1311, opened 2020-10-18
Unix run_thread_frame(): avoid taking kernel lock

francescolavra:
The only function call that needed the kernel lock was thread_log_internal(), which is now being changed to allocate its buffer from the stack instead of a global heap; as a side effect, this fixes a memory leak in thread_log_internal(), because the buffer was never deallocated and the "transient" heap is not actually transient.


sanderssj:
Looks good!

wjhun (approved):

PR #1313, opened 2020-10-20
Refactor vector get/set to use array indexing

sanderssj:
Profiling revealed unusual amounts of time in vector ops doing memcpy.
Since vectors are used as arrays of pointers, it makes more sense to index
as an array of pointers rather than memcpy in and out. Doing so yields a
small performance gain in webg of around 1%. Not a lot, but the patch
is small. This does assume aligned memory access for best performance.

### DIFF
--- a/src/runtime/vector.h
+++ b/src/runtime/vector.h
@@ -2,22 +2,19 @@ typedef buffer vector;
 
 static inline void *vector_get(vector v, int offset)
 {
-    void *res;
     bytes base = v->start + offset * sizeof(void *);
     if ((offset < 0) || ((base + sizeof(void *)) > v->end))
         // should be INVALID_VIRTUAL (? )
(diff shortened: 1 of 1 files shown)


src/runtime/vector.h
```diff
@@ -2,22 +2,19 @@ typedef buffer vector;
 
 static inline void *vector_get(vector v, int offset)
 {
-    void *res;
     bytes base = v->start + offset * sizeof(void *);
```
wjhun, line 5: Minor nit involving existing code, but this function could be simpler. A vector is really a buffer, so you could perhaps just change offset to a byte offset, compare it against buffer_length, and if valid return (void **)buffer_ref(v, offset).

wjhun (approved):
Looks good

PR #1315, opened 2020-10-23
Fix thread hang

### LINKED ISSUE #1314: thread_test hanging while scaling up thread count
Changing nthreads in main() to 64 or 128 appears to cause a hang. It's sort of a heisenbug ... if I add some debug printfs (note: use setbuf(stdout, 0) to avoid buffering) in the child thread code it completes (at least up to n = 1024, haven't tried more), so I suspect there could be a futex or other scheduling-related issue here.

It would also be nice to compute the final value in terminus() as a function of the thread count - and allow the count to be specified on the command line.

sanderssj:
This change fixes #1314 by increasing the thread_queue length. When enough threads are 
scheduled (as can be in thread_test), the enqueue in schedule_frame was silently failing
causing the thread never to be run again. I have also added an assert for that enqueue
so that it will panic if too many threads are scheduled.

I also cleaned up thread_test so that it can take an argument of the number of threads to test and now
calculates the expected sum instead of using the hard coded values.

### DIFF
--- a/src/kernel/schedule.c
+++ b/src/kernel/schedule.c
@@ -184,7 +184,7 @@ void init_scheduler(heap h)
     /* scheduling queues init */
     runqueue = allocate_queue(h, 64);
     bhqueue = allocate_queue(h, 2048);
-    thread_queue = allocate_queue(h, 1024);
+    thread_queue = allocate_queue(h, MAX_THREADS);
     runloop_timers = allocate_timerheap(h, "runloop");
     assert(runloop_timers != INVALID_ADDRESS);
     shutting_down = false;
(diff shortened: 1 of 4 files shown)


src/kernel/schedule.c
```diff
@@ -184,7 +184,7 @@ void init_scheduler(heap h)
     /* scheduling queues init */
     runqueue = allocate_queue(h, 64);
     bhqueue = allocate_queue(h, 2048);
-    thread_queue = allocate_queue(h, 64);
+    thread_queue = allocate_queue(h, 1024);
```
wjhun, line 187: Perhaps we should define these queue lengths in config.h?

wjhun (approved):
Looks good!

PR #1321, opened 2020-11-03
in-kernel loadable module support

wjhun:
These changes introduce dynamically-loadable library modules ("klibs") into the kernel. Klibs are statically-built, PIE ELF binaries located in either the root or boot filesystems (specified in manifest) and may be loaded (and unloaded) on demand.

The loading process does not perform explicit dynamic linking into the kernel, instead favoring a dlopen-style interface for resolving symbols by name. Symbols exported from the kernel (using the KLIB_EXPORT macro) are resolved in the klib by name using a "get_sym" method passed as an argument to the module init function. An "add_sym" method is used to export symbols from the klib. These are resolved within the kernel using klib_sym().

This interface is designed to allow the loading of optional components, e.g. crypto libraries, without needing to link them directly into the kernel.

Other notes / fixes:

- Note that module loading is an asynchronous operation (which may also follow asynchronous mounting of the boot filesystem if klibs are placed there). This may occur after other kernel initialization and program start have occurred. This is the reason for the "sleep(1)" call in the klibs runtime test; without it, the VM exits before the boot filesystem finishes mounting. It may be necessary to hold up program start until this process has finished if, for some reason, a klib needs to be initialized before the program executes.
- A load_offset argument was added to add_elf_syms() to accommodate PIE executables and ASLR. (Note that the klib code reads in the elf symbols for extra context on a crash, not for resolving calls into the kmod by symbol name.)
- The unix default_fault_handler was not acting appropriately for faults on kernel addresses or when a thread context was not available. This was fixed to allow a more graceful crash dump and shutdown on an unrecoverable fault in the kernel.
- BOOTFS_SIZE needed to be expanded to accommodate the test klib. It will likely need to grow more as larger klibs are included. Perhaps we could look at a way to set this size dynamically according to staged content.

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -3,13 +3,18 @@ PROGRAMS=	test
 SRCS-test= \
 	$(CURDIR)/test.c
 
-CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC
-LDFLAGS+=	-pie -nostdlib -DKLIB -T$(ARCHDIR)/klib.lds
-
-ifeq ($(ARCH),x86_64)
-CFLAGS+=	-mcmodel=kernel
-endif
-
 all: $(PROGRAMS)
 
 include ../rules.mk
+
+ifeq ($(UNAME_s),Darwin)
+ELF_TARGET=     -target x86_64-elf
+CFLAGS+=        $(ELF_TARGET)
+LD=             x86_64-elf-ld
+OBJDUMP=        x86_64-elf-objdump
+else
+LD=             $(CROSS_COMPILE)ld
+endif
+
+CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC -DKLIB
+LDFLAGS+=	-pie -nostdlib -T$(ARCHDIR)/klib.lds
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -4,10 +4,11 @@ SRCS-test= \
 	$(CURDIR)/test.c
 
 CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC
+LDFLAGS+=	-pie -nostdlib -DKLIB -T$(ARCHDIR)/klib.lds
 
-# should prob grab from platform dir
-# verify flags, borrowed from vdso build
-LDFLAGS+=	-n -pie -nostdlib -shared -DKLIB -T$(ARCHDIR)/klib.lds
+ifeq ($(ARCH),x86_64)
+CFLAGS+=	-mcmodel=kernel
+endif
 
 all: $(PROGRAMS)
 
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -17,4 +17,7 @@ LD=             $(CROSS_COMPILE)ld
 endif
 
 CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC -DKLIB
+
+# TODO should add stack protection to klibs...
+CFLAGS+=	-fno-stack-protector
 LDFLAGS+=	-pie -nostdlib -T$(ARCHDIR)/klib.lds
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -1,0 +1,14 @@
+PROGRAMS=	test
+
+SRCS-test= \
+	$(CURDIR)/test.c
+
+CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC
+
+# should prob grab from platform dir
+# verify flags, borrowed from vdso build
+LDFLAGS+=	-n -pie -nostdlib -shared -DKLIB -T$(ARCHDIR)/klib.lds
+
+all: $(PROGRAMS)
+
+include ../rules.mk
--- a/klib/test.c
+++ b/klib/test.c
@@ -11,7 +11,9 @@ static int foo(int x)
     return z[0] + y; /* test data */
 }
 
-int init(void *md, klib_add_sym add_sym)
+void (*memset)(void *a, unsigned char b, unsigned long len);
+
+int init(void *md, klib_get_sym get_sym, klib_add_sym add_sym)
 {
     /* test that bss is clear */
     for (int i = 0; i < Z_LEN; i++)
@@ -20,5 +22,12 @@ int init(void *md, klib_add_sym add_sym)
 
     add_sym(md, "foo", foo);
     add_sym(md, "bar", (void*)0);
-    return KLIB_INIT_OK;
+
+    memset = get_sym("runtime_memset");
+    if (!memset)
+        return KLIB_INIT_FAILED;
+
+    unsigned long a = -1ull;
+    memset(&a, 0, sizeof(unsigned long));
+    return a == 0 ? KLIB_INIT_OK : KLIB_INIT_FAILED;
 }
--- a/klib/test.c
+++ b/klib/test.c
@@ -1,0 +1,14 @@
+#include <klib.h>
+
+#define INVALID_ADDRESS ((void*)-1ull)
+static int foo(int x)
+{
+    return x + 123;
+}
+
+int init(void *md, klib_add_sym add_sym)
+{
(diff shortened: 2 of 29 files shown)


francescolavra (commented):
> It may be necessary to hold up program start until this process has finished if a klib needs to be initialized before the program executes. A module located in the root filesystem won't have this issue.

Actually, as far as I can see, this issue could occur even if a module is in the root filesystem, because the module binary is loaded asynchronously just like the application binary, so I can imagine an application to be started before a module is initialized, especially if the module binary size is larger than the application binary.
Anyway, if we ever encounter situations where a program needs a kernel module to be loaded, I guess we could suspend the application thread(s) when they make a syscall that requires a yet-to-be-loaded kernel module.

wjhun:
> Actually, as far as I can see, this issue could occur even if a module is in the root filesystem

Yes, that's true. I updated the description.

> There should be a deallocate_rangemap(kl->mappings).

> Perhaps we can add another boolean parameter to bootfs_complete, which indicates whether the kernel symbols should be ingested.

> The format string doesn't match the argument list.

> This definition is not used in this file.

Thanks, I have updated with fixes.

test/runtime/klibs.manifest
```diff
@@ -0,0 +1,25 @@
+(
+    boot:(
+          children:(
+                    kernel:(contents:(host:output/platform/pc/bin/kernel.img))
```
francescolavra, line 4: Specifying the kernel image file here is redundant, since it's overwritten by mkfs when invoked with the -k option.

src/kernel/klib.h
```diff
@@ -0,0 +1,37 @@
...
+
+typedef struct klib {
+    char name[KLIB_MAX_NAME];
+    table syms;
+    rangemap mappings;
+    buffer elf;
+} *klib;
+
+typedef closure_type(klib_handler, void, klib, status);
+typedef int (*klib_init)(void *md, klib_add_sym add_sym);
```
francescolavra, line 19: How about adding a "kern_get_sym" parameter to this function, which can then be used by a module to load kernel symbols?
Commit https://github.com/nanovms/nanos/commit/cacc023798ffb8a70e2df3ffa0a9fb5bbbfb1557 from my (work in progress) https://github.com/nanovms/nanos/tree/feature/telemetry branch has a possible implementation for that.

src/kernel/stage3.c
```diff
@@ -193,38 +218,51 @@ thunk create_init(kernel_heaps kh, tuple root, filesystem fs)
...
-            filesystem_read_entire(fs, v, heap_backed(kh),
-                                   closure(heap_general(kh),
-                                   kernel_read_complete, fs),
-                                   ignore_status);
-            break;
+    if (bound(klibs_in_bootfs)) {
+        init_klib(bound(kh), fs, boot_root);
+        if (table_find(bound(root), sym(klib_test))) {
+            load_klib("/klib/test", closure(heap_general(bound(kh)), klib_test_loaded));
+        }
```
francescolavra, line 244: How about loading the module from within `init_klib()` instead? This way, all modules to be loaded are loaded from a single place (`init_klib()`), regardless of whether modules are in the boot or root filesystem. Otherwise, in general there is no safe place from which a module can be loaded, because of the asynchronous nature of filesystem mounting.

Commit https://github.com/nanovms/nanos/commit/cacc023798ffb8a70e2df3ffa0a9fb5bbbfb1557 from my (work in progress) https://github.com/nanovms/nanos/tree/feature/telemetry branch has a possible implementation for that.

klib/test.c
```diff
@@ -0,0 +1,24 @@
+#include <klib.h>
+
+#define Z_LEN 1024
+
+int y = 123;
+unsigned long z[Z_LEN];
+
+static int foo(int x)
```
francescolavra, line 8: There seems to be a problem with dynamic relocations, which don't seem to work for global symbols. For example, if I change the `foo()` function above from static to non-static, the load address of the kernel module is not added to the function pointer, and as a result the kernel crashes when it calls foo(). Also calling a non-static foo() from the module itself results in a crash.

src/kernel/stage3.c
```diff
@@ -193,38 +218,51 @@ thunk create_init(kernel_heaps kh, tuple root, filesystem fs)
...
-    table_foreach(c, k, v) {
-        if (!buffer_strcmp(symbol_string(k), "kernel")) {
-            kernel_heaps kh = bound(kh);
-            filesystem_read_entire(fs, v, heap_backed(kh),
-                                   closure(heap_general(kh),
-                                   kernel_read_complete, fs),
-                                   ignore_status);
-            break;
+    if (bound(klibs_in_bootfs)) {
+        init_klib(bound(kh), fs, boot_root);
```
francescolavra, line 241: To complicate matters, if kernel symbols are not loaded, in general a kernel module cannot be loaded, because it wouldn't be able to resolve kernel symbols. So I think init_klib() should only be called after kernel symbols have been loaded. Which means that if the boot process doesn't go through stage2 (e.g. when booting via Frecracker), the kernel binary may need to be loaded even if not requested in the manifest file via `ingest_kernel_symbols`. This applies also if kernel modules are in the root filesystem...
Perhaps it would make sense to have a function in symtab.c that can be called to check whether kernel symbols have been loaded (it could be implemented by checking whether the symbol table covers an address known to belong to the kernel), so that the klib code could call this function and determine whether it's necessary to mount the boot filesystem and/or load the kernel binary.

klib/test.c
francescolavra: Commit https://github.com/nanovms/nanos/commit/1dd6799459b6eaf608ca498dafc40c5b08eae589 has a basic implementation (for x86_64) of dynamic relocations that solves the issue.

klib/Makefile
```diff
@@ -0,0 +1,20 @@
...
+ifeq ($(UNAME_s),Darwin)
+ELF_TARGET=     -target x86_64-elf
+CFLAGS+=        $(ELF_TARGET)
+LD=             x86_64-elf-ld
+OBJDUMP=        x86_64-elf-objdump
+else
+LD=             $(CROSS_COMPILE)ld
+endif
+
+CFLAGS+=	$(KERNCFLAGS) -I$(CURDIR) -fPIC -DKLIB
```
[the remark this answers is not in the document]
francescolavra: We may also need to remove `-ffunction-sections` from KERNCFLAGS and add it here, otherwise kernel functions that are never called inside the kernel will be discarded by the linker and so won't be found by kernel modules. For example, in the TLS module I need to call `direct_connect()`, which, being a new function (introduced in commit https://github.com/nanovms/nanos/commit/f8b445194f697c79d3319c4f94c7103096819f90), is never used in other parts of the kernel.
There are also other functions (such as strstr, calloc, free, rand, snprintf) which I'm now implementing inside the TLS module by could potentially be in the main kernel.

src/kernel/stage3.c
wjhun: I think the klib_get_sym interface is fine, but I suggest that we don't bother with using the elf symtab. We could just as well have a list of functions to export to the klibs that are added to a table in init_klib. This would make for a more efficient lookup, which becomes a hash table lookup on an internalized symbol rather than a linear search through all kernel symbols with string compares. Operation would not be affected if either kernel symbols are not loaded (the scenario you describe above) or the kernel (or klibs) are stripped of symbols. And the -ffunction-sections issue won't be a problem, either, as exported functions will be referenced and thus linked into the kernel binary.

I'll update with an implementation and the other fixes you have provided.

wjhun:
I have updated the branch with the table-based symbol resolution. It also includes the "KLIB_EXPORT" macro for adding symbols using special data sections that get ingested in init_klib. These can be placed in arbitrary places around the kernel, avoiding the need for a centralized array of symbols to export. The other issues @francescolavra mentioned have been addressed here as well. (Note that I wasn't able to reproduce the issue with global symbols not being relocated - changing the test "foo" function to global did not trigger a crash for me - but I still merged the relocation code anyway to cover any such case.)

PR #1322, opened 2020-11-05
Convert IPv4-mapped IPv6 addresses to IPv4 for LwIP

sanderssj:
A socket opened as AF_INET6 but with the any address accepts IPv4 and
IPv6 connections (dual-stack) but the address is always stored in IPv6
format. An IPv4-mapped address in a dual-stack configuration
implies using IPv4 for transport. LwIP doesn't convert for UDP, so this
patch will convert the IPv4-mapped address into an IPv4 address before
passing it along to LwIP so that it will use IPv4 for transport.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -275,6 +275,14 @@ static sysreturn sockaddr_to_addrport(int af, struct sockaddr *addr,
             return -EINVAL;
         struct sockaddr_in6 *sin6 = (struct sockaddr_in6 *)addr;
         sockaddr_to_ip6addr(sin6, ip_addr);
+        /* If this is an an IPv4 mapped address then this socket
+           is dual-stack, so convert the address to IPv4 to encourage
+           LwIP to use that transport
+        */
+        if (ip6_addr_isipv4mappedipv6(ip_2_ip6(ip_addr))) {
+            unmap_ipv4_mapped_ipv6(ip_2_ip4(ip_addr), ip_2_ip6(ip_addr));
+            IP_SET_TYPE_VAL(*ip_addr, IPADDR_TYPE_V4);
+        }
         *port = ntohs(sin6->port);
     }
     return 0;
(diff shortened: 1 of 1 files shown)


eyberg:
confirming this fixes the go issue me - thanks!

wjhun (approved):
This looks good to me. There's just one question: I see it's necessary to convert in socket_write_udp, but should the sockaddr reported to userspace in bind(2) and connect(2) also be converted?

ipv6(7) says "IPv4  connections can be handled with the v6 API by using the v4-mapped-on-v6 address type; thus a program needs to support only this API type to support both protocols. This is handled transparently by the address handling functions in the C library."

...but this seems kind of vague. It might take some digging into glibc to find a definitive answer here.

sanderssj:
> This looks good to me. There's just one question: I see it's necessary to convert in socket_write_udp, but the sockaddr reported to userspace in bind(2) and connect(2) are also being converted. Do they need to be?
> 
> ipv6(7) says "IPv4 connections can be handled with the v6 API by using the v4-mapped-on-v6 address type; thus a program needs to support only this API type to support both protocols. This is handled transparently by the address handling functions in the C library."
> 
> ...but this seems kind of vague. It might take some digging into glibc to find a definitive answer here.

I looked around and thought about this a bit--the call to socket(2) you would specify PF_INET6 for a dual-stack socket, so I think it makes sense that addresses used with this socket would always be IPv6.

I finally found RFC 3493 which says:
>   Applications may use AF_INET6 sockets to open TCP connections to IPv4
>   nodes, or send UDP packets to IPv4 nodes, by simply encoding the
>   destination's IPv4 address as an IPv4-mapped IPv6 address, and
>   passing that address, within a sockaddr_in6 structure, in the
>   connect() or sendto() call.  When applications use AF_INET6 sockets
>   to accept TCP connections from IPv4 nodes, or receive UDP packets
>   from IPv4 nodes, the system returns the peer's address to the
>   application in the accept(), recvfrom(), or getpeername() call using
>   a sockaddr_in6 structure encoded this way.

So I think you're right, the only conversion we need would be in socket_write_udp.

PR #1324, opened 2020-11-06
Unit Test for Bitmap

samantha3pen:
Hey @wjhun, I'm beginning to work on the unit tests for bitmaps, but I wanted to touch base with you before I get too far deep to make sure I'm on the right track with what you're expecting.  

Currently, I'm utilizing id_heap and id_range to create a bitmap. Is this a good way to go about allocating a bitmap to test? I know that this is similar to id_heap_test in that it doesn't exercise the bitmap interface directly, but I'm unsure as to how to create and test on an isolated bitmap.

### DIFF
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -14,7 +14,8 @@ PROGRAMS= \
 	table_test \
 	tuple_test \
 	udp_test \
-	vector_test
+	vector_test \
+	bitmap_test
 SKIP_TEST=	network_test udp_test
 
 SRCS-buffer_test= \
@@ -108,6 +109,11 @@ SRCS-udp_test= \
 
 SRCS-vector_test= \
 	$(CURDIR)/vector_test.c \
+	$(RUNTIME)\
+	$(SRCDIR)/unix_process/unix_process_runtime.c
+
+SRCS-bitmap_test= \
+	$(CURDIR)/bitmap_test.c \
 	$(RUNTIME)\
 	$(SRCDIR)/unix_process/unix_process_runtime.c
 
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -17,27 +17,6 @@ PROGRAMS= \
 	vector_test
 SKIP_TEST=	network_test udp_test
 
-RUNTIME = \
-	$(SRCDIR)/runtime/bitmap.c \
-	$(SRCDIR)/runtime/buffer.c \
-	$(SRCDIR)/runtime/extra_prints.c \
-	$(SRCDIR)/runtime/format.c \
-	$(SRCDIR)/runtime/heap/id.c \
-	$(SRCDIR)/runtime/memops.c \
-	$(SRCDIR)/runtime/merge.c \
-	$(SRCDIR)/runtime/pqueue.c \
-	$(SRCDIR)/runtime/random.c \
-	$(SRCDIR)/runtime/range.c \
-	$(SRCDIR)/runtime/rbtree.c \
-	$(SRCDIR)/runtime/runtime_init.c \
-	$(SRCDIR)/runtime/sg.c \
-	$(SRCDIR)/runtime/symbol.c \
-	$(SRCDIR)/runtime/table.c \
-	$(SRCDIR)/runtime/timer.c \
-	$(SRCDIR)/runtime/tuple.c \
-	$(SRCDIR)/runtime/string.c \
-	$(SRCDIR)/runtime/crypto/chacha.c 
-
 SRCS-buffer_test= \
 	$(CURDIR)/buffer_test.c \
 	$(RUNTIME)\
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -15,7 +15,7 @@ PROGRAMS= \
 	table_test \
 	tuple_test \
 	udp_test \
-	vector_test 
+	vector_test
 SKIP_TEST=	network_test udp_test
 
 SRCS-bitmap_test= \
--- a/test/unit/Makefile
+++ b/test/unit/Makefile
@@ -1,4 +1,5 @@
 PROGRAMS= \
+	bitmap_test \
 	buffer_test \
 	closure_test \
 	id_heap_test \
@@ -14,10 +15,14 @@ PROGRAMS= \
 	table_test \
 	tuple_test \
 	udp_test \
-	vector_test \
-	bitmap_test
+	vector_test 
 SKIP_TEST=	network_test udp_test
 
+SRCS-bitmap_test= \
+	$(CURDIR)/bitmap_test.c \
+	$(RUNTIME)\
+	$(SRCDIR)/unix_process/unix_process_runtime.c
+	
 SRCS-buffer_test= \
 	$(CURDIR)/buffer_test.c \
 	$(RUNTIME)\
@@ -109,11 +114,6 @@ SRCS-udp_test= \
 
 SRCS-vector_test= \
 	$(CURDIR)/vector_test.c \
-	$(RUNTIME)\
-	$(SRCDIR)/unix_process/unix_process_runtime.c
-
-SRCS-bitmap_test= \
-	$(CURDIR)/bitmap_test.c \
 	$(RUNTIME)\
 	$(SRCDIR)/unix_process/unix_process_runtime.c
 
--- a/test/unit/bitmap_test.c
+++ b/test/unit/bitmap_test.c
@@ -49,6 +49,14 @@ u64 test_alloc(id_heap i, id_range r, u64 pages, range subrange) {
     return bit;
 }
 
+/**
+ *  Tests the cloning functionality of bitmaps. 
+ */
+boolean test_clone(bitmap b) {
+    u64 b_cpy = bitmap_clone(b);
+    return (b_cpy == b);
+}
+
 boolean basic_test(heap h)
 {
     // from id_alloc in id.c
@@ -58,6 +66,13 @@ boolean basic_test(heap h)
     u64 b = test_alloc(i, r, pages, WHOLE_RANGE);
     if (bit == INVALID_PHYSICAL)
         //TODO: error check
+
+    // tests bitmap cloning
+    if (!test_clone(b)) {
+        msg_err("cloning bitmap failed.");
+        return false;
+    }
+
 
 }
 
--- a/test/unit/bitmap_test.c
+++ b/test/unit/bitmap_test.c
@@ -13,9 +13,9 @@
 bitmap test_alloc(heap h) {
     bitmap b = allocate_bitmap(h, h, infinity);
     bitmap_foreach_set(b, i) {
-        if (i != 0) { // TODO: figure out appropriate way to check bit is set
+        if (i) {
             msg_err("!!! allocation failed for bitmap\n");
-            return NULL; // TODO: figure out better return val
+            return NULL; 
         }
     }
     return b;
@@ -25,11 +25,15 @@ bitmap test_alloc(heap h) {
  *  Tests cloning of bitmap using bitmap_clone 
  *  function.
  */  
-boolean test_clone(bitmap b) { // TODO: fix per PR comments
+boolean test_clone(bitmap b) { 
+    for (int i = 0; i < 20; i++) 
+        bitmap_set(b, rand(), 1);
     bitmap b_cpy = bitmap_clone(b);
-    if (b_cpy == INVALID_ADDRESS) {
-        msg_err("!!! cloning failed for bitmap\n");
-        return false;
+    bitmap_foreach_set(b, j) {
+        if (bitmap_get(b, j) != bitmap_get(b_cpy, j)) {
+            msg_err("!!! cloning failed for bitmap\n");
+            return false;
+        }
     }
     return true;
 }
@@ -38,12 +42,18 @@ boolean test_clone(bitmap b) { // TODO: fix per PR comments
  *  Tests copying of bitmap using bitmap_copy
  *  function.
  */ 
-boolean test_copy(heap h, bitmap b) { // TODO: fix per PR comments
+boolean test_copy(heap h, bitmap b) { 
     bitmap b_cpy = allocate_bitmap(h, h, infinity);
+    for (int i = 0; i < 20; i++) {
+        bitmap_set(b_cpy, rand(), 1);
+        bitmap_set(b, rand(), 1);
+    }
     bitmap_copy(b, b_cpy);
-    if (b_cpy == NULL) { // TODO
-        msg_err("!!! copying failed for bitmap\n");
-        return false;
+    bitmap_foreach_set(b, j) {
+        if (bitmap_get(b, j) != bitmap_get(b_cpy, j)) {
+            msg_err("!!! copying failed for bitmap\n");
+            return false;
+        }
     }
     return true;
 }
--- a/test/unit/bitmap_test.c
+++ b/test/unit/bitmap_test.c
@@ -149,8 +149,10 @@ boolean test_bitmap_alloc(bitmap b, u64 start, u64 end) {
     while (true) {
         first = bitmap_alloc_within_range(b, nbits, start, end);
         if (first != INVALID_PHYSICAL) {
-            if (nbits > (end-start))
+            if (nbits > (end-start)) {
                 msg_err("!!! invalid value for nbits: %ld end-start: %ld\n", nbits, (end-start));
+                return false;
+            }
             break;
         }
         else
@@ -162,13 +164,13 @@ boolean test_bitmap_alloc(bitmap b, u64 start, u64 end) {
             msg_err("!!! bitmap_alloc failed for bitmap at bit %d | expected: %d actual: %d\n", 
                 i, 1, bitmap_get(b, i));
             if (!bitmap_dealloc(b, first, nbits))
-                msg_err("!!! bitmap_dealloc failed for bitmap");
+                msg_err("!!! bitmap_dealloc failed for bitmap\n");
             return false;
         }
     }
     // check that dealloc is successful
     if (!bitmap_dealloc(b, first, nbits)) {
-        msg_err("!!! bitmap_dealloc failed for bitmap");
+        msg_err("!!! bitmap_dealloc failed for bitmap\n");
         return false;
     }
     return true;
--- a/test/unit/bitmap_test.c
+++ b/test/unit/bitmap_test.c
@@ -144,15 +144,32 @@ boolean test_wrap(heap h) {
  *  bitmap_alloc_within_range and bitmap_dealloc functions.
  */ 
 boolean test_bitmap_alloc(bitmap b, u64 start, u64 end) {
-    // generate random number between 0 and (end-start)
     u64 nbits = rand();
     u64 first;
     while (true) {
         first = bitmap_alloc_within_range(b, nbits, start, end);
-        if (first != INVALID_PHYSICAL)
+        if (first != INVALID_PHYSICAL) {
+            if (nbits > (end-start))
+                msg_err("!!! invalid value for nbits: %ld end-start: %ld\n", nbits, (end-start));
             break;
+        }
         else
             nbits >>= 1;
+    }
+    // check that specified range is set properly
+    for (int i = first; i < first + nbits; i++) {
+        if (bitmap_get(b, i) != 1) {
+            msg_err("!!! bitmap_alloc failed for bitmap at bit %d | expected: %d actual: %d\n", 
+                i, 1, bitmap_get(b, i));
+            if (!bitmap_dealloc(b, first, nbits))
+                msg_err("!!! bitmap_dealloc failed for bitmap");
+            return false;
+        }
+    }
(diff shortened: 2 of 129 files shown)


wjhun:
Hello @samantha3pen. Sorry for the delay. You should be able to instantiate and drive the bitmap code directly, without using the id heap. A call to allocate_bitmap() is all you need, and you can use the heap returned by init_process_runtime() for both the meta and map heap parameters.

I don't see build rules for the test. You can add it to the PROGRAMS list in test/unit/Makefile and add a definition for SRCS-bitmap_test. Then you can build it with 'make -C test/unit bitmap_test'. The resulting binary will be in output/test/unit/bin/bitmap_test, which you can invoke directly in the shell (these tests don't involve nanos; tests that run under nanos are in test/runtime).

I suggest getting the test to build and run using only the most basic tests - maybe just start with allocating and deallocating the bitmap without performing operations on it - and then, once it compiles and runs, incrementally add the different checks that you plan to implement. This may be much easier than trying to write the test all at once before beginning to build and debug it.

test/unit/Makefile
```diff
@@ -14,7 +14,8 @@ PROGRAMS= \
 	table_test \
 	tuple_test \
 	udp_test \
-	vector_test
+	vector_test \
+	bitmap_test
```
francescolavra, line 18: This list is supposed to be in alphabetical order

test/unit/Makefile
```diff
@@ -111,6 +112,11 @@ SRCS-vector_test= \
 	$(RUNTIME)\
 	$(SRCDIR)/unix_process/unix_process_runtime.c
 
+SRCS-bitmap_test= \
+	$(CURDIR)/bitmap_test.c \
+	$(RUNTIME)\
+	$(SRCDIR)/unix_process/unix_process_runtime.c
+
```
francescolavra, line 119: This also should be in alphabetical order, i.e. before `SRCS-buffer_test`

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,80 @@
...
+/**
+ *  Tests allocation of bitmap using allocate_bitmap 
+ *  function.
+ */
+bitmap test_alloc() {
+    heap h = init_process_runtime();
+    bitmap b = allocate_bitmap(h, h, infinity);
+    if (b == INVALID_ADDRESS) 
+        msg_err("!!! allocation failed for bitmap\n");
+    return b;
```
francescolavra, line 18: Here you could test that a just-allocated bitmap doesn't have any bit set, e.g. by using the `bitmap_foreach_set` macro

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,80 @@
...
+ *  Tests cloning of bitmap using bitmap_clone 
+ *  function.
+ */  
+boolean test_clone(bitmap b) {
+    bitmap b_cpy = bitmap_clone(b);
+    if (b_cpy == INVALID_ADDRESS) {
+        msg_err("!!! cloning failed for bitmap\n");
+        return false;
+    }
+    return true;
```
francescolavra, line 31: Here you could test whether the cloning did what's supposed to do, i.e. the cloned bitmap must have the same bits set as the original bitmap. You could e.g. first set and clear some random bits in the original bitmap, then do the cloning, then checks the bits in the newly created bitmap.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ *  Tests copying of bitmap using bitmap_copy
+ *  function.
+ */ 
+boolean test_copy(bitmap b) {
+    heap h = init_process_runtime();
```
francescolavra, line 39: `init_process_runtime()` is supposed to be called only once at initialization, you are calling it both in test_alloc() and here. I suggest calling it outside these functions and passing the returned heap as argument to these functions.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    if (!basic_test()) 
+        goto fail;
+
+    rprintf("tests passing so far\n");
+    msg_debug("test passed\n");
+    exit(EXIT_SUCCESS);
+    fail:
+        msg_err("test failed\n");
+        exit(EXIT_FAILURE);
```
francescolavra, line 76: Indentation after the label should be the same as before the label

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,80 @@
...
+ */ 
+boolean test_copy(bitmap b) {
+    heap h = init_process_runtime();
+    bitmap b_cpy = allocate_bitmap(h, h, infinity);
+    bitmap_copy(b, b_cpy);
+    if (b_cpy == NULL) { // TODO
+        msg_err("!!! copying failed for bitmap\n");
+        return false;
+    }
+    return true;
```
francescolavra, line 46: Also here testing that the new bitmap has the same bits set as the original one would be useful.

test/unit/bitmap_test.c
samantha3pen: @francescolavra I looked at the bitmap_foreach_set macro and its application in poll.c, but I'm still kind of lost on what the second parameter is used for and how I could use it to check each bit. Could you explain its functionality or direct me to somewhere in the codebase that can explain it better?

francescolavra: The second parameter of the bitmap_foreach_set macro is just a name or your choice. This macro expands to a `for` loop that has as many iterations as there are bits set in the bitmap, and at each iteration the variable whose name you chose as second parameter will contain the value of a bit set in the bitmap. So you could use this macro to verify that a bitmap doesn't have any bit set, as below:
```
bitmap_foreach_set(b, i) {
    test failed!!! return error code
}
```

samantha3pen:
Hi @francescolavra I have a few more questions about the bitmap functions so I can better test them:

1. I use bitmap_foreach_set while testing other functions, but what would be the best way to test this and the other bitmap_foreach functions in isolation? 
2. Could you explain what bitmap_un/wrap achieves? I see that they use un/wrap_buffer under the hood, but I'm not sure I understand it correctly.
3. What does bitmap_de/alloc do? And in a similar vein, how does for_range_in_map contribute to its functionality?

francescolavra (commented):
@samantha3pen, here are my answers to your questions.

1. You could test implicitly bitmap_foreach_set() as I mentioned in my comments to the test_clone() function; I wouldn't worry about testing the other foreach macros in isolation
2. bitmap_wrap() allocates a bitmap that uses an already allocated memory area, so for example if this memory area is the location of a 64-bit variable, then all bits set in the variable value should also be set in the bitmap that wraps it. You could test bitmap_wrap() by assigning a random value to a 64-bit variable and then checking the correspondence of bits in the variable with bits in the bitmap. bitmap_unwrap() simply deallocates the internal data of the bitmap, not much to test here, just call the function after the bitmap_wrap() test.
3. bitmap_alloc() searches a bitmap for a range of `nbits` consecutive bits that are all cleared, and if such a range is found, sets all bits in the range and returns the first bit of the range; if not found, returns INVALID_PHYSICAL. bitmap_dealloc() checks that the range of `size` bits starting from `bit` is set in the bitmap; if this check is successful, the function clears all the bits in that range and returns true, otherwise it returns false.

for_range_in_map() is a helper function that:
  - if `set` is true, sets (if `val` is true) or clears (if `val` is false) all bits in the range [start, start + nbits)
  - if `set` is false, checks and returns whether all bits in the range [start, start + nbits) are set (if `val` is true) or cleared (if `val` is false)

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,115 @@
...
+    for (int i = 0; i < 20; i++) 
+        bitmap_set(b, rand(), 1);
+    bitmap b_cpy = bitmap_clone(b);
+    bitmap_foreach_set(b, j) {
+        if (bitmap_get(b, j) != bitmap_get(b_cpy, j)) {
+            msg_err("!!! cloning failed for bitmap\n");
+            return false;
+        }
+    }
+    return true;
```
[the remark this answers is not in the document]
samantha3pen: I updated the bit checking in both clone and copy in my latest commit. Does my implementation test both set and cleared bits?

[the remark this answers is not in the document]
francescolavra: No, it doesn't test cleared bits, because the bitmap_foreach_set loop iterates on set bits only. Also the check using bitmap_base() is unnecessarily complicated, I would replace it with just `if (!bitmap_get(b_cpy, j))`.
To test cleared bits, you could add a new bitmap_foreach_set loop that iterates on `b_cpy` instead of `b`.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ *  Tests wrapping and unwrapping of bitmap using 
+ *  bitmap_wrap and bitmap_unwrap functions.
+ */ 
+boolean test_wrap(heap h) {
+    u64 map = rand();
+    bitmap b = bitmap_wrap(h, &map, infinity);
```
francescolavra, line 99: The third argument to bitmap_wrap() is the map length in bits, so it should be 64 instead of infinity.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,162 @@
...
+    if (b->alloc_map->contents != &map) {
```
francescolavra, line 100: This doesn't test the correspondence between bits in `map` and bits in the bitmap `b`.
One way to test it would be:
```
for (int i = 0; i < 64; i++)
    if (((map & (1 << i)) && !bitmap_get(b, i)) || (!(map & (1 << i)) && bitmap_get(b, i)))
        test failed
```

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,162 @@
...
+        return false;
+    }
+    bitmap_unwrap(b);
+    return true;
+}
+
+boolean test_bitmap_alloc(bitmap b, u64 start, u64 end) {
+    u64 nbits = rand();
+    u64 first = bitmap_alloc_within_range(b, nbits, start, end);
+    if (first != INVALID_PHYSICAL && !bitmap_dealloc(b, first, nbits)) {
```
francescolavra, line 112: `nbits` should not be larger than (end - start), otherwise bitmap_alloc_within_range() fails and then bitmap_dealloc() is not tested.
So if `nbits` has an appropriate value, the above `if` can be replaced by `if ((first == INVALID_PHYSICAL) || !bitmap_dealloc(b, first, nbits))`

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,162 @@
...
+    if (!test_copy(h, b)) return false;
+
+    // tests bitmap set then get
+    if (!test_set_and_get(b)) return false;
+
+    // tests bitmap wrap then unwrap
+    if (!test_wrap(h)) return false;
+
+    // tests bitmap alloc then bitmap alloc within range
+    if (!test_bitmap_alloc(b, 0, infinity)) return false;
```
francescolavra, line 139: Using `infinity` as end value is not good, because there are limitations on how big a bitmap can be. I would put a maximum cap such as 4096, and use this as end value here.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,162 @@
...
+    u64 start = rand();
+    u64 end = rand() % (infinity + 1 - start) + start;
+    if(!test_bitmap_alloc(b, start, end)) return false;
```
francescolavra, line 142: Also here make sure that end is not larger than 4096.

test/unit/bitmap_test.c
samantha3pen: Per my latest commit, I realize that rand() returns a 32bit int. Do you know how I can generate a random u64 to use for testing the wrap function?

test/unit/bitmap_test.c
samantha3pen: My test for bitmap_alloc within a specified range (line 187) is failing because bitmap_alloc_internal fails the first condition check on line 102 (bitmap.c). 

I attempted to fix this issue by recalculating nbits until it satisfies a condition that wouldn't fail bitmap_alloc (lines 145-153 in bitmap_test.c), but this isn't an efficient solution.

Do you have any suggestions of how I can ensure a combination of _start, end,_ and _nbits_ will lead to a successful bitmap_alloc?

samantha3pen:
I have a couple more questions before I begin writing the tests for the remaining bitmap functions:

1. In _bitmap_range_check_and_set_, what does _validate_ represent?
2. Should I still test _bitmap_base_ if it simply uses _buffer_ref_ under the hood?

test/unit/bitmap_test.c
francescolavra: You could call rand() to generate the first 32 bits, and then shift the result and call rand() again to generate the other 32 bits, as in:
```
u64 map = rand();
map = (map << 32) | rand();
```

test/unit/bitmap_test.c
francescolavra: OK, I see now that a bit range is always allocated starting from a bit that is multiple of the range length, so as you noted having `nbits <= (end - start)` doesn't guarantee a successful allocation.
I wouldn't make the test code dependent on this internal implementation detail of the bitmap code (which could potentially change in the future), so I would just start with `nbits = rand()` and then call bitmap_alloc_within_range() until it succeeds, halving the `nbits` value after each failed attempt, as in:
```
while (true) {
    first = bitmap_alloc_within_range(b, nbits, start, end);
    if (first != INVALID_PHYSICAL)
        break;
    else
        nbits >>= 1;
}
```

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/**
+ *  Tests wrapping and unwrapping of bitmap using 
+ *  bitmap_wrap and bitmap_unwrap functions.
+ */ 
+boolean test_wrap(heap h) { 
+    u64 map = rand(); // how to generate random 64 bit 
+    bitmap b = bitmap_wrap(h, &map, 32);
+    for (int i = 0; i < 32; i++) {
```
francescolavra, line 128: The third argument in the bitmap_wrap() call should be 64, and also the `for` loop should go up to 64.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,207 @@
...
+
+    // tests bitmap set then get
+    if (!test_set_and_get(b)) return false;
+
+    // tests bitmap wrap then unwrap
+    if (!test_wrap(h)) return false;
+
+    // tests bitmap alloc then bitmap alloc within range
+    if (!test_bitmap_alloc(b, 0, 4096)) return false;
+    u64 start = rand() % (4096 + 1);
```
francescolavra, line 185: It seems to me that it should be 4096 instead of (4096 + 1)

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,207 @@
...
+    u64 end = rand() % (4096 + 1 - start) + start;
```
francescolavra, line 186: Same here

francescolavra (commented):
To answer your questions:
1. if `validate`is true, bitmap_range_check_and_set() checks whether all bits in the supplied bit range are cleared in the bitmap, before attempting to set them; otherwise, this check is skipped and the bits are set regardless of whether they were already set
2. No, I wouldn't test bitmap_base()

test/unit/bitmap_test.c
samantha3pen: Actually, I realized that I was incorrectly calculating `nbits` in the first place. I was generating a random number between `start `and `end` for `nbits` instead of a random number between 0 and (end-start).

test/unit/bitmap_test.c
samantha3pen: The wrap test seems to consistently fail when checking the 31st bit. In the `if` statement on line 130, the value of `(map & (1 << i))` in the first condition is 0, but the value of `!(map & (1 << i))` in the second condition is also 0. 
Do you know why this could be?

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/**
+ *  Tests bitmap range check and set of bitmap using
+ *  bitmap_range_check_and_set
+ */
+boolean test_range_check_set(bitmap b) {
+    u64 start = rand() % 4096;
+    u64 nbits = rand() % (4097 - start);
+    boolean validate = rand() & 1;
```
francescolavra, line 162: Since the bitmap at this point may have some random bits set (because it has been used in the functions called prior to `(test_range_check_set)`), you cannot be sure that the validation step in bitmap_range_check_and_set() will pass. So depending on what random values are generated, this test may pass or fail. To avoid this possibility, I would just set `validate` to false.

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,222 @@
...
+    return true;
+}
+
+/**
+ *  Tests bitmap range check and set of bitmap using
+ *  bitmap_range_check_and_set
+ */
+boolean test_range_check_set(bitmap b) {
+    u64 start = rand() % 4096;
+    u64 nbits = rand() % (4097 - start);
```
francescolavra, line 161: It seems to me that 4097 should be 4096 instead.

test/unit/bitmap_test.c
francescolavra: > The wrap test seems to consistently fail when checking the 31st bit. In the `if` statement on line 130, the value of `(map & (1 << i))` in the first condition is 0, but the value of `!(map & (1 << i))` in the second condition is also 0.
> Do you know why this could be?

It's because (1 << i) is a 32-bit integer, which is sign-extended to 64 bits when ANDed with `map`. To avoid this problem, you should use (1ULL << i) instead of (1 << i) (in three places, including the `msg_err` call). And in the `msg_err` call there should be "map: %ld" instead of "map: %d"

test/unit/bitmap_test.c
francescolavra: > Actually, I realized that I was incorrectly calculating `nbits` in the first place. I was generating a random number between `start `and `end` for `nbits` instead of a random number between 0 and (end-start).

Well, the fact that this test is passing now doesn't guarantee that it will pass with other random values. And if you try calling test_bitmap_alloc() many times with different random values, one of these calls will end up failing, because of the fact that a bit range is always allocated starting from a bit that is multiple of the range length. The test should be robust and not fail, regardless of how many times it's executed.

samantha3pen: I see, thanks for clarifying that. 
In that case, if we loop until `nbits` satisfies `bitmap_alloc_within_range`, there wouldn't be an opportunity for this test to fail, right? So there isn't any error checking necessary when testing this function?

samantha3pen:
I believe so far I've written basic unit tests for all the relevant bitmap functions, but what else can I add to this file to make these tests robust?

test/unit/bitmap_test.c
francescolavra: There is some more error checking that can be added, such as:
- if (first != INVALID_PHYSICAL), checks that nbits is <= (end - start)
- check that all bits in the allocated range are actually set (similarly to what is done in `test_range_check_set()`)
- at the end, call bitmap_dealloc() and check that it returns true

francescolavra:
> I believe so far I've written basic unit tests for all the relevant bitmap functions. Is there anything else I can add to this file to make these tests robust?

There are many different use cases for various bitmap functions that could potentially be tested, but I think the basic things are covered now (apart from what I wrote in the other comment).

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,244 @@
...
+ */ 
+boolean test_bitmap_alloc(bitmap b, u64 start, u64 end) {
+    u64 nbits = rand();
+    u64 first;
+    while (true) {
+        first = bitmap_alloc_within_range(b, nbits, start, end);
+        if (first != INVALID_PHYSICAL) {
+            if (nbits > (end-start))
+                msg_err("!!! invalid value for nbits: %ld end-start: %ld\n", nbits, (end-start));
+            break;
```
francescolavra, line 154: If (nbits > end - start), this function should return false, so that the test fails

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,244 @@
...
+        else
+            nbits >>= 1;
+    }
+    // check that specified range is set properly
+    for (int i = first; i < first + nbits; i++) {
+        if (bitmap_get(b, i) != 1) {
+            msg_err("!!! bitmap_alloc failed for bitmap at bit %d | expected: %d actual: %d\n", 
+                i, 1, bitmap_get(b, i));
+            if (!bitmap_dealloc(b, first, nbits))
+                msg_err("!!! bitmap_dealloc failed for bitmap");
```
francescolavra, line 165: There should be `\n` at the end of the string

test/unit/bitmap_test.c
```diff
@@ -0,0 +1,244 @@
...
+            return false;
+        }
+    }
+    // check that dealloc is successful
+    if (!bitmap_dealloc(b, first, nbits)) {
+        msg_err("!!! bitmap_dealloc failed for bitmap");
```
francescolavra, line 171: Ditto

PR #1326, opened 2020-11-12
Separate APIC id from cpu id and make sigstate SMP safe

sanderssj:
This PR contains fixes for two issues I found while making ftrace SMP safe. The first issue was that there was concurrent access to the sigstate lists and pending field that would cause a crash. That just needed a spinlock to protect those fields. I did not protect the mask field because I'm not convinced it's needed, but could use a second opinion. 

The second issue is that on ESXi the APIC ID for additional cpus is incremented by two and not by one, so the cpuinfo array would have holes in it since the APIC ID is used as cpu id. ESXi uses an x2APIC unlike QEMU, which may be the explanation. In any case, I made the two IDs independent: cpu id is now generated in ap_start with the new stack, and a separate array is used to map cpu id to APIC ID. IPIs are called with cpu id which get converted to the proper APIC ID in apic.c.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -377,7 +377,6 @@ u64 total_processors = 1;
 #ifdef SMP_ENABLE
 static void new_cpu()
 {
-    fetch_and_add(&total_processors, 1);
     if (platform_timer_percpu_init)
         apply(platform_timer_percpu_init);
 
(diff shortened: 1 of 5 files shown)


francescolavra (commented):
Looks good to me.

wjhun (approved):
Looks good

wjhun:
> x2APIC unlike QEMU, which may be the explanation.

Meant to ask - what is your qemu environment? x2APIC is supported and used on linux/kvm (on my setup, anyway), but not with -noaccel or macos. But that sparse id thing isn't a problem on qemu using x2APIC.

sanderssj:
This was on ESXi 6.7, so that's why it's weird!

PR #1327, opened 2020-11-12
ftrace improvements for analysis and smp

sanderssj:
This branch contains a number of improvements to ftrace capability, but ftrace is still a work in progress. I made this PR so that there would be a working ftrace option in master.
The changes are:
* move stack traces to cpuinfo from thread because thread identification was not reliable
* add always_inline attribute to functions so that an ftrace build works properly with -fno-inline (or lower optimization level)
* make ftrace SMP-safe by implementing locks for rbufs
* added additional analysis scripts


src/unix/thread.c
```diff
@@ -203,6 +203,8 @@ void thread_sleep_interruptible(void)
     disable_interrupts();
     assert(current->blocked_on);
     thread_log(current, "sleep interruptible (on \"%s\")", blockq_name(current->blocked_on));
+    ftrace_thread_switch(current, 0);
+    set_current_thread(0);
```
francescolavra, line 207: `set_current_thread(0)` shouldn't be necessary, since it's already being done when entering the runloop (by calling sched_thread_pause() which in turn calls thread_pause())

src/unix/thread.c
```diff
@@ -213,6 +215,8 @@ void thread_sleep_uninterruptible(void)
     assert(!current->blocked_on);
     current->blocked_on = INVALID_ADDRESS;
     thread_log(current, "sleep uninterruptible");
+    ftrace_thread_switch(current, 0);
+    set_current_thread(0);
```
francescolavra, line 219: Ditto

src/unix/thread.c
sanderssj: Thanks for the catch!

src/unix/ftrace.c
```diff
@@ -3,8 +3,14 @@
...
 
+#define FTRACE_DEBUG
+#ifdef FTRACE_DEBUG
+#define ft_debug(x, ...) do {rprintf("FTRACE: " x, ##__VA_ARGS__);} while(0)
+#else
+#define ft_debug(x, ...)
+#endif
 /* 64MB default size for the user's trace array */
-#define DEFAULT_TRACE_ARRAY_SIZE        (64ULL << 20)
+#define DEFAULT_TRACE_ARRAY_SIZE        (512ULL << 20)
```
wjhun, line 13: Maybe this is one for config.h?

src/unix/ftrace.c
```diff
@@ -1807,48 +1832,59 @@ ftrace_init(unix_heaps uh, filesystem fs)
...
+    for (int i = 0; i < MAX_CPUS; i++) {
+        cpuinfo ci = cpuinfo_from_id(i);
+        ftrace_cpu_deinit(ci);
+    }
     deallocate_http_listener(ftrace_heap, ftrace_hl);
 }
 
 int
-ftrace_thread_init(thread t)
+ftrace_cpu_init(cpuinfo ci)
```
wjhun, line 1854: This isn't related to this PR, but it would be nice if we could standardize per-cpu inits so that they can just be called altogether as CPUs come online. There's a timer-specific cpu init called in new_cpu(), but it could just become a list. (filed https://github.com/nanovms/nanos/issues/1328)

src/unix/ftrace.c
```diff
@@ -737,6 +762,7 @@ function_graph_toggle(boolean enable)
 static void
 print_switch_event(struct ftrace_printer * p, struct rbuf_entry_switch * sw)
 {
+    return;
```
wjhun, line 765: debug cruft?

src/unix/ftrace.h
```diff
@@ -96,7 +96,7 @@ ftrace_deinit(void)
 {}
```
wjhun, line 96: ftrace_deinit stub should be removed

wjhun (approved):
Just minor notes - looks good!

src/unix/ftrace.c
sanderssj: Sort of--thread switches aren't correct yet, so it screws up analysis tools with incomplete or misplaced messages. This is to be fixed later with the thread switches. I just changed it to be ifdef'd out and added a comment to explain what's going on so it's not so hacky.

Issue #1318, opened 2020-11-02
Stopping GCP instances takes too long time


fabioDMFerreira:
I think we need to handle the stop signal gracefully. After stopping an instance it seems the operation gets stuck. Probably related with #1065.

francescolavra:
@fabioDMFerreira yes, we need to catch and handle the ACPI poweroff signal that is sent by the host to the VM when a GCP instance is stopped. We are tracking this issue in #1065.

fabioDMFerreira:
thanks @francescolavra 

I'll wait for this to solve https://github.com/nanovms/ops/issues/684.

fabioDMFerreira:
@francescolavra there's a regression. Stopping instances in GCP is taking too long again.

Wondering if the reboot on exit feature has some influence.

francescolavra:
@fabioDMFerreira yes, the reboot on exit feature influences the behavior of an instance being stopped.
When this feature is not enabled, stopping an instance takes the usual time (I just measured 20 seconds for `instance stop` and 22 seconds for `instance delete`, in line with the time required to stop or delete a Linux instance via the GCP portal); when this feature is enabled, trying to stop an instance will actually make it reboot, which means that it can only be stopped forcefully by GCP (which it does after a 90 second timeout).

PR #1331, opened 2020-11-17
ACPI: add detection of power button event (x86_64 only)

### LINKED ISSUE #1065: graceful shutdowns on gcp (ACPI Power Off signal handling)
VMs on GCP are killed forcefully after timing out when being stopped or deleted. I have not tried on AWS.

According to [the docs](https://cloud.google.com/compute/docs/instances/stop-start-instance), an ACPI Power Off signal is sent to the guest os.

I tried this using ops with the latest release of nanos as well as the latest commit https://github.com/nanovms/nanos/commit/11be3e189560ed1b881f20adb016377126aa5897.

EDIT: Necessarily, the VM also does not shutdown after the program exits.

francescolavra:
This PR implements detection of the ACPI power button event (which is generated when a Google cloud instance is being stopped) via the PIIX4 ACPI-compatible power management interface.
Upon detecting this event, the kernel shuts down, and when the shutdown is completed the kernel puts the system in the ACPI Soft Off state (which notifies the hypervisor that the powerdown event has been acted upon).
Note: on AWS, even though the PIIX4 ACPI interface is emulated, an instance stop command does not generate an ACPI power button event. Instead, it writes the "poweroff" string in the "control/shutdown" domU XenStore node. Thus, to implement graceful shutdown in AWS, the Xen driver needs to support asynchronous notifications on the XenStore event channel.

Closes #1065, #1318.

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -80,6 +80,7 @@ SRCS-kernel.img= \
 	$(SRCDIR)/hyperv/storvsc/storvsc.c \
 	$(SRCDIR)/hyperv/utilities/vmbus_ic.c \
 	$(SRCDIR)/hyperv/utilities/vmbus_shutdown.c \
+	$(SRCDIR)/x86_64/acpi.c \
 	$(SRCDIR)/x86_64/ap.s \
 	$(SRCDIR)/x86_64/apic.c \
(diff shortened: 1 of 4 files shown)


sanderssj (commented):
Looks good to me!

wjhun (approved):
Looks good

Issue #1306, opened 2020-10-05
add futex runtime test


wjhun:
We don't currently have an explicit test for futexes. We could use an exhaustive test that stresses both the futex interface as well as correctness on single and multiple-core instances. The test should run under both a Linux host and nanos as a parity check.

uma-sethuraman:
Hello, my team members and I are students at the University of Texas at Austin. We are interested in contributing to Nanos as a part of a project for our virtualization class. Could we take on this issue?

wjhun:
> Hello, my team members and I are students at the University of Texas at Austin. We are interested in contributing to Nanos as a part of a project for our virtualization class. Could we take on this issue?

Sure. I have just assigned it to you. Thanks!

wjhun:
Note that the Linux Test Project has several futex tests. They rely on fork() to run, and thus won't work on nanos without modification. (Also note that simply changing "forks_child" in the tst_test structs won't work, either.) You could see whether it is possible to modify these tests to use pthreads instead of spawning child processes.

Please note, however, that LTP is GPL code - and thus we cannot import any of it into the nanos tree. While still need a futex runtime test of our own for catching regressions, adapting LTP for a single run of tests may help isolate futex bugs in the near term, which would still be very helpful.

See https://github.com/nanovms/nanos/wiki/Running-tests-from-Linux-Test-Project

uma-sethuraman:
Hello @wjhun, I have started working on this issue and familiarizing myself with how the futex tests work in the Linux Test Project. I was wondering if you could give me some more guidance about how I can adapt the LTP futex tests for nanos and what your expectations are.

PR #1335, opened 2020-11-22
Futex Runtime Tests

uma-sethuraman:
Hello @wjhun, I have added futex runtime tests which use pthreads for futex_wake, futex_wait, futex_wait_bitset, which are the methods listed here: https://github.com/nanovms/nanos/wiki/Running-tests-from-Linux-Test-Project

This pull request is for issue #1306. Please let me know how the tests look and if I should change or add any tests.

### DIFF
--- a/test/runtime/Makefile
+++ b/test/runtime/Makefile
@@ -43,8 +43,9 @@ PROGRAMS= \
 
 SRCS-futex_test= \
 	$(CURDIR)/futex_test.c \
-	$(SRCDIR)/unix_process/ssp.c
+	$(SRCDIR)/unix_process/ssp.c \
 LDFLAGS-futex_test=	-static
+LIBS-futex_test=	-lpthread
 
 SRCS-aio= \
 	$(CURDIR)/aio.c \
--- a/test/runtime/Makefile
+++ b/test/runtime/Makefile
@@ -14,6 +14,7 @@ PROGRAMS= \
 	getrandom \
 	hw \
 	hws \
+	klibs \
 	io_uring \
 	mkdir \
 	mmap \
@@ -105,6 +106,9 @@ SRCS-io_uring= \
 	$(CURDIR)/io_uring.c \
 	$(SRCDIR)/unix_process/ssp.c
 LDFLAGS-io_uring=	-static
+
+SRCS-klibs=		$(CURDIR)/klibs.c
+LDFLAGS-klibs=		-static
 
 SRCS-mmap= \
 	$(CURDIR)/mmap.c \
--- a/test/runtime/Makefile
+++ b/test/runtime/Makefile
@@ -1,5 +1,6 @@
 # these are built for the target platform (Linux x86_64)
 PROGRAMS= \
+	futex_test \
 	aio \
 	dup \
 	creat \
@@ -39,6 +40,11 @@ PROGRAMS= \
 	webs \
 	write \
 	writev
+
+SRCS-futex_test= \
+	$(CURDIR)/futex_test.c \
+	$(SRCDIR)/unix_process/ssp.c
+LDFLAGS-futex_test=	-static
 
 SRCS-aio= \
 	$(CURDIR)/aio.c \
--- a/test/runtime/Makefile
+++ b/test/runtime/Makefile
@@ -1,6 +1,5 @@
 # these are built for the target platform (Linux x86_64)
 PROGRAMS= \
-	futex_test \
 	aio \
 	dup \
 	creat \
@@ -11,6 +10,7 @@ PROGRAMS= \
 	fcntl \
 	fst \
 	ftrace \
+	futex \
 	getdents \
 	getrandom \
 	hw \
@@ -40,12 +40,6 @@ PROGRAMS= \
 	webs \
 	write \
 	writev
-
-SRCS-futex_test= \
-	$(CURDIR)/futex_test.c \
-	$(SRCDIR)/unix_process/ssp.c \
-LDFLAGS-futex_test=	-static
-LIBS-futex_test=	-lpthread
 
 SRCS-aio= \
 	$(CURDIR)/aio.c \
@@ -96,6 +90,12 @@ SRCS-ftrace= \
 	$(SRCDIR)/unix_process/ssp.c
 
 LDFLAGS-ftrace=	-static
+
+SRCS-futex= \
+	$(CURDIR)/futex.c \
+	$(SRCDIR)/unix_process/ssp.c
+LDFLAGS-futex=	-static
+LIBS-futex=	-lpthread
 
 SRCS-getdents=		$(CURDIR)/getdents.c
 LDFLAGS-getdents=	-static
--- a/test/runtime/futex.c
+++ b/test/runtime/futex.c
@@ -4,12 +4,10 @@
 #include <stdlib.h>
 #include <pthread.h>
 #include <runtime.h>
+#include <linux/futex.h>
 
 #define EXIT_FAILURE 1
 #define EXIT_SUCCESS 0
-#define FUTEX_WAIT 0
-#define FUTEX_WAKE 1
-#define FUTEX_WAIT_BITSET 9
 #define FUTEX_INITIALIZER 0
 
 /* Test Global Variables */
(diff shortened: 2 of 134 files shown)


wjhun:
Thank you @uma-sethuraman! We will take a look at the changes.

test/runtime/Makefile
```diff
@@ -40,6 +41,12 @@ PROGRAMS= \
 	write \
 	writev
 
+SRCS-futex_test= \
+	$(CURDIR)/futex_test.c \
+	$(SRCDIR)/unix_process/ssp.c \
+LDFLAGS-futex_test=	-static
```
francescolavra, line 47: The `\` character after ssp.c shouldn't be there, because it causes the LDFLAGS line to not be parsed correctly, which means that this program doesn't run in Nanos with the provided manifest file.
To run this program in Nanos, from the top-level folder of the repo execute `make run TARGET=futex_test`

Also, this should be in alphabetical order

test/runtime/futex_test.c
```diff
@@ -0,0 +1,149 @@
...
+        if (pthread_create(&(threads[index]), NULL, futex_wake_test_thread, (void*)(uaddr))) {
+            printf("Unable to create thread\n");
+            return false;
+        }
+    }
+    
+    sleep(1); /* for main thread */
+    int ret = syscall(SYS_futex, uaddr, FUTEX_WAKE, num_to_wake, 0, NULL, 0);
+    if (ret == expected_result)
+        return true; /* success */
```
francescolavra, line 39: Before returning, should `pthread_join()` all the threads, which would test that indeed all the threads have been woken up (or never went to sleep)

francescolavra (changes requested):
Hi @uma-sethuraman, these tests should exercise also other futex features that are implemented in the kernel (the futex implementation in the kernel can be found in src/unix/futex.c):
- FUTEX_CMP_REQUEUE
- FUTEX_WAKE_OP
- timeouts (for both FUTEX_WAIT and FUTEX_WAIT_BITSET)

Also, could you add this test program to RUNTIME_TESTS in the top-level Makefile (so that it is run as part of the CI tests)?

test/runtime/Makefile
```diff
@@ -1,5 +1,6 @@
 # these are built for the target platform (Linux x86_64)
 PROGRAMS= \
+	futex_test \
```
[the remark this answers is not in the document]
uma-sethuraman: I have fixed this problem and have updated the source file to be called futex.

test/runtime/futex_test.c
```diff
@@ -0,0 +1,149 @@
...
+#include <sys/syscall.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <runtime.h>
+
+#define EXIT_FAILURE 1
+#define EXIT_SUCCESS 0
+#define FUTEX_WAIT 0
+#define FUTEX_WAKE 1
+#define FUTEX_WAIT_BITSET 9
```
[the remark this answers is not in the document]
uma-sethuraman: I have fixed this so that the constants are not redefined.

test/runtime/futex_test.c
```diff
@@ -0,0 +1,149 @@
...
+#define FUTEX_INITIALIZER 0
+
+/* Test Global Variables */
+int wake_test_futex;
+int empty_futex;
+int wait_test_futex = FUTEX_INITIALIZER;
+int wait_bitset_test_futex = FUTEX_INITIALIZER;
+
+/* Helper Function Declarations */
+static void * futex_wake_test_thread(void *arg);
```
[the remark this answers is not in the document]
uma-sethuraman: I have updated the `*` style throughout the file to make sure there is no space after `*`.

test/runtime/futex_test.c
```diff
@@ -0,0 +1,149 @@
...
+
+/* FUTEX_WAKE test, creates num_to_wake threads which wait
+on uaddr and then wakes up all the threads */
+static boolean futex_wake_test(int* uaddr, int num_to_wake, int expected_result) {
```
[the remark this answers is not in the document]
uma-sethuraman: I have fixed this and adjusted all the opening braces.

test/runtime/Makefile
uma-sethuraman: Hello @francescolavra, I removed the `\ `character on this line. However, when I run `make run TARGET=futex`, I get the following error: 
```
Could not access KVM kernel module: No such file or directory
qemu-system-x86_64: failed to initialize KVM: No such file or directory
```

I think this may be because the` futex.manifest` file I created in `test/runtime` is incorrect.
When I run `make -C test/runtime futex` from the top-level folder of the repo, my test compiles and I do `./output/test/runtime/bin/futex` to run the test. Do you know how to fix the issue with `make run TARGET=futex`?

test/runtime/futex_test.c
uma-sethuraman: Hello @francescolavra, does this mean to iterate over all the thread IDs and `pthread_join()` them? And then only return true if all the `pthread_join` calls are successful?

uma-sethuraman: Hello @francescolavra, to clarify, after iterating over all the threads and calling `pthread_join` on them, I should only return true if all the `pthread_join` calls are successful?

uma-sethuraman:
Hello @francescolavra, thank you for your feedback! I will add in tests for those other futex features which you mentioned. I had questions about two of the changes you had requested in the code, and I have asked those questions above under those specific changes.

test/runtime/futex_test.c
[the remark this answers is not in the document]
francescolavra: I should have been more clear, the opening braces should go to the next line only for function definitions (and not in other instances), this is the coding style used throughout the code base

test/runtime/Makefile
francescolavra: It seems your machine doesn't have KVM properly set up. You can try loading the KVM kernel module manually (with `sudo modprobe kvm-intel`) and see if that fixes the problem. Or alternatively run the futex test on Nanos without KVM acceleration: the command is `make run-noaccel TARGET=futex`

test/runtime/futex_test.c
francescolavra: Yes, only return true if (ret == expected_result) and all the pthread_join calls are successful

test/runtime/Makefile
uma-sethuraman: Okay, I will try that. Thank you!

uma-sethuraman:
Hello @francescolavra, I have fixed what you mentioned earlier and I added tests for FUTEX_CMP_REQUEUE,
FUTEX_WAKE_OP, and timeout tests for FUTEX_WAIT and FUTEX_WAIT_BITSET.

uma-sethuraman:
Hello @francescolavra, can you let me know if all my updates look good or if there's anything left to do for this pull request?

francescolavra (commented):
Hi @uma-sethuraman, I have a few comments on the newly added code

uma-sethuraman:
Hi @francescolavra, I updated the timeout checks based on your previous comments. Now each timeout is only 0.3 seconds, and instead of checking the elapsed time, I check if `errno == ETIMEDOUT`.

Is there anything else I should change?

test/runtime/futex.c
```diff
@@ -0,0 +1,401 @@
...
+static boolean futex_wait_bitset_test_2() 
+{
+    int *uaddr = (int*)(&wait_bitset_test_futex);
+    int val = FUTEX_INITIALIZER; /* value at uaddr */
+    int bitset = 0xffffffff;
+    int expected_result = -1; 
+
+    /* timeout of 0.3 seconds */
+    struct timespec timeout = {.tv_sec = 0, .tv_nsec = 300000000};
+    int ret = syscall(SYS_futex, uaddr, FUTEX_WAIT_BITSET, val, &timeout, NULL, bitset);
```
francescolavra, line 144: When FUTEX_WAIT_BITSET is given a timeout, this timeout is interpreted (unlike in the FUTEX_WAIT case) as an absolute time. So here the timeout value should be at an offset (e.g. 0.3 seconds) from the time retrieved via clock_gettime(CLOCK_MONOTONIC), as you were doing before.

uma-sethuraman: @francescolavra, that makes sense. I updated that test to make the timeout value an offset, as I had in a previous version of the code. Does it look correct now?

francescolavra: The value in the tv_nsec field of struct timespec should always be in the range 0 to 999999999, and the current code doesn't ensure this.

uma-sethuraman: @francescolavra I've now added a check to make sure the timespec tv_nsec field is always between [0, 999999999]. Does it look correct now?

francescolavra: The result of the division between integer variables is always an integer, so there is no need to call `floor()`, and the math.h header file inclusion can be avoided as well.

uma-sethuraman: @francescolavra I have fixed this so that floor is not called, and I removed the math.h in the header. Do all the futex tests look good now or is there anything I should update?

PR #1337, opened 2020-11-23
locking heap wrapper

wjhun:
As enumerated in #1302, there are many paths in the kernel where the kernel lock is not held and yet allocations or deallocations are being made from one of the kernel heaps. To address this issue, these changes introduce a "locked" kernel heap, using a wrapper heap which wraps heap methods with spinlock aquires, which helps to make such paths safe.

As the interface for the id heap is fairly complex and wouldn't be covered by a generic heap wrapper, a "locking" flag is specified on ID heap creation. The virtual huge and page heaps are now locking and thus safe to use from any context. The "backed" heap uses the locking wrapper, and a "locked" variant of general (using another mcache instance) is created for general-purpose allocations outside of the kernel lock. Allocations made under protection of the kernel lock should continue to use the general heap, as this will avoid unnecessary spinlock operations.

Note: SMP is enabled temporarily in this branch just to run through CI - it will be removed before merging (and until we agree it's safe to enable by default).


francescolavra (commented):
In init_pagecache(), pc->completions is initialized as an objcache, and is not locked. It should be locked instead (there is also an XXX comment that says this at pagecache.c:1320), because completions can be allocated and deallocated from that objcache without the kernel lock.

francescolavra:
In platform/pc/service.c, offset_block_io() calls allocate_merge() on the general kernel heap; since this can be called without the kernel lock (e.g. via pagecache_write_sg_finish -> filesystem_storage_write -> write_extent), it should use a locked heap instead.

wjhun:
I addressed the errors @francescolavra found and removed the unnecessary size parameter from locking_heap_wrapper.

PR #1338, opened 2020-11-23
use irq-safe enqueue when scheduling to bhqueue and runqueue

wjhun:
The queue type that we use to implement scheduling queues allows for lock-free access between concurrent enqueues and dequeues. However, an enqueue that occurs in interrupt level could potentially hang if an enqueue to the same scheduling queue was interrupted. Thus any non-interrupt-level enqueues that could compete with an interrupt-level enqueue must be placed within a critical section / irq disable.

This isn't likely to be a source of failures at the moment given that most of our kernel processing occurs with interrupts disabled. But, in keeping with the precedent of not assuming that this will always be the case, irq-safe wrappers to the queue methods have been added here, and enqueues to the bhqueue or runqueue that aren't exclusively from within an interrupt handler have been changed to use enqueue_irqsafe. (Presently, only bhqueue and runqueue may have enqueues that occur directly from interrupt handlers.)


francescolavra (commented):
Looks good to me

PR #1340, opened 2020-11-25
Fix/smp fixes

francescolavra:
This PR contains 2 commits for concurrent access protection:
- backed heap: unmap virtual memory before deallocating it
- virtqueue: make message enqueuing SMP-safe


wjhun (approved):
Looks good

Issue #1336, opened 2020-11-22
fix 64kB stage2 size limit


wjhun:
The stage2 disk image must remain under 64kB to allow loading it in a single bios int 13H call from within stage1. While we don't want stage2 to get any larger, this can be a show stopper when stage2 code compiled with debug code or lower optimization causes this boundary to be exceeded.

Investigate improving the stage1 bootloader to make multiple bios calls to fulfill the stage2 read if needed.

estherlyoon:
My team members and I can start investigating this

wjhun:
> My team members and I can start investigating this

Ok! Have a look at platform/pc/boot/stage1.s. You'll see the int 13h call to read stage2. Try to inflate the size of stage2 to exceed 64kB (disabling the size check in mkfs) and see if you can trigger a failure in stage1. Then perhaps change stage1 to do a loop of read calls and see if that fixes the issue.

francescolavra:
Closed via https://github.com/nanovms/nanos/pull/1342.

PR #1342, opened 2020-11-26
stage2 size limit

nsharma1231:
fix 64kB stage2 size limit #1336


estherlyoon:
@wjhun we have artificially increased stage2's size and have removed the mkfs size check. However, we are unsure of how to actually run the VM in order to trigger a failure?

francescolavra:
@estherlyoon there is no way to ssh in the VM started via `make run`. If the console output shows the following lines:
```
assigned: 10.0.2.15
Server started on port 8080
```
then the VM is running correctly, otherwise there is a failure.

estherlyoon:
@francescolavra the lines you mentioned do indeed appear. Does this mean this issue doesn't actually need modifications to be made to account for the larger stage2 size?

francescolavra:
> @francescolavra the lines you mentioned do indeed appear. Does this mean this issue doesn't actually need modifications to be made to account for the larger stage2 size?

No, I think it means you didn't actually increase stage2 size beyond 64 KB. What is the size of your output/platform/pc/boot/boot.img file? It should be more than 64 KB in order to trigger a failure

estherlyoon:
> 
> No, I think it means you didn't actually increase stage2 size beyond 64 KB. What is the size of your output/platform/pc/boot/boot.img file? It should be more than 64 KB in order to trigger a failure

We were increasing the size by 1KB, but it turns out we should've gone higher. At around 66KB we get the output `((?>>1))` instead of the correct lines- would this indicate a failure?

francescolavra:
Yes, that indicates there is some error in the boot process

wjhun:
> > No, I think it means you didn't actually increase stage2 size beyond 64 KB. What is the size of your output/platform/pc/boot/boot.img file? It should be more than 64 KB in order to trigger a failure
> 
> We were increasing the size by 1KB, but it turns out we should've gone higher. At around 66KB we get the output `((?>>1))` instead of the correct lines- would this indicate a failure?

If you're getting the same behavior I saw, the bios int 13h, ah = 42h call, "extended read sectors from drive," is failing due to the number of sectors exceeding 128 (64kB). The read never occurs, and stage1 just jumps into uninitialized memory.

I would approach this by making the readsectors procedure in stage1.s into a loop which loads stage2 in 64kB sections. It could load dap.sector_count, which is computed by the assembler, into an accumulator, cap the count in the dap at 64kB, call int 13h, decrement the accumulator by the dap count, increment the load offset in the dap, and loop if nonzero. (Checking for an error return from int 13h and reporting to the console could be helpful, too.)

see:

https://wiki.osdev.org/ATA_in_x86_RealMode_(BIOS)
https://en.wikipedia.org/wiki/INT_13H#INT_13h_AH=42h:_Extended_Read_Sectors_From_Drive

estherlyoon:
@wjhun we've added the read loop and error checking. Please let us know if we should modify anything!

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,40 @@ dap:
...
+%ifdef DEBUG
+        jc sector_read_error
+%endif
+        sub eax, [dap.sector_count]
+        cmp eax, 0
+        jle done 
+        mov ecx, [dap.sector_count]
+        imul ecx, sectorsize
+        add [dap.offset], ecx
+        mov ecx, 0x80
```
wjhun, line 104: this should be the minimum of 0x80 and [dap.sector_count]

rules.mk
```diff
@@ -65,7 +65,7 @@ LCOV=		lcov
 GENHTML=	genhtml
 
 CFLAGS+=	-std=gnu11 -O3 -g
-CFLAGS+=	-Wall -Werror -Wno-char-subscripts
+CFLAGS+=	-Wno-char-subscripts
```
wjhun, line 68: I think we should keep CFLAGS as-is.

esthery.txt
```diff
@@ -0,0 +1,31 @@
+platform/pc/service.c
```
wjhun, line 1: looks like a notes file got checked in here?

wjhun (commented):
Just one comment about the sector count in stage1, but otherwise the updated asm looks good. Looks like some cruft made it in, and also the size check removal from mkfs was omitted - maybe it got lost in a merge / rebase?

platform/pc/boot/stage1.s
wjhun: Sorry, I meant to say [dap.sector_count] should be updated with the minimum of 0x80 and eax.

estherlyoon:
> Just one comment about the sector count in stage1, but otherwise the updated asm looks good. Looks like some cruft made it in, and also the size check removal from mkfs was omitted - maybe it got lost in a merge / rebase?

Oops, sorry about the extraneous stuff. Everything should be updated now.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
...
 
 
+%ifdef DEBUG
+bits 16
+%include "debug.inc"
+bits 32
+%endif
+
 readsectors:
+        mov eax, [dap.sector_count]
```
wjhun, line 85: sector_count is two bytes, so shouldn't this use ax instead of eax? (same applies to all uses of eax / ecx below)

wjhun (commented):
I tried the latest changes and was unable to get the image to load with an inflated (>64kB) stage2 image. Did it work for you?

On further inspection, I see that the accesses to [dap.sector_count] are 32-bits wide, whereas the field is only 16-bit. I changed instances of 'eax' and 'ecx' to 'ax' and 'cx', respectively (except for imul result), and was able to boot. (I didn't verify correctness other than check that it made two int 13h calls and booted up fine...)

Please check that all the accesses are the correct length and verify that you are able to boot a >64kB stage2 image.

(Note: I have had problems debugging real-mode code under qemu for some reason, so I used qemu tracing flags "-d in_asm,cpu -D outfile" to inspect the instruction stream with interleaved register dumps.)

platform/pc/boot/stage1.s
```diff
@@ -95,16 +95,16 @@ loop:
...
         jle done 
-        mov ecx, [dap.sector_count]
+        mov cx, [dap.sector_count]
         imul ecx, sectorsize
-        add [dap.offset], ecx
-        mov ecx, 0x80
-        cmp eax, ecx
-        cmovl ecx, eax 
-        mov [dap.sector_count], ecx
+        add [dap.offset], cx
```
wjhun, line 103: use ecx in this add - a sector_count of 0x80 would yield 2^16 in ecx or 0 in cx

estherlyoon: Right, thanks for pointing that out. I forgot to mention this earlier, but I verified that this booted properly with stage2 image sizes of 66KB and 128KB, respectively, by changing the bs flag for the dd command to those values. I couldn't figure out how to see the qemu trace- I appended the ones you mentioned to `QEMU_FLAGS` in `platform/pc/Makefile` but the output file was empty.

wjhun: > I couldn't figure out how to see the qemu trace- I appended the ones you mentioned to `QEMU_FLAGS` in `platform/pc/Makefile` but the output file was empty.

You'll need to run in emulated (TCG) mode; try making "run-noaccel" or remove --enable-kvm from the qemu command line.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
         .lba          dq 1
 
 
+%ifdef DEBUG
```
wjhun, line 78: I missed this before. We don't actually want this assembled conditionally since we're using it for error output...right now the error report isn't getting built. You'll probably need to specify src/x86_64 in the nasm include path...or just spell it out here.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
...
+bits 16
+%include "debug.inc"
+bits 32
```
wjhun, line 81: We don't want to go to bits 32 here...just the include without either bits statement is fine.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
...
+        jle loop
+        mov ecx, 0x80
+        mov [dap.sector_count], ecx
+loop:
         mov si, dap
         mov ah, 0x42
         mov dl, 0x80
         int 0x13
+%ifdef DEBUG
+        jc sector_read_error
```
wjhun, line 96: If I try this after removing the ifdef DEBUG, it reports an error. Indeed, the int 13h is returning with CF set and error code (AH) set to 0x09, which is "Attempt to DMA over 64kb boundary."

It looks like the jle above is not doing what is expected and is jumping to loop even though the sector_count is 0xff (in my test case). Apparently, this is because the 'cmp' above is doing only a signed byte comparison, and both 0x80 and 0xff are being compared as negative values. Changing the immediate from 0x80 to 0x0080 seems to do the trick (emits a cmpw). Since these are unsigned values, too, it should technically be jbe instead of jle, though that shouldn't matter in this case.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
...
 
+%ifdef DEBUG
+bits 16
+%include "debug.inc"
+bits 32
+%endif
+
 readsectors:
+        mov eax, [dap.sector_count]
+        cmp eax, 0x80
```
wjhun, line 86: These still need to be changed from eax to ax, and 0x80 to 0x0080 (see comment below about comparison width).

wjhun (commented):
It's close, but there are still issues here and I'm unable to boot with an inflated stage2. I highly recommend getting the qemu instruction traces going so that you can verify the behavior at an instruction level.

(Note, I inflated stage2 simply by adding a large dummy array, initialized to something non-zero, and referring to it in the code to make sure it gets linked in.)

platform/pc/boot/stage1.s
estherlyoon: @wjhun upon debugging, we discovered that dap.offset cannot be updated past 128 * 512 bytes (0x10000) due to its width. This seems to imply that the size of the stage2 image cannot exceed 64KB. Would the dap.offset field have to be changed to a long? We tried this, but it triggered a failure in the int 0x13.

wjhun:
> @wjhun upon debugging, we discovered that dap.offset cannot be updated past 128 * 512 bytes (0x10000) due to its width. This seems to imply that the size of the stage2 image cannot exceed 64KB. Would the dap.offset field have to be changed to a long? We tried this, but it triggered a failure in the int 0x13.

So the issue here is that the address is composed in segment:offset format, where the physical address will be segment * 16 + offset. So the loop needs to take this into account and advance the segment as well. E.g. if the next 64kB is going to be loaded at phys addr 0x10000, then the segmented address is 1000:0000.

It looks like some bioses won't like the buffer to cross segment boundaries. So in that case you can adjust the [sector_count] to be the smaller of 0x80, ax and (0x10000 - [dap.offset]). (Actually you should only have to take the offset into account for the first iteration; the rest will be aligned.)

nsharma1231:
@wjhun We tried to modify all of the dap values according to what you said, however, we are now hanging after the interrupt 13h. We were suspicious that this is because we had accidentally corrupted some of the values in the dap structure. However, we couldn't figure out where this was happening. Could you please help us try and figure out where we are going wrong?

platform/pc/Makefile
```diff
@@ -342,6 +342,7 @@ QEMU_NET=	-device $(NETWORK)$(NETWORK_BUS),mac=7e:b8:7e:87:4a:ea,netdev=n0 $(QEM
 QEMU_USERNET=	-device $(NETWORK)$(NETWORK_BUS),netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostfwd=tcp::9090-:9090,hostfwd=udp::5309-:5309
 #QEMU_USERNET+=	-object filter-dump,id=filter0,netdev=n0,file=/tmp/nanos.pcap
 QEMU_FLAGS=
+QEMU_FLAGS+= -d in_asm,cpu -D outfile
```
wjhun, line 345: Please remove this; this will generate a very large output file and could interfere with CI.

platform/pc/boot/stage1.s
```diff
@@ -75,11 +75,42 @@ dap:
...
+        mov [dap.sector_count], cx
+loop:
         mov si, dap
         mov ah, 0x42
         mov dl, 0x80
         int 0x13
+        jc sector_read_error
+        sub ax, [dap.sector_count]
+        cmp ax, 0
+        je done
```
wjhun, line 100: I found two issues here: as stated above, the segment needs to advance by 0x1000, as it is shifted left by 4 and added to the offset to make the effective address. Also, [dap.lba] needs to be advanced.

I was able to get the kernel to boot (with a >64kB stage2) using the following patch:

```
diff --git a/platform/pc/boot/stage1.s b/platform/pc/boot/stage1.s
index 3769fb69..2afc186f 100644
--- a/platform/pc/boot/stage1.s
+++ b/platform/pc/boot/stage1.s
@@ -97,10 +97,10 @@ loop:
         jc sector_read_error
         sub ax, [dap.sector_count]
         cmp ax, 0
-        je done 
-        mov dx, [dap.segment]
-        inc dx
-        mov [dap.segment], dx
+        je done
+        add [dap.lba], cx
+        mov cx, 0x1000
+        add [dap.segment], cx
         mov dx, 0x0
         mov [dap.offset], dx
         mov cx, 0x0080
```

wjhun (changes requested):
This is nearly there - please try the small fix in the comments and see if it works for you. Also, there's some cruft to remove here, like "nasm.out", "platform/pc/outfile" and the added qemu flags in the Makefile.

wjhun (approved):
I went ahead and applied the fixes and rebased. We've had some users run into this issue while building, so we'd like to expedite the merge.

@estherlyoon, @nsharma1231 and @samantha3pen: Thank you for the contribution!

PR #1343, opened 2020-11-28
Klibs: add cloud_init klib

### LINKED ISSUE #1319: add cloud-init klib for azure
right now we create an instance on azure but it takes a while to create and eventually times out

I think what's happening is that azure is trying to insert a user/ssh key into the vm (which it obv can't) - I don't know if this is just an api issue that we can set an option for? or perhaps we could hack around this and make it pretend like the operation succeeded?

francescolavra:
At the first boot of the kernel for a given image, this library detects (by reading the DMI tables) the cloud platform on which the kernel is running, and executes a cloud-specific initialization. The only cloud platform currently supported is Azure.

Closes #1319.

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -2,9 +2,14 @@ LWIPDIR=		$(VENDORDIR)/lwip
 MBEDTLS_DIR=	$(VENDORDIR)/mbedtls
 
 PROGRAMS= \
+	cloud_init \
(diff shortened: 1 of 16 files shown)


sanderssj (commented):
Looks good to me

wjhun (approved):
Looks good to me

PR #1344, opened 2020-11-28
Makefile: add kernel libraries to releases

### LINKED ISSUE #1339: nightly && release builds should include klibs in pkg
we're going to keep klibs tied to releases - while we could potentially have some sort of pkghub like interface in the future  let's just bundle them up in the release tarballs for now

francescolavra:
Kernel libraries are placed in the klibs subfolder of the release artifact.

Closes #1339.


sanderssj (commented):
Looks good to me!

wjhun (approved):

PR #1347, opened 2020-11-30
Fix syscall setup and implement getcpu syscall

sanderssj:
This PR fixes syscall setup for SMP by guaranteeing set_syscall_handler is run on the first processor. It also implements the getcpu syscall for userspace programs that want to know which cpu they're running on.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -177,7 +177,6 @@ void register_other_syscalls(struct syscall *map)
     register_syscall(map, open_by_handle_at, 0);
     register_syscall(map, clock_adjtime, 0);
     register_syscall(map, setns, 0);
-    register_syscall(map, getcpu, 0);
     register_syscall(map, process_vm_readv, 0);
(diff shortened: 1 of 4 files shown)


src/unix/syscall.c
```diff
@@ -2333,6 +2332,20 @@ sysreturn umask(int mask)
     return mask;
 }
 
+sysreturn getcpu(unsigned int *cpu, unsigned int *node, void *tcache)
+{
+    cpuinfo ci = current_cpu();
+    if (!validate_user_memory(cpu, sizeof *cpu, true) ||
+            !validate_user_memory(node, sizeof *node, true))
+        return -EFAULT;
```
francescolavra, line 2340: Before calling validate_user_memory(), the pointers should be checked against a null value (otherwise -EFAULT would be returned when either `cpu` or `node` is NULL, which is a valid value)

sanderssj: Good catch, thanks

wjhun (approved):
looks good

PR #1348, opened 2020-12-02
Radar klib: add crash reporting functionality

francescolavra:
Crash reporting to the radar server is enabled when the manifest file contains the RADAR_KEY environment variable.
A "crash" is defined as either a user application fatal error (e.g. anything that causes the application to terminate with a non-zero
exit code), or a kernel fatal error. When any of these happens, before shutting down the VM the kernel dumps on disk a trace of log messages (printed by both the user application and the kernel); at the next boot, the radar klib detects the log dump and sends it to the radar server as a crash report; the value of the "id" field of a crash report is the same as the "boot id" (sent right after each boot), so that a crash can be unequivocally associated to a boot.

The log dump is saved in a section of the disk that is being carved between stage2 and the boot filesystem.

Throughout the kernel, most calls to console() have been replaced by calls to a new function, rputs(), which prints the supplied string to both the console and the kernel log (so that in case of a crash the console messages will be sent in the crash report).
Analogously, the buffer_print() and stdout() functions have been changed so that their output goes to both the console and the
kernel log.

### DIFF
--- a/klib/radar.c
+++ b/klib/radar.c
@@ -3,12 +3,6 @@
 #include <lwip.h>
 
 #define RADAR_HOSTNAME  "radar.relayered.net"
-
-#define RADAR_IP_1  34
-#define RADAR_IP_2  94
-#define RADAR_IP_3  88
-#define RADAR_IP_4  162
-
 #define RADAR_PORT      443
 
 #define RADAR_CA_CERT   "-----BEGIN CERTIFICATE-----\n\
@@ -50,6 +44,7 @@ static struct telemetry {
     s64 boot_id;
     timestamp boot_backoff;
     closure_struct(boot_timer_func, boot_func);
+    void (*rprintf)(const char *format, ...);
     tuple (*allocate_tuple)(void);
     void (*table_set)(table z, void *c, void *v);
     void *(*table_find)(table z, void *c);
@@ -62,6 +57,8 @@ static struct telemetry {
     void (*bprintf)(buffer b, const char *fmt, ...);
     timer (*register_timer)(clock_id id, timestamp val, boolean absolute,
             timestamp interval, timer_handler n);
+    err_t (*dns_gethostbyname)(const char *hostname, ip_addr_t *addr,
+            dns_found_callback found, void *callback_arg);
     status (*http_request)(heap h, buffer_handler bh, http_method method,
             tuple headers, buffer body);
     int (*tls_connect)(ip_addr_t *addr, u16 port, connection_handler ch);
@@ -76,6 +73,20 @@ static struct telemetry {
 #define buffer_write_cstring(b, s)  kfunc(buffer_write)(b, s, sizeof(s) - 1)
 
 static void telemetry_boot(void);
+
+static void telemetry_dns_cb(const char *name, const ip_addr_t *ipaddr, void *callback_arg)
+{
+    connection_handler ch = (connection_handler)callback_arg;
+    if (ipaddr) {
+        if (telemetry.tls_connect((ip_addr_t *)ipaddr, RADAR_PORT, ch) == 0)
+            return;
+        else
+            kfunc(rprintf)("Radar: failed to connect to server\n");
+    } else {
+        kfunc(rprintf)("Radar: failed to look up server hostname\n");
+    }
+    deallocate_closure(ch);
+}
 
 static boolean telemetry_req(const char *url, buffer data, buffer_handler bh)
 {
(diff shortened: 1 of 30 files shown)


klib/radar.c
```diff
@@ -1,13 +1,14 @@
...
+#include <log.h>
 #include <lwip.h>
 
 #define RADAR_HOSTNAME  "radar.relayered.net"
 
-#define RADAR_IP_1  34
-#define RADAR_IP_2  94
-#define RADAR_IP_3  88
-#define RADAR_IP_4  162
+#define RADAR_IP_1  35
```
eyberg, line 8: can we replace this w/just the host lookup? this is going to change quite a lot

platform/pc/service.c
```diff
@@ -374,7 +380,14 @@ void kernel_shutdown(int status)
 {
     shutting_down = true;
     if (root_fs) {
-        storage_sync(closure(heap_general(&heaps), sync_complete, status));
+        heap h = heap_locked(&heaps);
+        status_handler completion = closure(h, sync_complete, status);
+        merge m = allocate_merge(h, completion);
+        status_handler sh = apply_merge(m);
+        if ((status != 0) && table_find(filesystem_getroot(root_fs), sym(crash_report)))
```
eyberg, line 387: do we need an extra bool for this? wouldn't having the RADAR_API env token specified should be enough?

klib/radar.c
francescolavra: OK, I will enable DNS support in lwIP and use that to do a host lookup

platform/pc/service.c
francescolavra: OK, I'll get rid of the crash_report flag, and use just the RADAR_KEY environment variable

src/kernel/log.h
```diff
@@ -0,0 +1,19 @@
+typedef struct klog_dump {
+    u8 header[4];
+    s64 boot_id;
+    int exit_code;
```
sanderssj, line 4: This is might be too nitpicky, but for portability on ILP64 platforms, to keep the header size correct it might be better to use a fixed size type for exit_code.

francescolavra: Good point, I made `exit_code` an s32

src/kernel/log.c
```diff
@@ -0,0 +1,107 @@
...
+}
+KLIB_EXPORT(klog_load);
+
+void klog_save(int exit_code, status_handler sh)
+{
+    bytes msg_len = 0;
+    if (klog.count > 0) {
+        klog_lock();
+        msg_len = MIN(klog.count, sizeof(klog.dump.msgs) - 1);
+        bytes index = (klog.count - msg_len) & KLOG_BUF_SIZE_MASK;
```
wjhun, line 84: It's not clear to me why msg_len is being subtracted from klog.count. Wouldn't (klog.count & KLOG_BUF_SIZE_MASK) suffice to find the top of the log?

Also, is there any particular reason that klog.count needs to be free-running (as opposed to wrapping) other than to allow the msg_len computation above? I might be missing something, but it seems like it might be simpler to store size and offset independently.

wjhun (approved):
Everything looks good here - just one question about the index calculation in klog_save that I probably just didn't grok.

src/kernel/log.c
francescolavra: (klog.count & KLOG_BUF_SIZE_MASK) would be the bottom of the log, to find the top we need to subtract the actual length of the log messages that we are going to save; if the in-memory log buffer is full, the top is at a fixed offset (given by the difference between the in-memory buffer size and the size of the on-disk message dump) from the bottom, otherwise it's 0.
I used a free-running counter mainly because it simplifies things when writing to the log. Storing size and offset instead of a free-running counter would make the code slightly simpler when saving to disk, at the cost of a slightly more complex housekeeping in klog_write(), so I don't see a net advantage in changing the current code.

klib/radar.c
```diff
@@ -75,8 +77,17 @@ static struct telemetry {
 /* To be used with literal strings only */
 #define buffer_write_cstring(b, s)  kfunc(buffer_write)(b, s, sizeof(s) - 1)
 
+static void telemetry_crash_report(void);
 static void telemetry_boot(void);
 
+static void telemetry_dns_cb(const char *name, const ip_addr_t *ipaddr, void *callback_arg)
+{
+    connection_handler ch = (connection_handler)callback_arg;
+    if (ipaddr && (telemetry.tls_connect((ip_addr_t *)ipaddr, RADAR_PORT, ch) == 0))
```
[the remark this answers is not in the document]
francescolavra: Yes, I added logging of connection errors here, so a failure to contact the server wouldn't be unnoticed.

Issue #457, opened 2019-01-25
implement get_robust_list, set_robust_list


wjhun:
This is needed to indicate which futexes should be unlocked upon thread exit.

wjhun:
Resolved in #1351

PR #1351, opened 2020-12-04
Add support for robust futex syscalls and handling

sanderssj:
This change implements the get_robust_list and set_robust_list syscalls
as well as adds support for handling locked mutices during thread exit
by marking the mutex and waking any waiters. The test program included
verifies basic robust mutex functionality. See #457 for more information.

### DIFF
--- a/src/unix/futex.c
+++ b/src/unix/futex.c
@@ -294,3 +294,94 @@ init_futices(process p)
         halt("failed to allocate futex table\n");
 
 }
+
+/* robust mutex handling */
+
+#define FUTEX_OWNER_DIED	0x40000000
+#define FUTEX_KEY_ADDR(x, o)    ((u32 *)((u8 *)(x) + (o)))
+
+typedef struct robust_list {
+    struct robust_list *next;
+} *robust_list;
+
+typedef struct robust_list_head {
+    robust_list list;
+    long futex_offset;
+    void *list_op_pending;
+} *robust_list_head;
+
+void wake_robust_list(process p, void *head)
+{
+    struct futex * f;
+    struct robust_list_head *h = head;
+    struct robust_list *l;
+    u32 *uaddr;
+
+    if (!h)
+        return;
+    /* XXX could keep a list of futexes and wake them at the end
+     * to let threads acquire multiple locks without blocking */
+    if (h->list_op_pending) {
+        uaddr = FUTEX_KEY_ADDR(h->list_op_pending, h->futex_offset);
+        if ((f = table_find(p->futices, uaddr))) {
+            *uaddr |= FUTEX_OWNER_DIED;
+            futex_wake_many(f, 1);
+        } else {
+            thread_log(current, "list_op_pending futex not found for key %p", uaddr);
+        }
+    }
+
+    for (l = h->list; (void *)l != (void *)h; l = l->next) {
+        uaddr = FUTEX_KEY_ADDR(l, h->futex_offset);
+        if ((f = table_find(p->futices, uaddr))) {
+            *uaddr |= FUTEX_OWNER_DIED;
+            futex_wake_many(f, 1);
+        } else {
+            rprintf("%s: list futex not found for key %p, aborting iteration\n",
+                __func__, uaddr);
+            break;
+        }
+    }
+}
+
+sysreturn get_robust_list(int pid, void *head, u64 *len)
+{
+    struct robust_list_head **hp = head;
+    if (!validate_user_memory(hp, sizeof *hp, true))
+        return set_syscall_error(current, EFAULT);
+    if (!validate_user_memory(len, sizeof *len, true))
+        return set_syscall_error(current, EFAULT);
+
+    thread_log(current, "get_robust_list syscall for pid %d\n", pid);
+
+    thread t = 0;
+    if (pid == 0)
+        t = current;
+    else {
+        process p = current->p;
+        thread tt;
+        vector_foreach(p->threads, tt) {
+            if (tt->tid == pid) {
+                t = tt;
+                break;
+            }
+        }
+    }
+    if (t == 0)
+        return set_syscall_error(current, ESRCH);
+    *hp = t->robust_list;
+    *len = sizeof **hp;
+    return set_syscall_return(current, 0);
+}
+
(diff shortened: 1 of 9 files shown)


src/unix/futex.c
```diff
@@ -294,3 +294,106 @@ init_futices(process p)
...
+        return;
+
+    /* XXX could keep a list of futexes and wake them at the end
+     * to let threads acquire multiple locks without blocking */
+    if (h->list_op_pending) {
+        uaddr = FUTEX_KEY_ADDR(h->list_op_pending, h->futex_offset);
+        if (valid_user_address(p, uaddr, sizeof *uaddr)) {
+            *uaddr |= FUTEX_OWNER_DIED;
+            if ((f = table_find(p->futices, uaddr)))
+                futex_wake_many(f, 1);
```
francescolavra, line 343: Shouldn't all waiters (instead of just one) be woken up? I think we should use INT_MAX (which is the value used by FUTEX_WAKE to indicate that all waiters should be woken up) instead of 1.
Also, we could use futex_wake_many_by_uaddr(), which does the table lookup.

src/unix/futex.c
```diff
@@ -294,3 +294,106 @@ init_futices(process p)
...
+
+    for (l = h->list; (void *)l != (void *)h; l = l->next) {
+        uaddr = FUTEX_KEY_ADDR(l, h->futex_offset);
+        if (!valid_user_address(p, l, sizeof *l))
+            break;
+        if (!valid_user_address(p, uaddr, sizeof *uaddr))
+            break;
+        *uaddr |= FUTEX_OWNER_DIED;
+        if ((f = table_find(p->futices, uaddr)))
+            futex_wake_many(f, 1);
```
francescolavra, line 355: Same here.

src/unix/futex.c
```diff
@@ -294,3 +294,106 @@ init_futices(process p)
...
+    }
+}
+
+sysreturn get_robust_list(int pid, void *head, u64 *len)
+{
+    struct robust_list_head **hp = head;
+    if (!validate_user_memory(hp, sizeof *hp, true))
+        return set_syscall_error(current, EFAULT);
```
francescolavra, line 363: I know that this is being done in other syscalls too, but it's a bit wasteful to call `set_syscall_error()` (or `set_syscall_return()`) in syscalls that return directly without blocking, because the return value is set in syscall_debug(), so I think in new code we should avoid this double setting and just return the value (with minus sign if it's an error).

src/unix/futex.c
sanderssj: I went with one since that's how it's described in Ingo Molnar's 2006 email about kernel behavior for robust lists, and that's also what pthread_mutex_unlock uses for the futex wake syscall. I think it makes sense to just wake the next in queue blocked on the lock. I might be missing something though...what situation are you thinking of where waking all would be beneficial?
Thanks for the tip about futex_wake_many_by_uaddr!

src/unix/futex.c
sanderssj: Good point, thanks!

src/unix/futex.c
francescolavra: I was under the assumption that a futex held by a dead thread wasn't usable anymore, which is of course not the case. Now I think you are right, it makes more sense to wake up just one blocked thread

src/unix/futex.c
```diff
@@ -294,3 +294,103 @@ init_futices(process p)
...
+    long futex_offset;
+    void *list_op_pending;
+} *robust_list_head;
+
+static boolean valid_user_address(process p, void *a, int len)
+{
+    u64 pg = u64_from_pointer(a) & ~PAGEMASK;
+    u64 end = (u64_from_pointer(a) + len) & ~PAGEMASK;
+
+    for (; pg <= end; pg += PAGESIZE) {
```
wjhun, line 318: I would suggest making this a more general function like "vmap_validate_range." Also, I think it would be much more efficient to use rangemap_range_find_gaps with a stack closure and a closed over a valid boolean - just set to false if a gap is found.

src/unix/futex.c
```diff
@@ -294,3 +294,103 @@ init_futices(process p)
...
+    *len = sizeof **hp;
+    return 0;
+}
+
+sysreturn set_robust_list(void *head, u64 len)
+{
+    thread_log(current, "set_robust_list syscall with head %p", head);
+    if (len != sizeof(struct robust_list_head))
+        return -EINVAL;
+    if (!validate_user_memory(head, len, true))
```
wjhun, line 391: Should validate_user_address be called for this range as well?

src/unix/futex.c
```diff
@@ -294,3 +294,103 @@ init_futices(process p)
...
+
+    thread_log(current, "get_robust_list syscall for pid %d\n", pid);
+
+    thread t = 0;
+    if (pid == 0)
+        t = current;
+    else {
+        process p = current->p;
+        thread tt;
+        vector_foreach(p->threads, tt) {
```
wjhun, line 372: Isn't p->threads indexed by tid (pid here)? I think this could just be vector_get(p->threads, pid).

src/unix/futex.c
```diff
@@ -294,3 +294,103 @@ init_futices(process p)
...
+}
+
+void wake_robust_list(process p, void *head)
+{
+    struct robust_list_head *h = head;
+    struct robust_list *l;
+    int *uaddr;
+
+    /* must be very careful accessing the head as well as the list */
+    if (!valid_user_address(p, h, sizeof *h))
```
wjhun, line 332: A stylistic nit, but elsewhere in the codebase we've been using parentheses with sizeof.

wjhun (approved):
Looks good ... just a few suggestions for optimization.

src/unix/futex.c
sanderssj: Refactored as you suggested and it looks a lot nicer!

src/unix/futex.c
sanderssj: I actually tried to roll this new validation into validate_user_address but it breaks some other runtime tests that use addresses that are not in the process vmaps ranges in syscalls, so I left it separate for now. I used the new validation here instead of validate_user_address.

wjhun (approved):
New changes look good!

Issue #1166, opened 2020-05-05
filesystem benchmark


wjhun:
We should either import or create a little benchmark runtime test to measure file I/O performance so we can keep track of improvements and regressions as we develop tfs.

wjhun:
This was addressed to some degree with the addition of a simple write test in #1191. Will keep this open until we come upon a more comprehensive test, however.

uma-sethuraman:
Hello @wjhun, my team members and I are interested in contributing to this issue. Is there still something we can add to the filesystem benchmark test?

wjhun:
@uma-sethuraman  You are certainly welcome to expand on the crude little test that lives in test/runtime/write.c, but we don't have a specific ask so much. We've mostly used it to compare performance of different drivers and to try and detect (by manual inspection, alas) performance regressions.

A more exhaustive test could certainly be useful, especially if it allows us to compare metrics across platforms. But we haven't enumerated any criteria (@francescolavra - do you have any suggestions here based on what you've found bringing up databases on nanos?), so you would kind of be on your own (or could just generally draw from what other generic fs benchmarks measure - or even just port one, if the license is permissible - e.g. no GPL)...

francescolavra:
I think a useful addition to the tests in test/runtime/write.c would be some code that creates a large number (says, thousands) of files, writes a small amount of data in each file, syncs the filesystem with a call to sync(), then deletes the files and syncs again. This would stress-test many parts of the FS and pagecache code, and provide a workload somewhat similar to that of a database application.

uma-sethuraman:
Thank you @francescolavra @wjhun ! We are interested in this issue and will start working on adding what Francesco mentioned to test/runtime/write.c.

PR #1352, opened 2020-12-06
Addition to Filesystem Benchmark Testing

uma-sethuraman:
Hello @wjhun @francescolavra, this pull request is for issue #1166. As mentioned in the comments of that issue, we added a filesystem stress test to `test/runtime/write.c` which creates a large number of files, writes data to each of those files, syncs the filesystem, deletes all the files, and then syncs again. Please let us know what you think of the test and if there's anything we should change or add.

### DIFF
--- a/test/runtime/write.c
+++ b/test/runtime/write.c
@@ -12,7 +12,6 @@
 #include <sys/time.h>
 #include <sys/eventfd.h>
 #include <sys/stat.h>
-#include <unix/syscalls.h>
 
 //#define WRITETEST_DEBUG
 #ifdef WRITETEST_DEBUG
@@ -622,6 +621,7 @@ static void fs_stress_test()
 {
     int num_files = 1000;
     int fds[num_files];
+    char *fd_names[num_files];
 
     /* Creating and writing to new files */
     int i;
@@ -632,21 +632,17 @@ static void fs_stress_test()
         sprintf(name, "fs_stress_test_%d", i);
         int fd = open(name, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
         fds[i] = fd;
+        fd_names[i] = name;
         if (fd < 0) {
-            printf("open error\n");
             perror("open");
             exit(EXIT_FAILURE);
         }
 
         _READ(buf, BUFLEN);
 
-        if (rv == 0)
-            writetest_debug("empty source file\n");
-
         if (rv >= BUFLEN)
             rv = BUFLEN - 1;
         buf[rv] = '\0';
-        writetest_debug("original: \"%s\"\n", buf);
 
         _LSEEK(0, SEEK_SET);
 
@@ -655,14 +651,25 @@ static void fs_stress_test()
     }
 
     /* sync the filesystem */
-    syscall(SYS_sync); 
+    sync(); 
 
     /* Deleting all files we just created */
-    for (i = 0; i < num_files; i++) 
+    for (i = 0; i < num_files; i++) {
         close(fds[i]);
+        char path[50];
+        sprintf(path, "./%s", fd_names[i]);
+        remove(path);
+
+        /* Confirms file is deleted */
+        int deleted_fd = open(fd_names[i], O_RDONLY, 0);
+        if (deleted_fd >= 0) {
+            perror("shouldn't be open");
+            exit(EXIT_FAILURE);
+        }
+    }
 
     /* sync the filesystem again */
-    syscall(SYS_sync); 
+    sync(); 
     return;
 
   out_fail:
--- a/test/runtime/write.c
+++ b/test/runtime/write.c
@@ -685,7 +685,7 @@ static void fs_stress_test()
         perror("fs_stress_test: clock_gettime");
         goto out_fail;
     }
-    print_op_stats("syncs and delete", &postwrite, &postsync, 0);
+    print_op_stats("sync & delete", &postwrite, &postsync, 0);
     print_op_stats("total", &start, &postsync, 0);
 
     return;
--- a/test/runtime/write.c
+++ b/test/runtime/write.c
@@ -628,16 +628,17 @@ static void fs_stress_test()
     for (i = 0; i < num_files; i++) {
         char buf[BUFLEN];
         ssize_t rv;
-        printf("fs_stress before open\n");
-        int fd = open("fs_stress_"+i, O_RDWR);
-        printf("fs_stress after open\n");
+        char name[25];
+        sprintf(name, "fs_stress_test_%d", i);
+        printf("name in STRESS: %s\n", name);
+        int fd = open(name, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
+        printf("fd in stress: %d\n", fd);
         fds[i] = fd;
         if (fd < 0) {
             printf("open error\n");
             perror("open");
             exit(EXIT_FAILURE);
         }
-        printf("outside of if statement\n");
 
         _READ(buf, BUFLEN);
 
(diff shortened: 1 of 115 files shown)


test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+        writetest_debug("original: \"%s\"\n", buf);
+
+        _LSEEK(0, SEEK_SET);
+
+        ssize_t len = strlen(str);
+        _WRITE(str, len);
+    }
+
+    /* sync the filesystem */
+    syscall(SYS_sync);
```
francescolavra, line 658: Couldn't you just call the sync() function here?

test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+
+    /* sync the filesystem */
+    syscall(SYS_sync); 
+
+    /* Deleting all files we just created */
+    for (i = 0; i < num_files; i++) 
+        close(fds[i]);
+
+    /* sync the filesystem again */
+    syscall(SYS_sync);
```
francescolavra, line 665: Same here, you could just call the sync() function

test/runtime/write.c
```diff
@@ -12,6 +12,7 @@
 #include <sys/time.h>
 #include <sys/eventfd.h>
 #include <sys/stat.h>
+#include <unix/syscalls.h>
```
francescolavra, line 15: This include shouldn't be needed anymore once `syscall(SYS_sync)` is replaced by `sync()`

test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+    for (i = 0; i < num_files; i++) {
+        char buf[BUFLEN];
+        ssize_t rv;
+        char name[25];
+        sprintf(name, "fs_stress_test_%d", i);
+        int fd = open(name, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
+        fds[i] = fd;
+        if (fd < 0) {
+            printf("open error\n");
+            perror("open");
```
francescolavra, line 637: The printf above is redundant, since there is the perror() call

test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+        ssize_t len = strlen(str);
+        _WRITE(str, len);
+    }
+
+    /* sync the filesystem */
+    syscall(SYS_sync); 
+
+    /* Deleting all files we just created */
+    for (i = 0; i < num_files; i++) 
+        close(fds[i]);
```
francescolavra, line 662: Closing a file does not delete it

test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+        if (fd < 0) {
+            printf("open error\n");
+            perror("open");
+            exit(EXIT_FAILURE);
+        }
+
+        _READ(buf, BUFLEN);
+
+        if (rv == 0)
+            writetest_debug("empty source file\n");
```
francescolavra, line 644: This writetest_debug() call is not really useful, since all files will be empty and this would just pollute the debug logs with 1000 identical lines

test/runtime/write.c
```diff
@@ -617,6 +618,59 @@ static void usage(const char *program_name)
...
+
+        if (rv >= BUFLEN)
+            rv = BUFLEN - 1;
+        buf[rv] = '\0';
+        writetest_debug("original: \"%s\"\n", buf);
```
francescolavra, line 649: Same here

francescolavra (changes requested):

test/runtime/write.c
samantha3pen: @francescolavra we used remove() and added a check to ensure the files were properly removed. Is this synonymous to deleting a file?

test/runtime/write.c
```diff
@@ -617,6 +617,67 @@ static void usage(const char *program_name)
...
+    /* Creating and writing to new files */
+    int i;
+    for (i = 0; i < num_files; i++) {
+        char buf[BUFLEN];
+        ssize_t rv;
+        char name[25];
+        sprintf(name, "fs_stress_test_%d", i);
+        int fd = open(name, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
+        fds[i] = fd;
+        fd_names[i] = name;
```
francescolavra, line 635: `name` is a variable that lives only within the `for()` loop, you cannot assign its address to fd_names[i] and then use fd_names[i] outside the `for()` loop.

test/runtime/write.c
```diff
@@ -617,6 +617,67 @@ static void usage(const char *program_name)
...
+        if (fd < 0) {
+            perror("open");
+            exit(EXIT_FAILURE);
+        }
+
+        _READ(buf, BUFLEN);
+
+        if (rv >= BUFLEN)
+            rv = BUFLEN - 1;
+        buf[rv] = '\0';
```
francescolavra, line 645: What's the purpose of the 3 lines above?

test/runtime/write.c
francescolavra: Yes, remove() can be used to delete a file, but if you check the return value of remove() in your code you will see that it's an error value for most (all?) files, so it's not being used correctly and is not actually deleting those files.

francescolavra (commented):
It would be good for the FS stress test to print on the standard output how long it takes to execute, and more specifically how long it takes up to the first `sync()` call, and then how long the remaining part take, so that we have some statistics shown on the terminal when the test runs, similarly to how it's done in bulk_write_test().

test/runtime/write.c
uma-sethuraman: We actually were using these lines with our `writetest_debug` statements, but since we got rid of those, these lines are no longer needed. We deleted these three lines.

uma-sethuraman:
@francescolavra we updated the test to print out statistics similar to bulk_write_test() and have made the changes you mentioned earlier. We also fixed the errors we were getting from the remove() statement. Let us know if there's anything else we should change.

test/runtime/write.c
```diff
@@ -617,6 +617,85 @@ static void usage(const char *program_name)
...
+        _READ(buf, BUFLEN);
+
+        _LSEEK(0, SEEK_SET);
+
+        ssize_t len = strlen(str);
+        _WRITE(str, len);
+    }
+
+    /* Time up to first sync() call */
+    struct timespec postwrite;
```
francescolavra, line 655: Actually I would like the first stats to be calculated after the first sync() call, not before.

test/runtime/write.c
```diff
@@ -617,6 +617,85 @@ static void usage(const char *program_name)
...
+    /* sync the filesystem again */
+    sync(); 
+
+    /* Time of remaining and total time*/
+    struct timespec postsync;
+    if (clock_gettime(CLOCK_MONOTONIC, &postsync) < 0) {
+        perror("fs_stress_test: clock_gettime");
+        goto out_fail;
+    }
+    print_op_stats("sync & delete", &postwrite, &postsync, 0);
```
francescolavra, line 688: Should be called just "delete", not "sync & delete"

test/runtime/write.c
```diff
@@ -688,6 +767,8 @@ int main(int argc, char **argv)
         persistence_write_test();
     }
 
+    fs_stress_test();
+
```
francescolavra, line 771: This should be under the "basic" tests, so this function call should be in the `if (op == WRITE_OP_ALL || op == WRITE_OP_BASIC_ONLY)` block, a few lines above

uma-sethuraman:
@francescolavra we updated the test based on the changes requested. Is there anything else we should change?

PR #1355, opened 2020-12-10
Zero out rest of page after bss during elf load

sanderssj:
The initial_len calculation was not zeroing out the remainder of the page
after the bss size. The ld-linux loader uses that space and expects it to be 
zero. The fix and elf debug code pulled from @wjhun's arm branch.

### DIFF
--- a/src/kernel/elf.c
+++ b/src/kernel/elf.c
@@ -2,6 +2,13 @@
 #include <elf64.h>
 #include <page.h>
 
+//#define ELF_DEBUG
+#ifdef ELF_DEBUG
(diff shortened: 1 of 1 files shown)


wjhun (approved):
The case covered here is when the data section ends in a middle of a page, yet bss is small enough to still not extend to the end of the page. elf/dl-minimal.c:malloc() in glibc will begin allocating from this unused portion after bss before turning to mmap. Assuming that such area is always zeroed, the accompanying calloc() doesn't actually zero memory. Nanos wasn't zeroing this area either, and thus calloc() in the loader was sometimes returning uncleared memory.

PR #1358, opened 2020-12-17
vm_exit(): add support for platform-specific VM halt function

francescolavra:
If the platform on which the kernel is running has a defined procedure to stop an instance, this procedure should be used not only when the instance is stopped from the cloud platform but also when the program finishes executing (unless the reboot_on_exit
manifest flag is set).
This PR adds a vm_halt global closure that can be set by platform-specific code and is called by vm_exit().

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -340,6 +340,8 @@ static void reclaim_regions(void)
     }
 }
 
+halt_handler vm_halt;
+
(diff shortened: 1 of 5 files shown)


sanderssj (commented):
Looks good to me!

wjhun (approved):
Looks good

PR #1359, opened 2020-12-17
TLB shootdowns

sanderssj:
This PR implements TLB shootdowns. The page invalidate interface has been modified
so that page invalidates are batched up using a new tracking structure called a
flush_entry. A finite number of flush_entry structs are created at startup so they can be
allocated without going through any heap allocators. The existing calls to page_invalidate
now take a pointer to a flush_entry, which is passed in to the page walkers. When all the
addresses have been added, a call to page_invalidate_sync is made which assigns a
generation number to the flush_entry and adds it to a central linked list. It then fires off
the flush ipi and processes the flush_entry list on the calling cpu. If there are more than
a certain number of addresses to invalidate then a full tlb flush (cr3 reload) is performed.
A refcount on the flush_entry tracks cpu progress, and when the last cpu processes the
flush_entry then a service thunk is queued to runqueue. This service routine processes
the flush_entry list, removing completed entries and running their completion thunks. It
also processes any completion thunks stored on a separate queue from flush_entries
that had no pages to invalidate.

There is also a new simple read-write spinlock implementation to protect the flush_entry
list and allow cpus to process the list concurrently inside of the flush interrupt. It builds
on the existing spinlock data and functions.

### DIFF
--- a/src/x86_64/flush.c
+++ b/src/x86_64/flush.c
@@ -2,51 +2,62 @@
 #include <page.h>
 #include <apic.h>
 
-// in order to keep SMP_TEST conditionalized
 static boolean initialized = false;
-static queue flush_queue;
 static int flush_ipi;
 static heap flush_heap;
 
-typedef struct flush_entry {
-    struct refcount r;
-    u64 page; // or INVALID_ADDRESS for a full flush
-} *flush_entry;
 
 static void invalidate (u64 page)
 {
-    asm volatile("invlpg (%0)" :: "r" (page) : "memory");            
+    asm volatile("invlpg (%0)" :: "r" (page) : "memory");
+}
+
+static void _flush_handler(void)
+{
+    cpuinfo ci = current_cpu();
+    void *p;
+    while ((p = dequeue(ci->inval_queue)) != INVALID_ADDRESS) {
+        invalidate((u64)p);
+    }
 }
 
 closure_function(0, 0, void, flush_handler)
 {
-    flush_entry f = queue_peek(flush_queue);
-    if (f->page == INVALID_PHYSICAL) {
-        flush_tlb();
-    } else {
-        invalidate(f->page);
-    }
-    if (refcount_release(&f->r))
-        deallocate(flush_heap, dequeue(flush_queue), sizeof(struct flush_entry));
+    _flush_handler();
 }
 
-void page_invalidate(u64 p, thunk completion)
+
+void page_invalidate(u64 p)
 {
     if (initialized) {
-        flush_entry f = allocate(flush_heap, sizeof(struct flush_entry));
-        init_refcount(&f->r, total_processors, completion);
-        enqueue(flush_queue, f);
-        // we can choose to delay/amortize this
-        apic_ipi(TARGET_EXCLUSIVE_BROADCAST, 0, flush_ipi);        
+        for (int i = 0; i < total_processors; i++) {
+            while (!enqueue(cpuinfos[i].inval_queue, (void *)p))
+                page_invalidate_sync(ignore);
+        }
     } else {
         invalidate(p);
-        apply(completion);
     }
 }
+
+void page_invalidate_sync(thunk completion)
+{
+    if (initialized) {
+        u64 flags = irq_disable_save();
+        int id = current_cpu()->id;
+        for (int i = 0; i < total_processors; i++) {
+            if (i != id)
+                apic_ipi(i, 0, flush_ipi);
+        }
+        _flush_handler();
+        irq_restore(flags);
+    }
+
+    if (completion)
+        apply(completion);
+    }
 
 void init_flush(heap h)
 {
-    flush_queue = allocate_queue(h, 128);
     flush_ipi = allocate_interrupt();
     register_interrupt(flush_ipi, closure(h, flush_handler), "flush ipi");
     flush_heap = h; // xxx - not really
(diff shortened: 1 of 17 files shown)


francescolavra (commented):
Looks good to me

src/x86_64/flush.c
```diff
@@ -2,54 +2,226 @@
...
+    while (ci->inval_gen != inval_gen) {
+        word oldgen = ci->inval_gen;
+        ci->inval_gen = inval_gen;
+        list_foreach(&entries, l) {
+            flush_entry f = struct_from_list(l, flush_entry, l);
+            if (f->gen <= oldgen)
+                continue;
+            if (f->gen > ci->inval_gen)
+                break;
+            if ((f->cpu_mask & U64_FROM_BIT(ci->id)) == 0)
```
wjhun, line 69: I'm curious why it's necessary to cpu_mask here, or why keeping the mask is necessary (aside from debugging and assertion). I thought the refcount takes care of accounting, and if f->gen > oldgen then the page could not have already been invalidated on this cpu...not sure if I'm missing something here.

src/x86_64/flush.c
```diff
@@ -2,54 +2,226 @@
...
+
+/* must be called with interrupts off */
+static void _flush_handler(void)
 {
-    flush_entry f = queue_peek(flush_queue);
-    if (f->page == INVALID_PHYSICAL) {
+    cpuinfo ci = current_cpu();
+    /* Each generation has at least one page, so if the gen difference is
+     * greater than FLUSH_THRESHOLD, just do a full tlb flush */
+    boolean full_flush = inval_gen - ci->inval_gen > FLUSH_THRESHOLD;
```
wjhun, line 57: Makes sense. It would be cool if there was a way to propagate full_flush backwards as well as forwards. So, for instance, when full_flush is set here, or later in the list traversal when f->flush is true, that would cause all previous entries in the list to be skipped and replaced with a flush for all other cpus, so they don't waste cycles invalidating pages when a flush is imminent. Two approaches come to mind: setting flush for all previous entries or keeping a global minimum generation number which is advanced to inval_gen whenever a full flush condition is met. Here, the full_flush assignment here would also be true if ci->inval_gen is less than this global minimum.

wjhun (approved):
Everything here looks great...my comments are nits, optimizations or general observations - I didn't see any major issues.

I like how the entries list works with the per-cpu generations, deferred cleanup, refcount completions, etc. Seems very elegant!

src/x86_64/flush.c
```diff
@@ -2,54 +2,226 @@
...
-        invalidate(p);
-        apply(completion);
+        if (completion)
+            apply(completion);
+    }
+}
+
+flush_entry get_page_flush_entry()
+{
+    if (initialized) {
```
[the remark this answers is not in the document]
sanderssj: The page flush code gets called before stage 3, so it has to have no-op versions, unfortunately.

src/x86_64/flush.c
sanderssj: I did this because I was thinking about a situation where if you knew a cpu wouldn't have the mapping you could avoid setting its bit and thus avoid the extra work on those cpus. However, that's not something we have today so I went ahead and removed this check since it's otherwise redundant.

src/x86_64/flush.c
```diff
@@ -2,54 +2,226 @@
...
+        flush_entry f = struct_from_list(l, flush_entry, l);
+        if (f->ref.c > 0)
+            continue;
+        list_delete(&f->l);
+        entries_count--;
+        if (trydefer) {
+            if (!enqueue(flush_completion_queue, f->completion))
+                apply(f->completion);
+        } else
+            apply(f->completion);
```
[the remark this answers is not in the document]
sanderssj: That's a great point, and definitely worthy of a comment. Added.

src/x86_64/flush.c
sanderssj: This is a great idea! I considered both ways you suggested, but decided to set flush on all the previous entries in the list because it happens in a single place with the flush_lock held. Keeping track of a full flush generation number would be happening with multiple cores concurrently traversing the list and updating the gen number. I think it would be too racy and full of cache thrashing.

src/x86_64/flush.c
```diff
@@ -2,54 +2,226 @@
...
+     * greater than FLUSH_THRESHOLD, just do a full tlb flush */
+    boolean full_flush = inval_gen - ci->inval_gen > FLUSH_THRESHOLD;
+
+    spin_rlock(&flush_lock);
+    while (ci->inval_gen != inval_gen) {
+        word oldgen = ci->inval_gen;
+        ci->inval_gen = inval_gen;
+        list_foreach(&entries, l) {
+            flush_entry f = struct_from_list(l, flush_entry, l);
+            if (f->gen <= oldgen)
```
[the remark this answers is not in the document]
sanderssj: I like the idea because I have seen one processor get hundreds of entries behind all the others, so being able to start mid-list would eliminate unnecessary traversal. However, when thinking through this, since the cpu would be tracking a list pointer that will eventually be removed when the list is cleaned up, it requires the cleaner to check every cpu to see if it has that pointer, and do this for every flush entry being removed. That sounds expensive enough that I'm not sure it will actually benefit, so I'm punting this for now. I think it's still worth investigating if we think of a good way to do it.

PR #1360, opened 2020-12-18
Add ability to print out klog contents from disk image

### LINKED ISSUE #1357: dump could take a command that grabs the crashdump from a disk image
you can export disk images from various clouds (like gcloud) - it'd be nice to have a command on dump that extracts the crash to make it easy for others to look at

sanderssj:
This PR adds a new switch to the dump tool to print out the contents of the klog from a raw disk image. Since
the klog size can be configurable, it searches backwards from the end of the log to find the beginning. It will
not print out any strings in the general klog area without a valid klog header magic, however.
Fixes #1357

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -378,7 +378,7 @@ closure_function(1, 1, void, sync_complete,
 
 static void storage_shutdown(int status, status_handler completion)
 {
-    if ((status == 0) || !table_find(get_environment(), sym(RADAR_KEY))) {
+    if (status == 0) {
         storage_sync(completion);
     } else {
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
As of now, a crash log is dumped to disk only if the image has a radar key (because so far this log is only used to send crash reports to radar).
Unless we want to restrict this feature to users that have a radar key, now it might make sense for the kernel to dump the log to disk regardless of the presence of the key.

eyberg:
yeh - i think it's fine to dump it to disk for everyone - sending to radar is an addon

wjhun (approved):
Sorry for the delay - I thought I had already approved this. All looks good.

PR #1361, opened 2020-12-21
Guarantee /proc/self/exe path is absolute

sanderssj:
Glibc has code that uses the symlinked path from /proc/self/exe and
asserts that the path is absolute, i.e. starts with a slash. This patch
inserts a slash in front of the exe path if not already included.

### DIFF
--- a/src/unix/special.c
+++ b/src/unix/special.c
@@ -118,8 +118,14 @@ void register_special_files(process p)
         assert(proc_self);
         value program = table_find(p->process_root, sym(program));
         assert(program);
-        buffer b = clone_buffer(h, program);
+        buffer b;
+        /* glibc expects exe path to be absolute */
+        if (peek_char(program) != '/')
+            b = buffer_cstring(h, "/");
+        else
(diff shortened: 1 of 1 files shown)


src/unix/special.c
```diff
@@ -118,8 +118,14 @@ void register_special_files(process p)
         assert(proc_self);
         value program = table_find(p->process_root, sym(program));
         assert(program);
-        buffer b = clone_buffer(h, program);
+        buffer b;
```
wjhun, line 121: Just a nit, but I would have just changed clone to allocate_buffer of size buffer_length((buffer)program) + 2 (to avoid buffer realloc), then just conditionally buffer_write_byte(b, '/') if program doesn't start with a slash. But really what's there is fine...

wjhun (approved):

src/unix/special.c
sanderssj: Yeah, that's a lot nicer way to do it, thanks!

PR #1364, opened 2020-12-30
Add syscall count summary

sanderssj:
This changeset adds support for outputting an 'strace -c' like list of syscall counts and times on
vm exit. The syscall tracking and summary output are controlled through a new manifest symbol
called 'syscall_summary'. New print formatting support for alignment, padding, and precision has
been added in order to make printing the report output simpler. The change also adds a list to 
track shutdown completions that are then run in kernel_shutdown. The existing fs sync is now added
to that list at shutdown time so it always runs last, and a merge is used to run the halt completion 
after the shutdown completions have finished.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -43,6 +43,7 @@ extern void init_interrupts(kernel_heaps kh);
 
 static struct kernel_heaps heaps;
 static filesystem root_fs;
+vector shutdown_completions;
 
 static heap allocate_tagged_region(kernel_heaps kh, u64 tag)
 {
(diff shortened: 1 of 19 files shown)


src/kernel/stage3.c
```diff
@@ -57,7 +57,7 @@ static buffer bulk_test_buffer(heap h)
 {
     buffer b = allocate_buffer(h, BULK_TEST_BUFSIZ);
     for (int i = 0; i < (BULK_TEST_BUFSIZ / 10); i += 8) {
-        bprintf(b, "%8d %8d %8d %8d %8d %8d %8d %8d\r\n",
+        bprintf(b, "%08d %08d %80d %08d %08d %08d %08d %08d\r\n",
```
francescolavra, line 60: In the format string there is a `%80d` instead of `%08d`.
I also found a few more places where we need to change the flags so that the resulting string is padded with '0' instead of ' ':
- src/unix/mmap.c, do_demand_page(): `0x%16lx` should be `0x%016lx`, or better yet `%p`
- src/xen/xennet.c, xennet_attach(): `%2x:%2x:%2x:%2x:%2x:%2x` should be `%02x:%02x:%02x:%02x:%02x:%02x`
- src/tfs/tfs.c, report_sha256(): `%16lx%16lx%16lx%16lx` should be `%016lx%016lx%016lx%016lx`
- platform/pc/boot/stage2.c, kernel_read_complete(): `%8lx` should be `%08lx`

platform/pc/service.c
```diff
@@ -381,42 +382,68 @@ closure_function(1, 1, void, sync_complete,
...
+static void storage_shutdown(int status, merge m)
 {
     if ((status == 0) || !table_find(get_environment(), sym(RADAR_KEY))) {
-        storage_sync(completion);
+        storage_sync(apply_merge(m));
     } else {
-        merge m = allocate_merge(heap_locked(&heaps), completion);
-        status_handler sh = apply_merge(m);
         klog_save(status, apply_merge(m));
         storage_sync(apply_merge(m));
```
francescolavra, line 391: `storage_sync(apply_merge(m))` is now executed regardless of whether the klog is being saved, so it can be moved out of the if/else block to avoid duplicated code.

src/kernel/stage3.c
sanderssj: Good catch, I did miss all the hex formatted stuff. Thanks!

src/runtime/runtime_init.c
```diff
@@ -50,11 +55,23 @@ static void format_number(buffer dest, struct formatter_state *s, vlist *a)
...
-
-    print_number(dest, x, base, s->width);
+    print_number(tmp, x, base, s->precision);
+    int len = buffer_length(tmp) + sign;
+    if (s->precision == 0 && x == 0)
+        len = 0;
+    if (sign && s->fill == '0')
+        push_u8(dest, '-');
+    if (len < s->width && s->align == 0)
+        for (int i = 0; i < s->width - len; i++) push_u8(dest, s->fill);
```
wjhun, line 68: Nit: duplicate for loops could be made into 'fill' helper function.

wjhun (approved):
Looks good!

platform/pc/service.c
sanderssj: Good point, I cleaned that up, thanks!

src/runtime/runtime_init.c
sanderssj: Good point, I've cleaned that up with a helper function.

PR #1366, opened 2021-01-03
TFS log fixes

francescolavra:
This PR fixes a couple of issues that are preventing creation of images with a large number of files, where the TFS log exceeds the size of one extension.

### DIFF
--- a/tools/mkfs.c
+++ b/tools/mkfs.c
@@ -500,7 +500,7 @@ int main(int argc, char **argv)
     }
     if (!img_size)
         img_size = current_size;
-    img_size = pad(img_size, TFS_LOG_DEFAULT_EXTENSION_SIZE);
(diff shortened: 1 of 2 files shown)


tools/mkfs.c
```diff
@@ -500,7 +500,9 @@ int main(int argc, char **argv)
     }
     if (!img_size)
         img_size = current_size;
-    img_size = pad(img_size, TFS_LOG_DEFAULT_EXTENSION_SIZE);
+    if ((img_size - offset) & (TFS_LOG_DEFAULT_EXTENSION_SIZE - 1))
```
wjhun, line 503: Maybe simpler to do "img_size = pad(img_size - offset, TFS_LOG_DEFAULT_EXTENSION_SIZE) + offset;" here?

wjhun (approved):

tools/mkfs.c
francescolavra: Your suggestion indeed makes the code simpler, thanks.

PR #1371, opened 2021-01-05
VDSO: getcpu(): avoid calling current_cpu()

### LINKED ISSUE #1368: java 11 pkg broke?
@francescolavra said it was segfaulting

francescolavra:
current_cpu() cannot be called from userspace because:
1)  it accesses the GS register, whose contents in user mode are usually different from the contents in kernel mode
2) it is supposed to return a pointer to kernel data (cpuinfo) that is not mapped for userspace access

This commit fixes the VDSO getcpu() implementation by storing the CPU ID in the TSC_AUX MSR, and retrieving it via the RDTSCP instruction.

For future reference, below is a C code snippet that exercises the VDSO getcpu() implementation:
```
extern void *_dl_vdso_vsym(const char *n, void *);
int (*__vdso_getcpu)(unsigned *cpu, unsigned *node, void *tcache) =
    _dl_vdso_vsym("__vdso_getcpu", NULL);
unsigned cpu;
__vdso_getcpu(&cpu, NULL, NULL);
```
Closes #1368.

### DIFF
--- a/src/kernel/kernel.h
+++ b/src/kernel/kernel.h
@@ -84,12 +84,7 @@ static inline __attribute__((always_inline)) cpuinfo cpuinfo_from_id(int cpu)
     return &cpuinfos[cpu];
 }
 
-static inline __attribute__((always_inline)) void cpu_setgs(int cpu)
-{
-    u64 addr = u64_from_pointer(cpuinfo_from_id(cpu));
-    write_msr(KERNEL_GS_MSR, 0); /* clear user GS */
-    write_msr(GS_MSR, addr);
-}
+void cpu_init(int cpu);
 
 static inline __attribute__((always_inline)) cpuinfo current_cpu(void)
 {
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -485,6 +485,7 @@ static void __attribute__((noinline)) init_service_new_stack()
     shutdown_completions = allocate_vector(heap_general(kh), SHUTDOWN_COMPLETIONS_SIZE);
     init_debug("pci_discover (for VGA)");
     pci_discover(); // early PCI discover to configure VGA console
+    init_clock();
     init_kernel_contexts(backed);
 
     /* interrupts */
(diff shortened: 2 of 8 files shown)


sanderssj:
This was my error with the new getcpu. Change looks good to me!

src/kernel/kernel.h
```diff
@@ -89,6 +89,7 @@ static inline __attribute__((always_inline)) void cpu_setgs(int cpu)
     u64 addr = u64_from_pointer(cpuinfo_from_id(cpu));
     write_msr(KERNEL_GS_MSR, 0); /* clear user GS */
     write_msr(GS_MSR, addr);
+    write_msr(TSC_AUX_MSR, cpu);    /* used by vdso_getcpu() */
```
wjhun, line 92: I'm not sure if a write to a non-present MSR would cause a fault, but it might be safer to first check for presence of rdtscp / IA32_TSC_AUX.

From intel SDM, vol 4, 2-46 on IA32_TSC_AUX:

"If CPUID.80000001H:EDX[27] = 1 or CPUID.(EAX=7,ECX=0):ECX [bit 22] = 1"

and in Vol 2A, 3-214, for cpuid EAX=80000001H:

"[EDX] Bit 27: RDTSCP and IA32_TSC_AUX are available if 1."

wjhun (approved):
Looks good!

src/kernel/kernel.h
francescolavra: OK, I made the MSR write conditional to __vdso_dat->platform_has_rdtscp.
The cpu_setgs() function is being replaced by a more generic cpu_init() function (moved from kernel.h to arch-specific code, which also solves issues due to double #includes of vdso.h), where TSC_AUX is initialized (if RDTSCP is supported).
Since the platform_has_rdtscp flag is initialized in init_clock(), I moved the init_clock() call to before the call to init_kernel_contexts() (which indirectly calls cpu_init() for CPU 0).

PR #1372, opened 2021-01-07
Network socket fixes

francescolavra:
The second commit fixes handling of socket connections closed by the remote peer (both if a syscall is ongoing when the connection is closed, and if a syscall is invoked after connection closure).
The first commit fixes a memory corruption bug that surfaced when running the runtime test modified by the second commit.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1675,7 +1675,6 @@ static sysreturn netsock_listen(struct sock *sock, int backlog)
     set_lwip_error(s, ERR_OK);
     tcp_arg(lw, s);
     tcp_accept(lw, accept_tcp_from_lwip);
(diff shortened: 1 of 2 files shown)


sanderssj:
Looks good to me! One question: should we be proactively setting the TCP_SOCK_UNDEFINED state in netsock_shutdown in the case where the pcb gets released?

francescolavra:
> Looks good to me! One question: should we be proactively setting the TCP_SOCK_UNDEFINED state in netsock_shutdown in the case where the pcb gets released?

Yes, good point, in fact without setting the TCP_SOCK_UNDEFINED state in netsock_shutdown, a read() call after a shutdown(SHUT_RDWR) crashes the kernel. I fixed that and added this test case as well to the runtime test.

wjhun (approved):
LGTM

PR #1373, opened 2021-01-07
Convert process threads vector to rbtree

sanderssj:
Previously, a process keeps a list of thread pointers in a vector, indexed
by thread id. This means that as each new thread is created and the id
gets incremented, this vector grows unbounded, even if threads are
deleted, which eventually leads to a failed memory allocation to
grow the vector.
This change puts thread pointers in an rbtree instead, with thread id as
the key. While this makes simple iteration over the thread list more
complicated, it still allows the tid to be incremented for new threads,
avoiding problems with tid reuse and old thread references in the user
program while maintaining a fast lookup by id.

### DIFF
--- a/src/unix/signal.c
+++ b/src/unix/signal.c
@@ -302,6 +302,38 @@ void deliver_signal_to_thread(thread t, struct siginfo *info)
         sig_debug("failed to interrupt\n");
 }
 
+closure_function(2, 1, boolean, deliver_signal_handler,
+                 thread *, can_wake, u64, sigword,
+                 rbnode, n)
+{
+    u64 sigword = bound(sigword);
+    thread *can_wake = bound(can_wake);
+    thread t = struct_from_field(n, thread, n);
+    if (thread_is_runnable(t)) {
+        /* Note that we're only considering unmasked signals
+            (handlers) here, nothing in the interest masks. The
+            thread is runnable, so it can't be blocked on
+            rt_sigtimedwait or a signalfd (poll wait or blocking
+            read). */
+        if ((sigword & sigstate_get_mask(&t->signals)) == 0) {
+            /* thread scheduled to run or running; no explicit wakeup */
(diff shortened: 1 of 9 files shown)


francescolavra (commented):
LGTM

src/unix/signal.c
```diff
@@ -318,7 +318,11 @@ void deliver_signal_to_process(process p, struct siginfo *info)
 
     /* If a thread is set as runnable and can handle this signal, just return. */
     thread t, can_wake = 0;
-    vector_foreach(p->threads, t) {
+    vector v = allocate_vector(heap_general(get_kernel_heaps()), 8);
+    spin_lock(&p->threads_lock);
+    threads_to_vector(p, v);
+    spin_unlock(&p->threads_lock);
+    vector_foreach(v, t) {
```
wjhun, line 325: I suppose the transfer to vector could have performance implications if an instance contains a lot of threads and regularly accepts signals delivered at the process level (e.g. timer_create / SIGEV_SIGNAL).

Could there be a way to embed this thread search in an rbtree_traverse directly? Most of this is just checking flags and should be safe to do while threads_lock is held. The call to signalfd_dispatch could move outside the loop. (I was trying to recall why there wasn't a break after can_wake is assigned - I think it's to keep looking for a thread that is already runnable.)

wjhun (approved):
Looks good; I just made a minor comment about a possible performance impact in process-level signal delivery.

src/unix/signal.c
sanderssj: I moved the search into the traversal, but kept the signalfd_dispatch out. This changes behavior very slightly in that before, if applying the dispatch call made the thread runnable, it could jump out of the loop instead of continuing the search, and this can't happen now with signalfd_dispatch out of the loop.

I do think the new code makes more sense, avoids an alloc/dealloc and so is more performant, and I don't think the missing short circuit is a big deal compared to that.

PR #1374, opened 2021-01-12
Backed heap: add alloc_map() and dealloc_unmap() functions

francescolavra:
alloc_map() allocates memory from both the virtual and the physical heap, and returns to the caller both the virtual and the physical address; dealloc_unmap() takes as arguments both the virtual and the physical address to deallocate.
These new functions can be used as alternatives to alloc() and dealloc() when the caller wants to use the physical address, and allow avoiding calls to physical_from_virtual() (both when passing a physical address to a device, and when deallocating memory), which show up in ftrace logs as taking a significant amount of time when running heavy I/O workloads.
In addition, the backed heap now implements internal locking, which can be enabled when a heap is created, so that both locking and non-locking versions can be used.

In the virtIO code, vqmsg_push() now takes a physical address (instead of a virtual address) as argument; this allows avoiding calls to physical_from_virtual() when the vqmsg_push() caller already knows the physical address to be used in a virtqueue message. The SCSI, storage and network virtIO drivers now use the new alloc_map() and dealloc_map() functions of
the backed heap, so they can pass to vqmsg_push() the physical addresses being allocated.

### DIFF
--- a/src/kernel/backed_heap.c
+++ b/src/kernel/backed_heap.c
@@ -1,66 +1,119 @@
 #include <kernel.h>
 #include <page.h>
 
-typedef struct backed {
-    struct heap h;
-    heap physical;
-    heap virtual;
-} *backed;
+#define backed_heap_lock(bh)    u64 _flags = spin_lock_irq(&(bh)->lock)
+#define backed_heap_unlock(bh)  spin_unlock_irq(&(bh)->lock, _flags)
 
-void physically_backed_dealloc_virtual(heap h, u64 x, bytes length)
+void physically_backed_dealloc_virtual(backed_heap bh, u64 x, bytes length)
 {
-    backed b = (backed)h;       /* XXX need to keep track of heap type... */
-    u64 padlen = pad(length, h->pagesize);
-    if ((x & (h->pagesize-1))) {
+    u64 padlen = pad(length, bh->h.pagesize);
+    if (x & (bh->h.pagesize - 1)) {
 	msg_err("attempt to free unaligned area at %lx, length %x; leaking\n", x, length);
 	return;
     }
 
     unmap(x, padlen);
-    deallocate(b->virtual, pointer_from_u64(x), padlen);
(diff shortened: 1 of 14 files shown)


src/kernel/backed_heap.c
```diff
@@ -38,7 +34,7 @@ static void physically_backed_dealloc(heap h, u64 x, bytes length)
 
 static u64 physically_backed_alloc(heap h, bytes length)
```
wjhun, line 35: There's some redundancy here, and it seems like physically_backed_alloc and physically_backed_dealloc should be replaced with wrappers that call backed_alloc_map and backed_dealloc_unmap, respectively. If the backed_* functions are made inline and only conditionally set *phys if phys == 0, then the compiler should be able to factor out this conditional. The inline shouldn't affect the vft assignment, so these functions will be inlined within this file and also have incarnate forms that are referenced by the vft.

src/kernel/backed_heap.c
```diff
@@ -52,15 +48,87 @@ static u64 physically_backed_alloc(heap h, bytes length)
...
+            map(u64_from_pointer(virt), *phys, len, PAGE_WRITABLE | PAGE_NO_EXEC);
+        else
+            deallocate_u64(bh->physical, *phys, len);
+    } else {
+        virt = INVALID_ADDRESS;
+    }
+    return virt;
+}
+
+void backed_dealloc_unmap(backed_heap bh, void *virt, u64 phys, bytes len)
```
wjhun, line 68: static inline, and continuing from previous comment:

```
if (phys == 0) {
    phys = physical_from_virtual(pointer_from_u64(x));
    assert(phys != INVALID_PHYSICAL);
}
```

wjhun (approved):
Looks good...and good riddance to those gratuitous page table lookups.

src/kernel/backed_heap.c
francescolavra: OK, I made the the backed_* functions static inline, and they are now called by the existing alloc/dealloc functions, so there is no code duplication.

PR #1379, opened 2021-01-15
Fix missing uuid and label during log_extend_link

sanderssj:
The first log_ext in the log stores the uuid and label of the fs.
Currently these fields are pushed on the header in log_create, but
this means that when the log_ext is rewritten as in an fs rebuild, these
fields are then missing which are not detected until reboot. This change
moves adding these fields to the header into log_extend_link which makes
sure they are pushed every time, as log_extend_link is always called
on creation as well as when the log needs to be rewritten.

### DIFF
--- a/src/tfs/tlog.c
+++ b/src/tfs/tlog.c
@@ -306,6 +306,11 @@ static void log_extension_init(log_ext ext)
     assert(push_buffer(ext->staging, alloca_wrap_buffer(tfs_magic, TFS_MAGIC_BYTES)));
     push_varint(ext->staging, TFS_VERSION);
     push_varint(ext->staging, range_span(ext->sectors));
+    if (ext->sectors.start == 0) {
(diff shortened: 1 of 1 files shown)


francescolavra:
It seems to me that a more natural place to push these fields on the extension header would be in log_extension_init(). (This would require the `log_extension_init(init_ext)` call in log_create() to be moved after `random_buffer(uuid)`.)

sanderssj:
> It seems to me that a more natural place to push these fields on the extension header would be in log_extension_init(). (This would require the `log_extension_init(init_ext)` call in log_create() to be moved after `random_buffer(uuid)`.)

I'm glad you mentioned that--I had tried moving it there initially and was having issues with it but I think I didn't fully understand the problem yet. log_extension_init is the right place for it.

wjhun (approved):
LGTM

PR #1380, opened 2021-01-15
Fix log flush compaction blocking

sanderssj:
There is a race condition where if log_flush starts a compaction and then the system shuts down, 
the log_flush from the sync operation will return immediately and not let the compaction (and
storage i/o) to finish. This change fixes that problem by checking for compaction in log_flush and
adding the completion to the flush_completions vector, which is now also run after compaction
is finished.
There is an additional small fix related to the situation where shutting down is waiting on i/o to complete
and the cpu enters runloop again and would crash because all threads are invalid.

### DIFF
--- a/src/tfs/tlog.c
+++ b/src/tfs/tlog.c
@@ -417,18 +417,23 @@ static inline boolean log_write_internal(log tl, merge m)
     return true;
 }
 
+static void run_flush_completions(log tl)
+{
+    if (tl->flush_completions) {
+        status_handler sh;
+        vector_foreach(tl->flush_completions, sh)
+            apply(sh, s);
+        vector_clear(tl->flush_completions);
+    }
+}
+
 closure_function(1, 1, void, log_flush_complete,
                  log, tl,
                  status, s)
 {
     /* would need to move these to runqueue if a flush is ever invoked from a tfs op */
     bound(tl)->dirty = false;
-    if (bound(tl)->flush_completions) {
-        status_handler sh;
-        vector_foreach(bound(tl)->flush_completions, sh)
-            apply(sh, s);
-        vector_clear(bound(tl)->flush_completions);
-    }
+    run_flush_completions(bound(tl));
     bound(tl)->flushing = false;
     closure_finish();
 }
@@ -463,6 +468,9 @@ closure_function(2, 1, void, log_switch_complete,
         tlog_debug("  deallocating extension at %R\n", __func__, ext->r);
         deallocate_u64((heap)fs->storage, ext->r.start, range_span(ext->r));
     }
+
+    run_flush_completions(old_tl);
+
     refcount_release(&to_be_destroyed->refcount);
     timm_dealloc(s);
     closure_finish();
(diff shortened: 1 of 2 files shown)


wjhun (approved):
Looks good!

src/tfs/tlog.c
```diff
@@ -471,14 +479,14 @@ closure_function(2, 1, void, log_switch_complete,
...
+    if (!tl->dirty && !tl->compacting) {
         if (completion)
             apply(completion, STATUS_OK);
         return;
     }
     if (completion)
         vector_push(tl->flush_completions, completion);
-    if (tl->flushing)
+    if (tl->flushing || tl->compacting)
         return;
```
francescolavra, line 490: This is what may cause the log's tuple_staging buffer to fill up during a log compaction and consequently log_write_eav() to fail, leading to a NOSPACE filesystem error when doing "heavy" file write operations (#1473).
It seems to me that an ongoing log compaction shouldn't prevent nor delay a log flush.

wjhun: I agree that the check for tl->compacting here shouldn't be necessary, but I don't think that is the culprit, as compacting is set on the original log being compacted, whereas the log that is overflowing is fs->temp_log, which is set after log_write is called from filesystem_log_rebuild. But fs->temp_log->flushing is set during the long duration of the backlogged log write (see https://github.com/nanovms/nanos/issues/1473#issuecomment-840823845), and fs->temp_log->tuple_staging extends past the log extension size, causing log_write_eav to fail.

While the short-term issue is the poor performance of virtio-storage (#1483), the hard failure on a log write can be fixed by removing the tuple_staging size limit in log_write_eav and log_write. These checks were probably left over from an earlier version of tlog that did not allow tuple encodings to span multiple log extensions. Now that they can, I think these hard limits are probably unnecessary.

Removing these checks allows the fs_full test to get further with STORAGE=virtio-blk, however the system runs out of memory as neither the log data nor queued data is making it out to disk due to the storage issue.

francescolavra: The checks in log_write_eav and log_write have been added because otherwise there is no way for the kernel to report out-of-space errors to userspace for filesystem operations that don't involve allocations of new extents (e.g. creating directories or empty files). While those checks can certainly be improved, I still think the staging tuple cannot be allowed to grow indefinitely in these functions.

wjhun: But those checks don't actually have anything to do with an out-of-storage condition and could cause a failure on a transient condition, such as a backlog of storage requests. It makes sense that log_write_eav and log_write should fail on out of storage conditions, though. Right now it seems that a storage allocation failure would result in an error status being passed to a log_flush completion, which could in turn be passed to a filesystem_flush completion or, in some cases like on flush timer or in log_set_dirty, just be silently ignored.

Perhaps such an error should be recorded in the log structure and checked by future calls to log_write / log_write_eav. And some hard limit for tuple_staging can be enforced, too, so that allocations aren't runaway if a storage device becomes critically unresponsive (such as in the current situation with virtio-blk).

francescolavra: Recording a storage allocation failure in the log structure and checking it in future calls to log_write could be an option, with the caveat that such error would only be reported to userspace at a later syscall, whereas the syscall that actually should have returned the error return success instead. So things such as mkdir() or rename() could return success but not be actually recorded to disk.

PR #1382, opened 2021-01-20
Fix counting of sectors to load in stage1 for weird BIOS

sanderssj:
The stage1 code keeps the current count of sectors to load in AX, which
is currently always small and completely stored in the lower byte (AL). The
BIOS extended read sectors 13h AH=42h interrupt should store the return
code in AH. However, some hypervisor BIOS (e.g. Hyper-V) appear to set
the whole word (AX) as return code, clearing the count held in AL, causing
the loading code to return after a single read. This commit moves
the total sector count to BX, which is not used by the BIOS interrupt.

### DIFF
--- a/platform/pc/boot/stage1.s
+++ b/platform/pc/boot/stage1.s
@@ -79,7 +79,7 @@ dap:
 
 readsectors:
         mov cx, [dap.sector_count]
-        mov ax, cx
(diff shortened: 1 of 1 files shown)


wjhun (approved):
Looks good

PR #1383, opened 2021-01-22
stage1: minimize number of INT 13H BIOS calls to load stage2

francescolavra:
For some reason, issuing more than one INT 13H call in stage1 when running on Xen locks up the VM (the BIOS never returns from the second INT 13H call).
This change minimizes the number of INT 13H calls needed to load stage2, by setting the dap.offset field to 0 and adjusting dap.segment to align with the beginning of stage2: in this way, stage1 can issue INT 13H calls with a sector count up to 0x80 even if the destination buffer is not aligned to a multiple of 0x10000, while still satisfying the constraint that the buffer fits in a single segment.
This change allows booting Nanos on AWS t2 instances, on the condition that the stage2 size does not grow beyond 64 kB.

### DIFF
--- a/platform/pc/boot/stage1.s
+++ b/platform/pc/boot/stage1.s
@@ -70,8 +70,8 @@ dap:
         db 0x10
         db 0
         .sector_count dw STAGE2SIZE/sectorsize
-        .offset       dw stage2
-        .segment      dw 0
+        .offset       dw 0
+        .segment      dw (stage2 >> 4)
         .lba          dq 1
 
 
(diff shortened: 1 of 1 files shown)


wjhun (approved):
The solution looks fine, but it is unfortunate that there exists such a limit on Xen. (Though it sounds the like root cause was not found?) This effectively returns us back to having a 64kB limit for stage2 (though mkfs no longer checks for it...so if it is exceeded, the instance will fail silently without warning).

sanderssj:
Just making sure yesterday's change of tracking sector count in BX instead of AX didn't also fix this by chance (which was causing the load loop to abort after one read for some hypervisors)?

francescolavra:
> Though it sounds the like root cause was not found?

Right, I didn't find the root cause. It doesn't seem to be an intrinsic limitation of the BIOS being used (which is the BIOS from the Bochs x86 emulator, http://bochs.sourceforge.net/cgi-bin/lxr/source/bios/), because if I run stage1 with that BIOS under Bochs the issue doesn't occur.

francescolavra:
> Just making sure yesterday's change of tracking sector count in BX instead of AX didn't also fix this by chance (which was causing the load loop to abort after one read for some hypervisors)?

No, that change didn't fix the issue, unfortunately. Under Xen, the loop continues after the first read, but the second read gets stuck somewhere in the INT 13H handler.

PR #1384, opened 2021-01-25
aarch64 support

### LINKED ISSUE #1219: arm64 entry points
Write initial assembly entry points for the arm port, including transfer from bootloader to kernel start and exception vectors. Can be used to test early kernel init prior to page table setup.

wjhun:
This introduces support for the aarch64 architecture (a.k.a. arm64) and the qemu 'virt' machine platform.

Cross-compiling and testing is possible with an aarch64-linux-gnu-* toolchain (which may be overrided by specifying CROSS_COMPILE in make) and qemu-system-aarch64. For instance, to run webg under qemu emulating a virt machine, run:

```make PLATFORM=virt TARGET=webg run-noaccel```

If the target program is a dynamic executable, set NANOS_TARGET_ROOT to a path that contains an arm64 root image (such a root image may be found at https://storage.googleapis.com/testmisc/arm64-target-root.tar.gz). Runtime tests may be run with the following:

```make PLATFORM=virt NANOS_TARGET_ROOT=<target root directory> runtime-tests-noaccel```

While some factoring of architecture and platform-dependent code was merged a while back, this rounds out that work, moving architecture-specific definitions to src/\<arch\> directories, architecture and platform independent code to more common locations like src/kernel, in particular moving the initialization process from platform/pc/service.c to src/kernel/init.c.

Resolves #1219, #1220, #1222

### CODE REVIEW

File: src/aarch64/interrupt.c [hunk 1]
```diff
@@ -0,0 +1,361 @@
...
+        console("\n       far: ");
+        print_u64_with_sym(far);
+    }
+}
+
+void print_frame(context f)
+{
+    u64 v = f[FRAME_VECTOR];
+    console(" interrupt: ");
+    print_u64(v);
```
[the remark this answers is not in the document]
wjhun: Ok. It looks like there are still a lot of calls to console, even in x86_64 code (like interrupt.c), so perhaps those should be changed over too.

What is the purpose of having both console() and rputs()? There could just be one call that does the console_write, and the klog_write can just be a no-op if klog isn't initialized.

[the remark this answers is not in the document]
francescolavra: The remaining console() calls in x86_64/interrupt.c are there because when they happen the VM is about to be shut down via a direct vm_exit() call, so anything that we write to the log isn't going to be flushed to disk anyway.

I created an rputs() function separated from console() because the kernel log seemed to me conceptually different from a console, but I don't have a strong opinion on whether they should be unified.

src/unix/syscall.c
```diff
@@ -1486,16 +1309,20 @@ sysreturn fdatasync(int fd)
...
-    tuple m;
-    int ret = resolve_cstring_follow(0, current->p->cwd, name, &m, 0);
-    if (ret) {
-        return set_syscall_return(current, ret);
+    tuple m = 0;
+    int ret;
+    if (flags & AT_SYMLINK_NOFOLLOW) {
+        ret = resolve_cstring(0, cwd, pathname, &m, 0);
+        if (!ret && is_symlink(m))
+            ret = -ELOOP;
```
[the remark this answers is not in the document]
wjhun: Thanks. It actually turns out that this whole block isn't necessary, as the faccessat syscall does not take the flags argument, but rather glibc checks for the flags and operates on them accordingly. I fixed this and added some unit tests for access and faccessat to the symlink test.

src/unix/vdso.c
```diff
@@ -36,7 +37,7 @@ void init_vdso(process p)
         size = PAGESIZE;
         paddr = physical_from_virtual((void *)&vvar_page);
         assert(paddr != INVALID_PHYSICAL);
-        map(vaddr, paddr, size, PAGE_USER | PAGE_NO_EXEC | PAGE_READONLY);
+        map(vaddr, paddr, size, flags);
```
francescolavra, line 40: This should be mapped with no_exec flags.

src/unix/vdso.c
```diff
@@ -50,7 +51,7 @@ void init_vdso(process p)
 #endif
         if (paddr != INVALID_PHYSICAL) {
             __vdso_dat->pvclock_offset = paddr & PAGEMASK;
-            map(vaddr, paddr & ~PAGEMASK, size, PAGE_USER | PAGE_NO_EXEC | PAGE_READONLY);
+            map(vaddr, paddr & ~PAGEMASK, size, flags);
```
francescolavra, line 54: Same here

src/aarch64/interrupt.c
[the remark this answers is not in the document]
wjhun: That's fine. I left the existing calls to console in place and changed those in src/aarch64 that should be logged to klog.

src/aarch64/kernel_machine.h
```diff
@@ -0,0 +1,375 @@
...
+    /* Default frame and stack installed at kernel entry points (init,
+       syscall) and calls to runloop. +8 */
+    kernel_context kernel_context;
+
+    /* One temporary for syscall enter to use so that we don't need to touch the user stack. +16 */
+    u64 tmp;
+
+    /*** End of fields touched by kernel entries ***/
+
+    u32 id;
```
sanderssj, line 189: Does it make sense to move the rest or some of the rest of this struct to an arch independent struct to remove redundancy and make a distinction between arch and arch-independent fields? It might be a lot of churn so perhaps not worth it.

src/x86_64/kernel_machine.h
```diff
@@ -111,13 +128,214 @@ extern void write_xmsr(u64, u64);
...
+
 #define switch_stack(__s, __target) {                           \
         asm volatile("mov %0, %%rdx": :"r"(__s):"%rdx");        \
         asm volatile("mov %0, %%rax": :"r"(__target));          \
         asm volatile("mov %%rdx, %%rsp"::);                     \
         asm volatile("jmp *%%rax"::);                           \
     }
 
+/* syscall entry */
+
```
sanderssj, line 338: This is just a little thing, but I think this comment is out of place--looks like it should go with init_syscall_handler further up (as that's how it's used in aarch)

src/aarch64/kernel_machine.h
wjhun: I agree that there should be a common struct, so I moved struct cpuinfo to kernel.h and made a 'struct cpuinfo_machine', which is defined in src/\<arch\>/kernel_machine.h, as its first field.

platform/pc/service.c
```diff
@@ -395,3 +422,69 @@ void init_service(u64 rdi, u64 rsi)
...
+
+    /* storage */
+    init_virtio_blk(kh, sa);
+    init_virtio_scsi(kh, sa);
+    init_pvscsi(kh, sa);
+    init_nvme(kh, sa);
+    init_ata_pci(kh, sa);
+
+    /* misc / platform */
+    init_acpi(kh);
```
francescolavra, line 489: init_acpi() should be called regardless of hypervisor detection. Not calling it means not registering the PIIX4 powerdown handler, which is needed in different x86_64 cloud instance types (including AWS Xen instances) to stop an instance when the program terminates.

wjhun: Ok, I made the call to init_acpi() unconditional.

test/runtime/time.c
```diff
@@ -119,7 +120,9 @@ static void test_clock_nanosleep(clockid_t clock_id, unsigned long long nsec)
     if (clock_gettime(clock_id, &end) < 0)
         fail_perror("clock_gettime");
 
-    delta = validate_interval(&start, &end, 1, nsec);
+    /* given the skew between sampling start and starting the timer,
+       we can't really validate the interval - validate that end > req instead */
+    delta = validate_interval(&req, &end, 1, 0);
```
francescolavra, line 125: It seems to me that no matter how big the skew, the difference between `start` and `end` should always be larger than nsec.

wjhun: Agreed; this check was failing because of an unrelated issue, and I mistook the actual duration of the sleep with the interval between start and end. Reverted.

PR #1385, opened 2021-01-25
Kernel clock: add NTP synchronization

francescolavra:
This change adds a new klib that periodically retrieves the time from an NTP server and synchronizes the kernel clock.
Clock synchronization is done gradually (by applying calibration parameters set via clock_adjust()) so as not to create any discontinuity in the wall clock time. The CLOCK_ID_MONOTONIC clock is now different from CLOCK_ID_MONOTONIC_RAW, and is calculated by applying calibration values obtained when comparing the local time with the time retrieved via NTP. Since retrieving the monotonic time is now slightly less efficient because of the calibration step, various pieces of code in the kernel have been converted to use CLOCK_ID_MONOTONIC_RAW instead.
When the local time is adjusted via clock_adjust(), the kernel timer queue is re-ordered so that any changes in timer expiry times resulting by clock calibration parameter changes are taken into account. In addition, the RTC is updated to reflect the correct time.

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -1,0 +1,284 @@
+#include <kernel.h>
+#include <lwip.h>
+
+#define NTP_SERVER_DEFAULT  "pool.ntp.org"
+#define NTP_PORT_DEFAULT    123
+
+#define NTP_EPOCH_DELTA 2208988800ul    /* Number of seconds between 1900 and 1970 */
+
+#define NTP_QUERY_INTERVAL_DEFAULT  (10 * seconds(60))
+#define NTP_QUERY_INTERVAL_MAX      (5120 * seconds(60))
+
+#define NTP_QUERY_ATTEMPTS  8
+
+#define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
+
+struct ntp_ts {
+    u32 seconds;
+    u32 fraction;
+};
+
+struct ntp_packet {
+    u8 mode:3;
+    u8 vn:3;
+    u8 li:2;
+    u8 stratum;
+    u8 poll;
+    u8 precision;
+    u32 root_delay;
+    u32 root_dispersion;
+    u32 reference_id;
+    struct ntp_ts reference_ts;
+    struct ntp_ts originate_ts;
+    struct ntp_ts receive_ts;
+    struct ntp_ts transmit_ts;
+} __attribute((packed));
+
+declare_closure_struct(0, 1, void, ntp_query_func,
+    u64, overruns);
+
+static struct {
+    char server_addr[256];
+    u16 server_port;
+    struct udp_pcb *pcb;
+    timer query_timer;
+    closure_struct(ntp_query_func, query_func);
+    boolean query_ongoing;
+    timestamp query_interval;
+    int query_errors;
+    timestamp last_raw;
+    void (*rprintf)(const char *format, ...);
+    timer (*register_timer)(clock_id id, timestamp val, boolean absolute, timestamp interval,
+            timer_handler n);
+    err_t (*dns_gethostbyname)(const char *hostname, ip_addr_t *addr,
+            dns_found_callback found, void *callback_arg);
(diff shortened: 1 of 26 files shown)


klib/ntp.c
```diff
@@ -0,0 +1,158 @@
...
+{
+    if (ipaddr)
+        ntp_query(ipaddr);
+}
+
+define_closure_function(0, 1, void, ntp_query_func,
+                        u64, overruns)
+{
+    ip_addr_t server_addr;
+    err_t err = ntp.dns_gethostbyname(NTP_SERVER, &server_addr, ntp_dns_cb, 0);
```
wjhun, line 134: Perhaps the ntp server address and port could be configurable through the manifest.

klib/ntp.c
```diff
@@ -0,0 +1,158 @@
...
+{
+    struct pbuf *p = ntp.pbuf_alloc(PBUF_TRANSPORT, sizeof(struct ntp_packet), PBUF_RAM);
+    if (p == 0)
+        return;
+    struct ntp_packet *pkt = p->payload;
+    ntp.runtime_memset(p->payload, 0, sizeof(*pkt));
+    pkt->vn = 3;    /* NTP version number */
+    pkt->mode = 3;  /* client mode */
+    timestamp_to_ntptime(ntp.now(CLOCK_ID_REALTIME), &pkt->transmit_ts);
+    ntp.udp_sendto(ntp.pcb, p, server_addr, NTP_PORT);
```
wjhun, line 79: The result of udp_sendto could be checked here, reporting to the user if the server cannot be contacted. Also, though the period of the update timer is long, it may be wise to either disable it or exponentially back off the period if NTP requests are not being completed after some threshold of attempts.

klib/ntp.c
```diff
@@ -0,0 +1,158 @@
...
+                      const ip_addr_t *addr, u16 port)
+{
+    struct ntp_packet *pkt = p->payload;
+    if (p->len == sizeof(*pkt)) {
+        timestamp wallclock_now = ntp.now(CLOCK_ID_REALTIME);
+        timestamp origin = ntptime_to_timestamp(&pkt->originate_ts);
+        /* round trip delay */
+        timestamp rtd = wallclock_now - origin - ntptime_diff(&pkt->transmit_ts, &pkt->receive_ts);
+        s64 offset = ntptime_to_timestamp(&pkt->transmit_ts) - wallclock_now + rtd / 2;
+        double temp_cal, cal;
```
wjhun, line 93: Could these computations be achieved without floating point? To my knowledge, this would be the only use of fp in the kernel. In the arm port, generation of fp instructions are explicitly disabled as fpsimd registers are only saved on a thread switch - otherwise they would need to be saved on each entry from userspace.

It seems that computations using temp_cal and cal could instead be achieved with scaled integers.

klib/ntp.c
francescolavra: The NTP server address and port are now configurable through the manifest, via entries such as below:
```
    ntp_address:pool.ntp.org
    ntp_port:123
```

klib/ntp.c
francescolavra: I added an exponential backoff of the NTP query interval if NTP requests fail for NTP_QUERY_ATTEMPTS consecutive times (with NTP_QUERY_ATTEMPTS currently set to 8). Also, now errors are printed to the console.

klib/ntp.c
francescolavra: I removed floating point computations and used fixed-point math instead. Calibration values are now 64-bit values with `CLOCK_CALIBR_BITS` fractional bits (`CLOCK_CALIBR_BITS` is currently defined as 32).

francescolavra:
A couple more changes in this updated PR:
- I found a bug in gmtime_r(), and fixed it in https://github.com/nanovms/nanos/pull/1385/commits/6b4c7471e6b957939f15cb11f5421042af273ebf
- I slightly changed the computation of the long-term calibration value in `ntp_input()`, by subtracting half of the round trip delay from raw timestamps (as in `raw -= rtd / 2`), so as to minimize the effect of jitter in network transfer delays

wjhun (approved):
Looks good

PR #1386, opened 2021-01-26
Add pbuf_free after udp sends

sanderssj:
As discussed in #1381, a pbuf_free is required after a pbuf_alloc and udp_send or udp_sendto or else the the pbuf gets leaked. I also verified this with sending enough data to udploop that it would eventually crash nanos without the fix, and likewise does not crash with the fix.


francescolavra (commented):
Looks good

wjhun (approved):

PR #1388, opened 2021-01-27
mcache: serve large allocations from parent heap

wjhun:
The mcache previously only allowed allocations up to 2^max_order in size. With this change, allocations of larger sizes are served directly from the mcache parent heap. Such fallback allocations cannot be deallocated with a size parameter of -1ull (which is allowed for compatibility with vendor code that expects a malloc/free allocator interface). Assertions have been added to these malloc-type wrappers to make sure that the 2^max_order limit is not exceeded.

Now that the mcache can serve arbitrarily large allocations, the max_order for general mcaches in the kernel has been lowered from 1MB to 64kB; allocations larger than 64kB occur rarely in the kernel (only a few on program load), and larger objcache sizes lead to significant waste of memory (objcache meta stored within parent page allocations results in the loss of one object per page).

### DIFF
--- a/klib/mbedtls.c
+++ b/klib/mbedtls.c
@@ -296,9 +296,16 @@ void mbedtls_platform_zeroize( void *buf, size_t len )
 
 void *mbedtls_calloc(size_t n, size_t s)
 {
-    void *p = allocate(tls.h, n * s);
+    /* To maintain the malloc/free interface with mcache, allocations must stay
+       within the range of objcaches and not fall back to parent allocs. */
+    size_t total = n * s;
+    if (total > U64_FROM_BIT(MAX_MCACHE_ORDER)) {
+        tls.rprintf("%s: %ld bytes exceeds max alloc order\n", __func__, total);
+        return 0;
+    }
+    void *p = allocate(tls.h, total);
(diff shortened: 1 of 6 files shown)


sanderssj:
LGTM

klib/mbedtls.c
```diff
@@ -296,6 +296,12 @@ void mbedtls_platform_zeroize( void *buf, size_t len )
 
 void *mbedtls_calloc(size_t n, size_t s)
 {
+    /* To maintain the malloc/free interface with mcache, allocations must stay
+       within the range of objcaches and not fall back to parent allocs. */
+    if (s > U64_FROM_BIT(MAX_MCACHE_ORDER)) {
+        tls.rprintf("%s: size %ld exceeds max alloc order\n", __func__, s);
```
francescolavra, line 302: It should be `n * s` instead of `s` (both in the `if` condition and in the `rprintf` call).
I verified that mbedtls does not exceed the max alloc order when connecting to an HTTPS server (in my test, the maximum amount it allocates at once is 16717 bytes).

wjhun: Ok, I fixed this. Thanks for checking mbedtls allocs.

PR #1390, opened 2021-01-28
Radar: add /api/v1 prefix to server URL endpoints; add imageName and privateIP attributes in JSON messages

francescolavra:
The /api/v1 prefix is needed to match the current Radar server configuration.
The "privateIP" attribute is a string representation of the IP address assigned to the default network interface.
The "imageName" attribute is sent if a "RADAR_IMAGE_NAME" environment variable is present.


wjhun (approved):
LGTM

PR #1393, opened 2021-01-29
Fix SMP startup-related problems

sanderssj:
This change fixes several problems related to SMP startup. The long mode
assembly has been unified between the bootstrap and other processors so
a consistent set of flags is set on cr4 and cr0. This fixes UD2 exceptions
happening because XSAVE was enabled only on the APs. Moving the assembly
to a separate file also necessitated a change to the Makefiles so that
include dependencies were used for assembly files. Also, a workaround
was added to initialize the MXCSR register, since HVM did not always do
this properly for SMP which could result in SIMD exceptions.

### DIFF
--- a/platform/pc/boot/longmode.inc
+++ b/platform/pc/boot/longmode.inc
@@ -1,0 +1,24 @@
+    %define CR4_PAE (1<<5)
+    %define CR4_PGE (1<<7)
+    %define CR4_OSFXSR (1<<9)
+    %define CR4_OSXMMEXCPT (1<<10)
+    %define CR4_OSXSAVE (1<<18)
+
+    %macro PREPARE_LONG_MODE 1
+    mov %1, CR4_PAE | CR4_PGE | CR4_OSFXSR | CR4_OSXMMEXCPT ;| CR4_OSXSAVE
+    mov cr4, %1
+
+    mov ecx, 0xC0000080 ; Read from the EFER MSR.
+    rdmsr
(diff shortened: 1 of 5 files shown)


platform/pc/service.c
```diff
@@ -446,11 +446,19 @@ void __attribute__((noreturn)) kernel_shutdown(int status)
 u64 total_processors = 1;
 
 #ifdef SMP_ENABLE
+/* hvm does not always properly initialize mxcsr register */
+static void init_mxcsr() {
+    u32 m = 0x1f80;
```
wjhun, line 451: I see that this is the reset value right out of the SDM - so maybe it's not necessary to break it down into a union of individual flags - but maybe a define for this constant with a reference to its origin would be in order?

platform/pc/boot/longmode.inc
```diff
@@ -0,0 +1,24 @@
...
+
+    mov ecx, 0xC0000080 ; Read from the EFER MSR.
+    rdmsr
+    or %1, 0x00000900  ; Set the LME bit and nxe
+    wrmsr
+    %endmacro
+
+    %macro ENTER_LONG_MODE 1
+    mov %1, cr0        ; Activate long mode -
+    or %1, 0x80000001  ; - by enabling paging and protection simultaneously.
```
wjhun, line 19: I know it was already like this, but since this is being rehashed, maybe defines for these constants too?

wjhun (approved):
LGTM

PR #1394, opened 2021-02-03
NTP: add dynamic adjustment and configurability of poll interval

francescolavra:
The first commit fixes a bug that showed up when executing runtime tests with NTP enabled.
The second commit changes the NTP klib so that the poll interval is adjusted dynamically between a minimum and a maximum value, based on the measured jitter of offset values between the local time and the NTP time, using an algorithm similar to that of the NTP reference implementation, whose purpose is to maximize clock precision while minimizing network traffic overhead.
The minimum and maximum poll interval values are expressed as bit order of the number of seconds, and are set by default to 6 (i.e. 64 seconds) and 10 (i.e. 1024 seconds) respectively. These values are configurable via the "ntp_poll_min" and "ntp_poll_max" manifest options, respectively, whose allowed values range from 4 (i.e. 16 seconds) to 17 (i.e. 36.4 hours).

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -6,10 +6,12 @@
 
 #define NTP_EPOCH_DELTA 2208988800ul    /* Number of seconds between 1900 and 1970 */
 
-#define NTP_QUERY_INTERVAL_DEFAULT  (10 * seconds(60))
-#define NTP_QUERY_INTERVAL_MAX      (5120 * seconds(60))
+#define NTP_QUERY_INTERVAL_MIN  4   /* 16 seconds */
+#define NTP_QUERY_INTERVAL_MAX  17  /* 36.4 hours */
 
 #define NTP_QUERY_ATTEMPTS  8
+
+#define NTP_JIGGLE_THRESHOLD    30
 
 #define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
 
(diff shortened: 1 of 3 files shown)


klib/ntp.c
```diff
@@ -174,18 +184,34 @@ static void ntp_input(void *z, struct udp_pcb *pcb, struct pbuf *p,
...
             cal = -cal;
+        s64 jitter = offset - ntp.last_offset;
+        if (ABS(offset) > ABS(jitter) * 4) {
+            ntp.jiggle_counter += ntp.query_interval;
+            if ((ntp.jiggle_counter > NTP_JIGGLE_THRESHOLD) && (ntp.query_interval < ntp.pollmax)) {
+                ntp.query_interval++;
+                ntp.jiggle_counter = 0;
+            }
+        } else {
+            ntp.jiggle_counter -= 2 * ntp.query_interval;
```
wjhun, line 204: Out of curiosity, why is the query_interval factor doubled here but not in the previous block?

wjhun (approved):

klib/ntp.c
francescolavra: The rationale here would be that the NTP client show react faster (decreasing the poll interval) when the offset increases, whereas it could react more slowly when things are improving. It's what the code skeleton in the NTP RFC does (https://tools.ietf.org/html/rfc5905, page 102).
By the way, I realized there was a mistake, the `(ABS(offset) > ABS(jitter) * 4)` condition should be `(ABS(offset) < ABS(jitter) * 4)` instead (I was misled by https://www.eecis.udel.edu/~mills/ntp/html/poll.html, which contains a mistake). I corrected this now and force-pushed.

PR #1399, opened 2021-02-06
pc: init_service(): set CR4 and CR0 when booting without stage2

francescolavra:
When the kernel ELF file is loaded directly by the hypervisor in long mode, the boot process does not go through stage1 and stage2. Thus, appropriate settings for the CR4 and CR0 registers of the bootstrap processor must be enforced in the kernel code. This change moves initialization of CR4 and CR0 from the PREPARE_LONG_MODE and ENTER_LONG_MODE assembly macros to the cpu_init() function. Now the assembly macros only do what is strictly necessary to enter long mode.
This fixes UD2 exceptions being triggered when running under AWS Firecracker.

### DIFF
--- a/platform/pc/boot/longmode.inc
+++ b/platform/pc/boot/longmode.inc
@@ -14,7 +14,7 @@
     %define EFER_NXE (1<<11)
 
     %macro PREPARE_LONG_MODE 1
-    mov %1, CR4_PAE | CR4_PGE | CR4_OSFXSR | CR4_OSXMMEXCPT ;| CR4_OSXSAVE
+    mov %1, CR4_PAE
     mov cr4, %1
 
     mov ecx, MSR_EFER ; Read from the EFER MSR.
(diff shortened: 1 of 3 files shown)


sanderssj:
I think it would be nice to change stage1/2 to set up only the flags that are set by other loaders like Firecracker in the PREPARE_LONG_MODE and ENTER_LONG_MODE macros, which might be only PE/PG in cr0 and LME in EFER (have to doublecheck)? And move everything else into a separate function that can be called by init_service and new_cpu (for SMP) so that we can avoid having flags set in two different places.

francescolavra:
In the updated code I moved CR4 and CR0 initialization to the cpu_init() function, keeping in the PREPARE_LONG_MODE and ENTER_LONG_MODE assembly macros only what is strictly necessary to enter long mode.

wjhun (approved):
Looks good

PR #1402, opened 2021-02-11
Fix smp bugs

sanderssj:
This PR fixes a couple of SMP related bugs. First, it deals with user threads running on other processors during exit by setting the shutting_down boolean in exit_group before destroying threads to prevent runloop from doing user thread work, and it adds checks to syscall to jump immediately back to runloop when shutting down or with an invalid frame.
Second, it moves init_syscall_handler out of init_unix (where it might not run on cpu 0) and ap_init_stack and into cpu_init where it will always run on the bootstrap processor and the APs.
Finally, it adds a check to schedule_frame so that invalid frames (as might result from thread exit) will not crash by failing the assert.

### DIFF
--- a/platform/pc/boot/longmode.inc
+++ b/platform/pc/boot/longmode.inc
@@ -14,7 +14,7 @@
     %define EFER_NXE (1<<11)
 
     %macro PREPARE_LONG_MODE 1
-    mov %1, CR4_PAE | CR4_PGE | CR4_OSFXSR | CR4_OSXMMEXCPT ;| CR4_OSXSAVE
+    mov %1, CR4_PAE
     mov cr4, %1
 
     mov ecx, MSR_EFER ; Read from the EFER MSR.
@@ -26,8 +26,6 @@
     %macro ENTER_LONG_MODE 1
     mov %1, cr0        ; Activate long mode -
     or %1, CR0_PG | CR0_PE  ; - by enabling paging and protection simultaneously.
-    and %1, ~CR0_EM       ; clear EM
-    or %1, CR0_MP         ; set MP
     mov cr0, %1
     %endmacro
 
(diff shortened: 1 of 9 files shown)


wjhun (commented):
I think we may want to revisit the way exit_group works. The various checks for invalid_frame seem to be trying catch a case that really shouldn't happen anyway - which is that a thread's frames and other resources are being released while the thread may still be running - or about to run - somewhere. Perhaps it would be better to just omit the calls to thread_exit from exit_group and just leak thread resources, knowing full well that the VM is about to go away anyhow.

I think we sort of need a shutdown IPI that will forcefully interrupt running threads on all cores. We actually have a shutdown_vector that is used for terminal faults in common_handler - but that forces other CPUs into a halt, which isn't really a good idea if cpu 0 is halted and unable to honor interrupts for storage / klog flushing. So maybe that IPI could be repurposed for the "normal" shutdown case with exit_group (as well as on fault) to suspend thread execution but still allow cores to handle device interrupts.

If such a hard stop of user thread execution is implemented, then I don't think the checks for shutting_down in syscall_debug / syscall_schedule would be necessary, as presumably the kernel lock is held by the core executing exit_group, and a deferred syscall will just be discarded anyway.

The init_syscall_handler bug is more critical, though, so it may make sense to expedite the merging of that fix and address the shutdown stuff in another PR. But either way is fine by me.

sanderssj:
> The init_syscall_handler bug is more critical, though, so it may make sense to expedite the merging of that fix and address the shutdown stuff in another PR. But either way is fine by me.

I removed the shutdown changeset. I played around with the idea of a 'shutdown user' function that can be called from exit_group that would set shutting_down and send a wakeup ipi to everyone to force them into runloop. I'm still thinking about that, so I'll just leave it out for now to get the syscall fix in.

wjhun (approved):

PR #1404, opened 2021-02-16
Reserve space in linker_script for vdso data

sanderssj:
I discovered that the linker seems to discard the .vvar section where __vdso_vdso_dat lives, causing this symbol to point inside the following section, rodata, so that changes made to the vdso_dat object would overwrite strings. I modified the linker_script to advance the location counter by a page to guarantee the space is reserved.


wjhun (approved):
Looks good

PR #1406, opened 2021-02-17
Discover and map ACPI tables and use MADT for SMP startup

sanderssj:
This PR adds support for finding and mapping in the ACPI tables and then using the MADT to discover APIC information including the number of processors in the system (stored in present_processors) and how they map to LAPIC IDs. Each ACPI table is mapped individually as they may not be laid out in a contiguous fashion. AP startup has also changed to use the information from the MADT so now the APs are started one at a time with a unicast IPI instead of broadcast. There are some other minor fixes such as unmapping the apboot memory (which is at virtual address of 0) after AP startup so that null pointer derefs will page fault properly, updating the deallocate_stack function to take the same pointer as returned by allocate_stack, and fixing one of the wait times in AP startup to match the MP spec.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -212,7 +212,7 @@ static void count_processors()
         init_debug("ACPI reports %d processors", present_processors);
     } else {
         present_processors = 1;
-        init_debug("ACPI MADT not found, default to 1 processor");
+        rprintf("warning: ACPI MADT not found, default to 1 processor");
     }
     /* config override */
     present_processors = MIN(present_processors, MAX_CPUS);
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -164,6 +164,7 @@ void vm_exit(u8 code)
 }
 
 u64 total_processors = 1;
+u64 present_processors = 1;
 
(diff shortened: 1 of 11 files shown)


francescolavra (commented):
Looks good to me

platform/pc/service.c
```diff
@@ -187,16 +188,49 @@ static void new_cpu()
...
+static void count_processors()
+{
+    acpi_madt madt = acpi_get_table(ACPI_SIG_MADT);
+    if (madt) {
+        present_processors = 0;
+        acpi_walk_madt(madt, stack_closure(count_processors_handler));
+        init_debug("ACPI reports %d processors", present_processors);
+    } else {
+        present_processors = 1;
+        init_debug("ACPI MADT not found, default to 1 processor");
```
wjhun, line 215: Should the MADT be available in every installation? If so, maybe this message should be printed unconditionally as a warning. This way we can detect if there's an oddball platform that isn't conforming and might be running single core when multiple are specified.

wjhun (approved):
Looks good!

PR #1407, opened 2021-02-18
Sockets: add netlink support

### LINKED ISSUE #453: netlinks socket support?
Many networking apps out there use something like `GetAllNetworkInterfaces" which usually is implemented using netlink socket. It's a very linux only thing and i am not sure if we should support it.

francescolavra:
Netlink sockets allow userspace to retrieve from the kernel the list of network interfaces and IP addresses, and to be notified of events such as network interfaces being added or removed, link status changes, and IP address changes.

Closes #453.


sanderssj:
Looks good to me!

wjhun (approved):
LGTM

PR #1410, opened 2021-02-24
Stop and prevent user threads from running on exit 

sanderssj:
In SMP, user threads may still be running on cpus other than the one calling
exit_group. A boolean, shutting_down, is used by runloop and interrupt
handlers to know to skip any user thread scheduling steps. This now gets set
in exit_group, followed by a no-op interrupt to send the cpus back into
runloop where they can no longer schedule user threads and will idle while
shutdown completes.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -2058,19 +2058,11 @@ void exit(int code)
 
 sysreturn exit_group(int status)
 {
-    thread t;
-    process p = current->p;
-
-    vector v = allocate_vector(heap_general(get_kernel_heaps()), 8);
-    spin_lock(&p->threads_lock);
-    threads_to_vector(current->p, v);
-    spin_unlock(&p->threads_lock);
-
+    /* Set shutting_down to prevent user threads from being scheduled
+     * and then try to interrupt the other cpus back into runloop
+     * so they will idle while running kernel_shutdown */
     shutting_down = true;
     wakeup_or_interrupt_cpu_all();
-    vector_foreach(v, t)
-        exit_thread(t);
-    deallocate_vector(v);
     kernel_shutdown(status);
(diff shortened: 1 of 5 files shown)


src/unix/syscall.c
```diff
@@ -2066,8 +2066,10 @@ sysreturn exit_group(int status)
     threads_to_vector(current->p, v);
     spin_unlock(&p->threads_lock);
 
+    shutting_down = true;
+    wakeup_or_interrupt_cpu_all();
     vector_foreach(v, t)
-            exit_thread(t);
+        exit_thread(t);
```
wjhun, line 2072: If the sending of wakeup IPIs is asynchronous, could there possibly be a race here where a thread's frame is deallocated and yet it gets scheduled on another core before the IPI is caught? It kind of seems unlikely - but also that there's no guarantee that it couldn't happen.

Is there really a need to call exit_thread at all if the system is shutting down and threads can't get scheduled anyway?

sanderssj: Setting shutting_down before bumping the other cpus to runloop should prevent them from being scheduled before tearing them down, since the other places to get scheduled should be protected by the kernel lock or a check for shutting_down, but thinking about it some more I'm less certain about what happens when another cpu is in the middle of random interrupt.
I was assuming that the teardown and flushing of queues that happens in exit_thread was necessary, but I suppose if the user program is calling exit_group then it has already made sure that all the work it needs done is complete. Eliminating the calls to exit_thread is a much simpler and more reliable way to handle any possible races.

wjhun (approved):
Looks good to me. I agree that a race between the calling of exit_thread and that thread running on another core seems unlikely, but I don't see a downside to skipping the calls to exit_thread.

PR #1411, opened 2021-02-24
fix build under gcc 10

wjhun:
Rather than specify -Wno-address-of-packed-member to gcc, use runtime_memcpy to make safe accesses of possibly unaligned data. Remove the packed attribute for structs where 32 and 64-bit fields are already aligned.

Also, specify -Wno-unknown-warning-option to gcc to avoid warnings about possibly truncated string output.


sanderssj:
Looks good to me!

PR #1412, opened 2021-02-26
kernel_shutdown(): fix stall when called from interrupt context

francescolavra:
If kernel_shutdown() is called from interrupt context, it should exit this context before calling runloop(), otherwise if a shutdown handler needs interrupts to work (e.g. for flushing filesystem data to disk), the handler will never complete because pending interrupts will not be processed.
This change implements interrupt_exit() for both x86_64 and aarch64, and adds a call to interrupt_exit() from kernel_shutdown().

### DIFF
--- a/src/aarch64/kernel_machine.c
+++ b/src/aarch64/kernel_machine.c
@@ -1,4 +1,5 @@
 #include <kernel.h>
+#include <gic.h>
 
 //#define TAG_HEAP_DEBUG
 #ifdef TAG_HEAP_DEBUG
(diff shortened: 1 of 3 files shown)


wjhun (approved):
LGTM

sanderssj:
Just a thought--if this is for the kernel_shutdown in the acpi power button interrupt, why not just queue kernel_shutdown as a thunk to runqueue?

francescolavra:
> Just a thought--if this is for the kernel_shutdown in the acpi power button interrupt, why not just queue kernel_shutdown as a thunk to runqueue?

I thought about that, but kernel_shutdown() is also called from halt(), and halt() is called from a myriad of places, including interrupt handlers, so I think it makes sense to make kernel_shutdown() able to deal with being called in interrupt context.

PR #1413, opened 2021-02-26
Add support for Oracle VM Virtualbox

### LINKED ISSUE #1400: virtual box support
from https://github.com/nanovms/ops/issues/888 we currently don't have virtual box support - i know vbox used to provide their own tap kernel extension which would be nice to have on mac os, although I don't know what the status is for that on big sur, also - vbox can be managed headlessly ala vagrant

regardless I think we are missing at least one driver

francescolavra:
To create a virtual hard disk file (vdi file format) for VirtualBox from a raw image, the following command can be used:
```
vboxmanage convertdd disk.raw disk.vdi
```
When creating a virtual machine from the VirtualBox GUI:
- select "Other" as OS type, and "Other/Unknown (64-bit)" as OS version
- specify at least 256 MB as memory size
- select the vdi file created above as hard disk image

After creating the virtual machine, modify the following settings:
- System -> Acceleration -> Paravirtualization Interface: select "KVM"
- Network -> Adapter 1 -> Advanced -> Adapter Type: select "Paravirtualized Network (virtio-net)"

Tested with VirtualBox version 6.1.18.
Closes #1400.

### DIFF
--- a/src/x86_64/interrupt.c
+++ b/src/x86_64/interrupt.c
@@ -15,6 +15,12 @@
 #define INTERRUPT_VECTOR_START 32 /* end of exceptions; defined by architecture */
 #define MAX_INTERRUPT_VECTORS  256 /* as defined by architecture; we may have less */
 
+typedef struct inthandler {
+    struct list l;
+    thunk t;
+    const char *name;
+} *inthandler;
+
 static const char *interrupt_names[MAX_INTERRUPT_VECTORS] = {
     "Divide by 0",
     "Reserved",
@@ -288,6 +294,7 @@ void common_handler()
 }
 
 static id_heap interrupt_vector_heap;
+static heap int_general;
 
 u64 allocate_interrupt(void)
 {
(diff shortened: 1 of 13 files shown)


src/x86_64/interrupt.c
```diff
@@ -323,6 +329,41 @@ void unregister_interrupt(int vector)
...
+        inthandler h = struct_from_list(l, inthandler, l);
+        int_debug("   invoking handler %s (%F)\n", h->name, h->t);
+        apply(h->t);
+    }
+}
+
+u64 allocate_shirq(void)
+{
+    u64 v = allocate_interrupt();
+    heap h = interrupt_vector_heap->meta;
```
wjhun, line 345: Sort of a nit, but could we just stash the general heap in init_interrupts rather than peek into id heap private fields?

src/x86_64/interrupt.c
```diff
@@ -323,6 +329,41 @@ void unregister_interrupt(int vector)
...
+    register_interrupt(v, t, "shirq");
+    return v;
+}
+
+void register_shirq(int v, thunk t, const char *name)
+{
+    if (!handlers[v])
+        halt("%s: vector %d not allocated\n", __func__, v);
+    list shirq_handlers = closure_member(shirq_handler, handlers[v], handlers);
+    inthandler handler = allocate(interrupt_vector_heap->meta, sizeof(struct inthandler));
```
wjhun, line 360: here too

src/x86_64/interrupt.c
```diff
@@ -323,6 +329,41 @@ void unregister_interrupt(int vector)
...
+    list handlers = allocate(h, sizeof(struct list));
+    assert(handlers != INVALID_ADDRESS);
+    list_init(handlers);
+    thunk t = closure(h, shirq_handler, handlers);
+    assert(t != INVALID_ADDRESS);
+    register_interrupt(v, t, "shirq");
+    return v;
+}
+
+void register_shirq(int v, thunk t, const char *name)
```
wjhun, line 355: This isn't critical, but is there any particular reason to have another interface for shared ints rather than just allow multiple handler registrations per vector as aarch64 does? I guess the latter adds a dereference of the list head on dispatch even for non-shared ints, whereas this involves calling another handler to in turn dispatch shared handlers.

wjhun (approved):
Looks good - just minor comments

src/x86_64/interrupt.c
francescolavra: I wrote the allocate_shirq() and register_shirq() functions before seeing the aarch64 implementation (but later I borrowed the inthandler struct from there), so I thought I would leave it as it is and see if it makes sense to you guys. Yes, the only advantage I see is that it avoids dereferencing the list head for non-shared interrupts. For performance-critical code we should try to avoid shared interrupts at all (if the VM allows some control over interrupt routing); I think the cost of invoking multiple handlers per interrupt and have them check the device status needlessly is much higher than the cost of having the "shirq" handler that dispatches the other handlers.

src/x86_64/interrupt.c
francescolavra: OK, I declared a static heap int_general variable, assigned in init_interrupts.

PR #1415, opened 2021-03-01
exec_elf(): allow changing access protections for program sections

francescolavra:
In Linux it is permissible to call mprotect() on an address mapped from a program's ELF sections. In particular, mprotect() can be used to change a code section to be writable.
The .NET runtime, as of version 5.0.102, relies on this feature, and errors out with `Failed to create CoreCLR, HRESULT: 0x80004005` if unable to change the access protections of a mapping (https://github.com/nanovms/ops-examples/issues/47).
This commit changes exec_elf() to allow read, write and exec access to all the program sections loaded from the ELF file, unless the exec_protection manifest flag is present.
A new test case has been added to the mmap test suite, to check that write access cannot be enabled on the code section of a program when the exec_protection manifest option is present.

### DIFF
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -146,8 +146,8 @@ closure_function(0, 1, void, load_interp_fail,
     halt("read interp failed %v\n", s);
 }
 
-closure_function(2, 4, void, exec_elf_map,
-                 process, p, kernel_heaps, kh,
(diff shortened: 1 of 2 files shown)


sanderssj:
Looks good to me!

wjhun (approved):
LGTM

PR #1418, opened 2021-03-01
Always invalidate pages on pagetable additions

sanderssj:
The old code only invalidated pages if the PAGE_PRESENT flag was set in the
PTE, but if the current cpu hasn't processed the tlb shootdown from
another cpu clearing the present flag, the old incorrect translation could
still be in the TLB. This change unconditionally sets invalidate to guarantee
the current cpu does not have any stale entries in the TLB. A second small
change is to move the page_invalidate_flush to the top of runloop to catch
a cpu coming from an interrupt handler that may not have processed a
shootdown yet.

### DIFF
--- a/src/kernel/schedule.c
+++ b/src/kernel/schedule.c
@@ -183,6 +183,8 @@ NOTRACE void __attribute__((noreturn)) runloop_internal()
                 queue_length(bhqueue), queue_length(runqueue), queue_length(ci->thread_queue),
                 idle_cpu_mask, ci->have_kernel_lock ? " locked" : "");
     ci->state = cpu_kernel;
+    /* Make sure TLB entries are appropriately flushed before doing any work */
+    page_invalidate_flush();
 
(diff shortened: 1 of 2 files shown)


src/kernel/schedule.c
```diff
@@ -247,8 +249,6 @@ NOTRACE void __attribute__((noreturn)) runloop_internal()
                     ci->last_timer_update = here + runloop_timer_max;
                 }
             }
-           /* Make sure TLB entries are appropriately flushed before
-             * returning to userspace */
             page_invalidate_flush();
```
wjhun, line 252: Did you mean to also leave a call to page_invalidate_flush here?

wjhun (approved):
Looks good

src/kernel/schedule.c
sanderssj: Yeah, that's not supposed to be there. Thanks!

PR #1419, opened 2021-03-02
Syscalls: stat(): set st_blocks and st_blksize for regular files

francescolavra:
st_blocks contains the number of 512-byte blocks allocated for a file, while st_blocksize contains the "preferred" block size for efficient filesystem I/O.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -43,6 +43,15 @@ void fsfile_set_length(fsfile f, u64 length)
     pagecache_set_node_length(f->cache_node, length);
 }
 
+u64 fsfile_get_blocks(fsfile f)
+{
+    u64 blocks = 0;
+    rangemap_foreach(f->extentmap, n) {
+        blocks += range_span(n->r);
+    }
+    return blocks;
+}
+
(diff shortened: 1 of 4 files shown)


sanderssj:
Should we be setting some of the other values to avoid weirdness in user apps? For example, st_blocks (which is probably what the st_blksize was dividing into) could be used to determine behavior related to sparse file maybe? The are some other fields we don't fill out but it kind of looks like st_blocks is the only other one that might influence user behavior.

wjhun (approved):

francescolavra:
> Should we be setting some of the other values to avoid weirdness in user apps? For example, st_blocks could be used to determine behavior related to sparse file maybe? The are some other fields we don't fill out but it kind of looks like st_blocks is the only other one that might influence user behavior.

Good point, in fact the use case that prompted this PR (https://github.com/facebook/rocksdb/blob/master/env/io_posix.cc#L1244) makes use of the st_blocks field too. I added a new fsfile_get_blocks() function in tfs.c that returns the number of allocated blocks, so that fill_stat() in syscall.c can set the st_blocks field.

PR #1421, opened 2021-03-03
Add memory debug heap wrapper

sanderssj:
This change adds a heap wrapper to check various conditions or faults for certain behavior. The wrapper
does the following:
* Validates that the passed deallocation size matches the allocation size
* Adds a redzone area around both sides of the allocation with a pattern that is checked on deallocation for buffer overruns
* Writes a pattern over the requested allocation to force fault for uninitialized memory access
* Writes a pattern over the entire allocation to force fault for a use after free.
The latter three items are each optional, configured by three defines in mem_debug.c.

There is a heap and backed_heap version of the wrapper, each taking a padding size argument to define how big the
padding is on either side of the requested allocation, which can be used if the memory needs to be aligned. A zero 
padding size argument on creation just uses the wrapper's default value, 64, or PAGESIZE for the backed heap version.

I also tried the wrapper on a few major systems and this change includes a few fixes from that:
* Saving the completion in pagecache_write_sg_finish to a local since the closure could be deallocated when bound(completion) is called
* Match the dealloc size in virtio_scsi_request_complete to the alloc (it was padded anyway so not technically an error)
* Use allocate_zero to allocate the virtqueue since access to the msgs array depend on that area being initialized.


src/runtime/heap/mem_debug.c
```diff
@@ -0,0 +1,199 @@
...
+    *nb = *padding * 2 + b;
+    if (padsize >= PAGESIZE)
+        *nb = pad(*nb, PAGESIZE);
+}
+
+/* These functions use volatile so the hdr address won't be optimized out when debugging. */
+static u64 alloc_check(volatile mem_debug_hdr hdr, bytes b, bytes padding)
+{
+    #ifdef MEMDBG_OVERRUN
+    set_pattern(hdr, padding, &pat_redzone, sizeof(pat_redzone));
```
wjhun, line 87: Should this be hdr + 1? I see the hdr fields get written afterwards, but wouldn't this cause a mismatch if the check expects the pattern to start at hdr + 1?

src/runtime/heap/mem_debug.c
```diff
@@ -0,0 +1,199 @@
...
+    set_pattern(buf, b, &pat_init, sizeof(pat_init));
+    #endif
+    return u64_from_pointer(buf);
+}
+
+static void dealloc_check(volatile mem_debug_hdr hdr, u64 a, bytes b, bytes nb, bytes padding)
+{
+    assert(hdr->sig == DBG_HDR_SIG);
+    assert(b == hdr->allocsize);
+    #ifdef MEMDBG_OVERRUN
```
wjhun, line 104: Just an aesthetic nit, but in general we haven't been indenting preprocessor directives in the source.

wjhun (approved):
One question about a set_pattern start pointer, but otherwise looks great. This feature is a big win!

src/runtime/heap/mem_debug.c
```diff
@@ -0,0 +1,199 @@
...
+static void mem_debug_backed_dealloc_unmap(backed_heap h, void *v, u64 p, bytes b)
+{
+    mem_debug_backed_heap mdh = (mem_debug_backed_heap)h;
+    bytes padding, nb;
+    u64 a = u64_from_pointer(v);
+
+    get_debug_alloc_size(b, mdh->padsize, &nb, &padding);
+    mem_debug_hdr hdr = (mem_debug_hdr)pointer_from_u64(a - padding);
+    dealloc_check(hdr, a, b, nb, padding);
+    dealloc_unmap(mdh->parent, hdr, 0, nb);
```
francescolavra, line 175: I think this should set `p - padding` as physical address argument (if `p` is not zero), so that dealloc_unmap() doesn't have to call physical_from_virtual().

src/runtime/heap/mem_debug.c
sanderssj: It's writing over the entire padding instead of the non-header part (as the check calls it), and it works out since the header size is a multiple of the padding size. It might not work if you added a small field to the header however. I'll change it to match the check pattern since that's clearer and more robust. Thanks!

src/runtime/heap/mem_debug.c
sanderssj: Good point, thanks!

sanderssj:
> `p` could be 0 (e.g. if this function is called via a `dealloc` callback), so `p - padding` should be used only if `p` is not 0.

Thanks for catching that!

PR #1422, opened 2021-03-04
x86_64: implement TSC calibration using PIT

francescolavra:
This allows the kernel to run on hypervisors that don't expose a known paravirtualization interface and don't implement the HPET.
The HPET is now used only as a last resort, if the calibrated TSC cannot be used.
The "failed to initialize HPET; invalid femtoperiod" error message, which would appear when probing a non-existing HPET, is being removed because it may no longer be a fatal error.

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -462,6 +462,7 @@ void init_service(u64 rdi, u64 rsi)
 }
 
 extern boolean init_hpet(kernel_heaps kh);
+extern boolean init_tsc_timer(kernel_heaps kh);
 
 void detect_hypervisor(kernel_heaps kh)
 {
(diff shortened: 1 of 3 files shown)


sanderssj:
Nice! Now that it's an option I think we should consider preferring the TSC over the HPET in the case of no PV timing interfaces (provided the TSC is reliable) as it vastly outperforms HPET. Perhaps that should be done on a hypervisor case-by-case basis and not here, but I wanted to bring it up just in case.

sanderssj:
Also, LGTM!

platform/pc/service.c
```diff
@@ -470,8 +471,8 @@ void detect_hypervisor(kernel_heaps kh)
         if (!xen_detect(kh)) {
             if (!hyperv_detect(kh)) {
                 init_debug("no hypervisor detected; assuming qemu full emulation");
-                if (!init_hpet(kh)) {
-                    halt("HPET initialization failed; no timer source\n");
+                if (!init_hpet(kh) && !init_tsc_timer(kh)) {
```
wjhun, line 474: As Justin noted, it may be preferable to use a calibrated TSC and default to HPET as a last resort. And, for the sake of debugging, could this report which timer was initialized via init_debug?

wjhun (approved):
LGTM

platform/pc/service.c
francescolavra: OK, now the calibrated TSC takes precedence over the HPET, and INIT_DEBUG shows which timer is being used.

PR #1423, opened 2021-03-05
Unify and abstract tfs sector map operations

sanderssj:
Tfs (and tlog) reserve, allocate, and free from the id heap for the sector
map with a few different operations: allocate_u64/deallocate_u64 and
id_heap_set_area, depending on the situation. This can be a problem if
deallocate_u64 is used on a part of the bitmap set by id_heap_set_area
as they might have different alignment requirements. This change adds
a couple new operations for managing the sector heap to go with
filesystem_reserve_storage for allocation and freeing. The free function
uses id_heap_set_area instead of deallocate to note free sectors so
there are no alignment requirements.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -174,10 +174,24 @@ static inline extent allocate_extent(heap h, range file_blocks, range storage_bl
     return e;
 }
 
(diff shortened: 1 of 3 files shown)


wjhun (approved):
LGTM

PR #1425, opened 2021-03-07
Refactor cpu state assignment locations

sanderssj:
The scheduler was setting the cpu state in run_thunk before calling the
thunk, but this can be a problem if something in the thunk causes a state
change (such as a page fault). This change refactors the state assignment
into the thunks themselves. This means that now the cpu state is set to
user right before the frame return, rather than before performing other
operations such as signal handling. Thunks from the bh and run queue
should be kernel state, which is set by the scheduler upon entry, so the
only thunk update required are the two that will run user code.


wjhun (approved):
Looks good!

PR #1426, opened 2021-03-10
Configure MSI before allocating resources in attach_vtpci

sanderssj:
The allocation of config resources requires knowing whether or not MSI is
supported in order to initialize with the correct offset.

Fixes nanovms/ops#945


wjhun (approved):
Looks good. I addressed this in the gic-v2m branch but didn't think it was affecting anything on the Intel side, otherwise I would have PRed separately. Is this related to the VirtualBox support?

wjhun:
> Is this related to the VirtualBox support?

Please disregard - I missed the description about GCP.

Issue #344, opened 2019-01-08
Need to wait for init_virtio_network intialization


tijoytom:
Looks like http.get fails since there is no valid ip4 on the default adapter.The problem is that we don't wait for lwip callback (status_callback) to complete before en-queuing elf startup. It should be done like init_virtio_storage with a closure passed to status_callback and status_callback apply the closure to enqueue start method.

francescolavra:
Detection and initialization of network interfaces is done synchronously, before starting the application. The DHCP client is also started before the application.
Waiting for an IP address from the DHCP server before starting the application doesn't sound right, the server may or may not be available when a VM is started, and its unavailability or network delays shouldn't prevent the VM from booting or delay the boot process.
Closing.

PR #1429, opened 2021-03-11
Set default GW and NetMask when IP is present

am11:
Implements https://github.com/nanovms/ops/commit/587cbf9ce2ded490d8c1319eb2f91713b9b5e146 in nanos, so VM instances (created
by `ops instance create`) can also use the same defaults as `ops run`.
See https://github.com/nanovms/ops/issues/944#issuecomment-796829571.

### DIFF
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -196,10 +196,21 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
     ip4_addr_t netmask;
     ip4_addr_t gw;
 
-    if (!get_config_addr(root, sym(ipaddr), &ip) ||
-        !get_config_addr(root, sym(gateway), &gw) ||
-        !get_config_addr(root, sym(netmask), &netmask))
+    if (!get_config_addr(root, sym(ipaddr), &ip))
         return false;
+
+    if (!get_config_addr(root, sym(netmask), &netmask))
+        ip4_addr_set_u32(&netmask, lwip_htonl(0xffffff00)); // 255.255.255.0
+
+    if (!get_config_addr(root, sym(gateway), &gw)) {
+        // common best practices are: network + 1 or broadcast - 1,
+        // so we will use latter if former is in use.
+        u32_t ip_after_network = (netmask.addr & ip.addr) + lwip_htonl(1);
+        if (ip_after_network == ip.addr)
+            ip4_addr_set_u32(&gw, (~(netmask.addr) | ip.addr) - lwip_htonl(1));
+        else
+            ip4_addr_set_u32(&gw, ip_after_network);
+    }
 
     if (trace) {
         rprintf("NET: static IP config:\n");


src/net/net.c
```diff
@@ -196,11 +196,16 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
     ip4_addr_t netmask;
     ip4_addr_t gw;
 
-    if (!get_config_addr(root, sym(ipaddr), &ip) ||
-        !get_config_addr(root, sym(gateway), &gw) ||
-        !get_config_addr(root, sym(netmask), &netmask))
+    if (!get_config_addr(root, sym(ipaddr), &ip))
         return false;
 
+    if (!get_config_addr(root, sym(gateway), &gw))
```
wjhun, line 202: I suggest first checking if the netmask is configured, defaulting to /24 if not, and then afterwards choosing the default gateway (if unconfigured) based on the actual netmask. This will allow the gateway default setting to work if a netmask other than /24 is used.

Also, the stored addresses are in network byte order, so you'll want to make use of lwip_ntohl / lwip_htonl.

src/net/net.c
```diff
@@ -196,11 +196,16 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
...
+        // assumes the default gateway is the first or last IP in the network range
+        gw.addr = (ip.addr & ~0xff) | ((ip.addr & 0xff) == 1 ? 254 : 1);
+
+    if (!get_config_addr(root, sym(netmask), &netmask))
+        netmask.addr = 4294967040; // 255.255.255.0
```
wjhun, line 207: Hex would be more readable here, and again use lwip_htonl. It may also be preferable to set addresses using "ip4_addr_set_u32" should the fields in ip4_addr_t change in a future lwIP release.

src/net/net.c
am11: Fixed. Thanks for the byte order tip, I didn't realize that the web app running in nanos was actually _not_ responding (although the desired IP was assigned, the mask was wrong)

src/net/net.c
am11: Converted to hex.

francescolavra (commented):
Looks good to me

am11:
@francescolavra, aside, would it make sense to block on DHCP call? currently the user code starts before and we get the IP assigned asynchronously with dhcp.

Static (expected execution order)
image

DHCP (unexpected order)
image

francescolavra:
In my opinion, the program shouldn't be blocked waiting for an address to be assigned by the DHCP server, because the DHCP server is external to the instance (i.e. the instance doesn't have control over it), and a well behaved program should be able to deal with the IP address being assigned late or not being assigned at all (see also https://github.com/nanovms/nanos/issues/344).

am11:
Opened #1431 for DHCP asynchronous behavior question.

am11:
Copy that, closed the issue. Thanks for reference link! (i should have searched it before opening the issue) 8-)

am11:
Opsie, didn't mean to close this one.

src/net/net.c
```diff
@@ -196,11 +196,22 @@ static boolean get_static_config(tuple root, struct netif *n, boolean trace) {
...
+    if (!get_config_addr(root, sym(ipaddr), &ip))
         return false;
 
+    if (!get_config_addr(root, sym(netmask), &netmask))
+        ip4_addr_set_u32(&netmask, lwip_htonl(0xffffff00)); // 255.255.255.0
+
+    if (!get_config_addr(root, sym(gateway), &gw)) {
+        // common best practices are: network + 1 or broadcast - 1,
+        // so we will use latter if former is in use.
+        u32_t ip_after_network = (netmask.addr & ip.addr) + 1;
```
wjhun, line 208: I think you still need lwip_ntohl around (netmask.addr & ip.addr) here (and also (~(netmask.addr) | ip.addr) below) before applying any computation, given that the addresses are in network byte order.

am11: I have simplified it so we don't need to convert it twice.

wjhun (approved):
Looks good. Thanks!

PR #1433, opened 2021-03-18
reduce kernel image size and footprint

wjhun:
These changes are aimed at lowering the minimum memory requirements to run a Nanos instance. The kernel image file is stripped of debug info (while retaining symbols), reducing the size of the pc kernel image from 7.9M to 1M, or ~87%. Reducing the image size also speeds up the image boot time, as the kernel load in stage2 consumes a significant portion of this time.

Also, the stage2 working heap size has been reduced to 4MB, which is sufficient for loading a boot filesystem of at least 2M in size. With these changes, a webg instance can run in a qemu VM with 128MB of memory. Further optimizations, such as more judicious use of 2M size pages (such as in the mcache), could further lower this minimum requirement.

This also resolves a bug where a section of an ELF file (.strtab in the found case) that shares a page with a loadable segment could be corrupted by the zeroing of the BSS area in the segment. Instead of mapping this partial page from the ELF buffer and zeroing the remainder, the loadable bits are copied into the space allocated and mapped for the BSS section.

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -1,8 +1,8 @@
 SUBDIR=		boot
 LWIPDIR=	$(VENDORDIR)/lwip
-PROGRAMS=	kernel.img
+PROGRAMS=	kernel.elf
 
-SRCS-kernel.img= \
+SRCS-kernel.elf= \
 	$(CURDIR)/service.c \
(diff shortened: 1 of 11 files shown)


sanderssj:
Looks good to me!

PR #1434, opened 2021-03-19
revert id heap default allocation behavior to first-fit

wjhun:
The id heap now defaults to first-fit allocation. Previously, the id heap
would default to starting the allocation search from the location of a
previous allocation or deallocation. This behavior was leading to excessive
fragmentation of resources, as could be easily demonstrated by making
alternating requests of 2M and 4K-sized pages from the physical heap. After a
4K allocation, a subsequent 2M allocation would start at the next 2M
boundary (to satisfy the id heap alignment property), and the preceding unused
space would not be considered for allocation unless the allocation search
wrapped within the id range or the next_bit was explicitly reset using
id_heap_set_next.

Next-fit behavior may still be induced by the calling code with use of
id_heap_set_next.

### DIFF
--- a/src/runtime/heap/id.c
+++ b/src/runtime/heap/id.c
@@ -122,7 +122,6 @@ static u64 id_alloc_from_range(id_heap i, id_range r, u64 pages, range subrange)
     if (bit == INVALID_PHYSICAL)
         return bit;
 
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
Looks good to me

PR #1436, opened 2021-03-22
Reserve extension during flush_log_extension

sanderssj:
There is a scenario in which a fileystem rebuild creates a number of new log
extensions and multiple flush_log_extensions calls and a eventually a
log_extend_link which will call flush_log_extension again to close one of the
new log extensions before it has finished its first flush_log_extension
and thus a use-after-free that leads to a crash.
The change bumps the refcount on the extension for a flush_log_extension.


wjhun (approved):
LGTM

PR #1438, opened 2021-03-22
Perform immediate shutdown after page fault panics

sanderssj:
Currently the page fault handler calls halt to try to perform an orderly
shutdown and save crash dump information to disk on panic, but this is often
a problem if a bad page fault occurs in certain kernel code paths, such as
the page cache, where spinlocks are held. Since the spinlocks aren't
cleaned up, the shutdown code accessing the pagecache will generate
an assertion failure (with lock debugging) or deadlock. This change
performs an immediate shutdown to prevent a deadlock or assertion
panic loop. However, this precludes saving crash dump information
for page faults.

We would need an alternative, polling, storage driver in order to bypass the
pagecache and work with interrupts off to save the crash dump.

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -228,7 +228,8 @@ define_closure_function(1, 1, context, default_fault_handler,
 //        init_tcp_gdb(heap_general(get_kernel_heaps()), p, 9090);
 //        thread_sleep_uninterruptible();
     }
-    halt("halt\n");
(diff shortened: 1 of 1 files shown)


wjhun (approved):
I think this is the right move as the kernel may be in a very inconsistent state here, and trying to keep the runloop, pagecache and interrupt-driven storage drivers would be risky business anyway. I agree with the idea of using a special polled I/O routine to dump in a known location on storage (or over network).

PR #1442, opened 2021-03-24
pvscsi: take queue_lock before enqueueing to hcb_queue

wjhun:
This trivial fix was found through inspection.

### DIFF
--- a/src/vmware/pvscsi.c
+++ b/src/vmware/pvscsi.c
@@ -135,11 +135,11 @@ static void pvscsi_action_io_queued(pvscsi dev, struct pvscsi_hcb *hcb, u16 targ
         spin_unlock(&dev->queue_lock);
         return;
     }
-    spin_unlock(&dev->queue_lock);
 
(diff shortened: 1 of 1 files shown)


src/vmware/pvscsi.c
```diff
@@ -138,7 +138,9 @@ static void pvscsi_action_io_queued(pvscsi dev, struct pvscsi_hcb *hcb, u16 targ
     spin_unlock(&dev->queue_lock);
```
sanderssj, line 138: I noticed that pvscsi_rx_service_bh holds dev->queue_lock while processing the list, which includes calling pvscsi_action_io. Would it better to move this unlock below the pvscsi_action_io here instead of unlocking and re-locking?

francescolavra:
LGTM

src/vmware/pvscsi.c
wjhun: I agree - update pushed.

Issue #1416, opened 2021-03-01
add hyper-v scsi (gen2) support


eyberg:
https://github.com/nanovms/ops/issues/547

nanos currently works on hyper-v gen1 and it looks like we are using that on azure as well:

```
azure_image.go:                 HyperVGeneration: compute.HyperVGenerationTypesV1,
```

scsi should be much faster

am11:
Some notes from https://community.spiceworks.com/topic/451467-hyperv-scsi-controller-v-ide-controller?page=1#entry-3877743

#### Gen 1

The IDE controller is emulated (vmwp.exe) where as the SCSI controller is synthetic.

The IDE controller implements an emulated IDE controller which means there is extra processing before the I/O is sent to the disk.

The SCSI controller in Windows Server virtualization is not emulated, instead it uses the VMBUS (Virtual Machine BUS). These devices have an advantage over emulated devices because they leverage the existing Windows driver architecture.

With SCSI, we can hot-swap a drive, resize (incl. shrink), without shutting down the VM, which is not possible with IDE.

It seems, we cannot boot with gen 1 SCSI drives.

#### Gen 2

Supports UEFI and therefore supports native VMBUS on boot. "Gen2" VM's, SCSI and IDE both, utilize VMBUS therefore there is no performance gains on "Gen2" VM's in using either.

Not sure about hot-swapping or resizing of live drive with Gen 2 IDE, but SCSI has this ability with both generations.

---

Verdict: `add hyper-v scsi (gen2) support` (aka title of this issue) is the way to enhance HyperV support! 😎

francescolavra:
SCSI is actually supported already by Nanos, which can use SCSI volumes just fine. The problem is that hyper-v gen1 VMs cannot be configured to boot from SCSI, because the hyper-v BIOS supports only IDE as bootable hard disks. So SCSI disks in hyper-v gen1 VMS can only be used as additional (non-bootable) volumes, while the boot disk must be IDE.

In order to support gen2 VMS (which use UEFI and can boot from SCSI disks), we need to add support for UEFI-based booting to the Nanos bootloader.

am11:
Ah, makes sense. Should we re-purpose this issue for UEFI support (regardless of target provider)?

Seems like the [RH's shim](https://github.com/rhboot/shim) (signed by MSFT's key) is used with two approaches:
* [In Fedora](https://fedoraproject.org/wiki/Features/SecureBoot)
[...]

PR #1443, opened 2021-03-28
x86 boot: add UEFI bootloader

francescolavra:
This allows Nanos to boot on UEFI platforms, such as Hyper-V Generation-2 instances (#1416).
The UEFI bootloader file is called bootx64.efi and is located in the ESP (EFI System Partition), in a FAT32 filesystem in the EFI/Boot/ folder. In order to create this partition in the Nanos image file, the ENABLE_UEFI environment variable needs to be defined when running `make`.
Example command to create a Hyper-V image:
`make vhdx-image ENABLE_UEFI=1`
The mkfs and dump tools, as well as the stage2 and kernel code, have been modified in order to properly handle images with and without the ESP.

For the Mac build, the objcopy tool (from the binutils package) is being used instead of x86_64-elf-objcopy, because the latter does not support the efi-app-x86_64 BFD target that is used to create the UEFI bootloader file.


src/boot/uefi.c
```diff
@@ -0,0 +1,272 @@
...
+# define uefi_debug(x, ...) rprintf("UEFI: " x "\n", ##__VA_ARGS__)
+#else
+# define uefi_debug(...)
+#endif
+
+struct efi_guid uefi_block_io_proto = {
+        0x964e5b21, 0x6459, 0x11d2, {0x8e, 0x39, 0x00, 0xa0, 0xc9, 0x69, 0x72, 0x3b}
+};
+struct efi_guid uefi_rng_proto = {
+        0x3152bca5, 0xeade, 0x433d, {0x86, 0x2e, 0xc0, 0x1c, 0xdc, 0x29, 0x1f, 0x44}
```
am11, line 25: Do these correspond to https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs? Would it make sense to use existing Linux' GUID for nanos (as it is already trusted)?

francescolavra: No, these are not partition UUIDs, they are UEFI protocol UUIDs, defined in the UEFI spec (see for example https://gitlab.com/ncroxon/gnu-efi/-/blob/master/inc/efiprot.h for these and other protocol-related definitions).

tools/mkfs.c
```diff
@@ -419,6 +626,12 @@ int main(int argc, char **argv)
         if (nr < 0) {
             halt("couldn't read from boot image file %s: %s\n", bootimg_path, strerror(errno));
         }
+
+        /* The on-disk kernel log dump section is immediately before the first partition. */
+        offset += KLOG_DUMP_SIZE;
+
+        if (uefi_loader)
+            offset = write_uefi_part(out, offset, uefi_loader);
```
am11, line 634: If the hypervisor does not support UEFI and image was built with UEFI, would it still work? i.e. the legacy-BIOS bits are still written to the image.

* if so, would it make sense to always include UEFI support (and add option to skip UEFI if it makes a a significant difference in size)?
* if not, would it make sense to make it work with both, legacy-BIOS and UEFI, modes (like modern OS support)?

francescolavra: Yes, an image built with UEFI will still work on hypervisors that don't support UEFI.
I'm still working on adding the required functionality in Ops, but the idea is to include UEFI support only when needed, since it adds 33 MB to the image size.

am11: Great! and wow, 33MB is significant enough. 😁
(must be the bare minimum size to support UEFI)

I think in ops, `hyper-v2` and `azure2` providers could be (additionally) added for users to explicitly choose gen1 or gen2 of Hyper-V VMs.

tools/mkfs.c
```diff
@@ -15,6 +15,119 @@
...
+        res = !strcmp((str) + slen - sslen, (substr));  \
+    else                                                \
+        res = false;                                    \
+    res;                                                \
+})
+
+/* Must be large enough to contain the minimum number of clusters (64K) for a
+ * 32-bit File Allocation Table. */
+#define UEFI_PART_SIZE  (33 * MB)
+
```
sanderssj, line 32: Can we add a comment about where these blobs come from or how they are generated?

src/boot/boot.h
```diff
@@ -0,0 +1,46 @@
...
+    ta->h.dealloc = tagged_deallocate;
+    ta->tag = tag;
+    ta->parent = h;
+    return (heap)ta;
+}
+
+static inline void *tag(void *v, u16 tval)
+{
+    *((u8 *)v-1) = tval;
+    return v;
```
sanderssj, line 35: I know these functions come from old code, but should we just make tval into a u8 since the upper byte is thrown away anyway?

src/boot/boot.h
```diff
@@ -0,0 +1,46 @@
...
+}
+
+static inline u16 tagof(void *v)
+{
+    return *((u8 *)v-1);
+}
```
sanderssj, line 41: Same as above, should return a u8 instead?

tools/mkfs.c
francescolavra: OK, I added a comment for each blob that says what the blob represents. I also found that I wasn't building the File Allocation Table generically enough, since the number of table entries depends on the file size; this is now fixed in write_uefi_part(), and as a side effect the code is also less cryptic :)

src/boot/boot.h
francescolavra: Right, I changed tval to a u8

src/boot/boot.h
francescolavra: Returning a u8 now

francescolavra:
In the Makefile I added handling of the the ENABLE_UEFI environment variable, so that it's possible to create an image with the UEFI partition via `make`.
Example command to create an image for Hyper-V Gen 2 VMs:
`make vhdx-image ENABLE_UEFI=1`

The commit with the page flag fix in init_vsyscall() has been dropped since the fix has been applied in #1445.

.circleci/config.yml
```diff
@@ -56,6 +56,7 @@ jobs:
       - run: CI=1 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
       - run: brew update && brew install nasm go wget qemu ent
       - run: brew tap nanovms/homebrew-x86_64-elf
```
wjhun, line 58: We added this tap before x86_64-elf-binutils was added to homebrew-core. Since it was added, the install step below has been silently preferring the one in homebrew-core, which seems to be just fine. So I would just remove these instances of 'brew tap'.

wjhun (approved):
Looks great - only minor comments

src/boot/uefi.c
```diff
@@ -0,0 +1,272 @@
...
+
+static void uefi_dealloc(struct heap *h, u64 a, bytes b)
+{
+    UBS->free_pool(pointer_from_u64(a));
+}
+
+/* Returns page-aligned addresses */
+static u64 uefi_alloc_aligned(heap h, bytes b)
+{
+    u64 buf = uefi_alloc(h, b + h->pagesize);
```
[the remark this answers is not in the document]
francescolavra: OK, now the size being allocated is multiple of the page size.

tools/mkfs.c
```diff
@@ -250,7 +366,90 @@ closure_function(4, 2, void, fsc,
     closure_finish();
 }
 
-static void write_mbr(descriptor f)
+static void write_blob_padded(descriptor out, u8 *blob, size_t len, boolean with_trailer)
+{
+    assert(write(out, blob, len) == len);
+    u8 trailer[] = {0x55, 0xAA};
+    if (len <= SECTOR_SIZE - (with_trailer ? sizeof(trailer) : 0)) {
```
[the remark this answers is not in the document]
francescolavra: Removed this redundant expression by using a `padded_len` helper variable

.circleci/config.yml
francescolavra: OK, removed the nanovms/homebrew-x86_64-elf tap.

am11:
> `make vhdx-image ENABLE_UEFI=1`

@francescolavra, does this also mean that we will have two separate builts of nanos; uefi and non-uefi, for `ops(1)` to consume?

francescolavra:
@am11 no, the kernel build continues to be a single build, the `ENABLE_UEFI` environment variable only affects whether UEFI is included in a target image (the Makefile can be used not only to build Nanos itself, but also to build complete disk images).

tools/mkfs.c
francescolavra: In https://github.com/nanovms/ops/pull/981, I opted to not define additional cloud providers, but add handling of the `ImageType` CloudConfig attribute (and for Hyper-V the `Flavor` CloudConfig attribute as well).

PR #1444, opened 2021-03-29
pvscsi: add support for non-MSI interrupts

wjhun:
This change restores use of the pvscsi device under qemu. Add "STORAGE=pvscsi" to the make command line to enable the pvscsi device.

### DIFF
--- a/src/vmware/pvscsi.c
+++ b/src/vmware/pvscsi.c
@@ -4,6 +4,7 @@
 #include "storage.h"
 #include "virtio/scsi.h"
 
+//#define PVSCSI_DEBUG
 #ifdef PVSCSI_DEBUG
 # define pvscsi_debug rprintf
 #else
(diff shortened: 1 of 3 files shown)


src/vmware/pvscsi.c
```diff
@@ -345,7 +347,9 @@ closure_function(4, 0, void, pvscsi_report_luns_done,
     u16 target = bound(target);
     pvscsi_debug("%s: target %d, host_status %d, scsi_status %d\n",
         __func__, target, hcb->host_status, hcb->scsi_status);
-    if (hcb->host_status != BTSTAT_SUCCESS || hcb->scsi_status != SCSI_STATUS_OK) {
+    /* BTSTAT_DATARUN is just residue, not an error */
+    if ((hcb->host_status != BTSTAT_SUCCESS && hcb->host_status != BTSTAT_DATARUN)
```
sanderssj, line 351: Is the data overrun is because the lun list is larger than fits in the output buffer? Are we sure we wouldn't miss a legitimate lun if we don't requery with a larger buffer?

wjhun: In this case it isn't an overrun but rather a short transfer. The request size is sufficient for the maximum number of LUNs that can be reported, but there is only one LUN reported, causing a response that is short of the requested size. The pvscsi device in qemu sets the completion hostStatus to BTSTAT_DATARUN in such a case. (See pvscsi_command_complete in hw/scsi/vmw_pvscsi.c.)

sanderssj: Gotcha, I thought it seemed a little weird to be an error but I can see how it can make sense.

sanderssj:
LGTM!

wjhun:
I probably broke MSI as it was using the handler that was modified for INTx. I just pushed a fix and will test on ESX before merging.

PR #1445, opened 2021-03-29
enable C0_WP in cpu_init

wjhun:
This enables write protection while running at CPL 0. Any attempts to write to
a read-only page at kernel level will cause a fault. To comply with this
restriction, map_and_zero() and init_vsyscall() have been updated to first map
read-only pages as writable, initialize the page data and then update the map
flags to remove the writable flag.


francescolavra (commented):
Looks good to me.

PR #1446, opened 2021-04-02
Add objcache-compatible heap debug

sanderssj:
This PR adds support for the objcache to the mem debug heap wrapper since the objcache was not
previously compatible with it. The objcache implementation had to change slightly since a pointer to
the objcache is cached in each page, bypassing the wrapper. The footer can now store either the
objcache pointer or the wrapper pointer depending on how the objcache is created.

By applying the debug wrapper to the objcaches in the mcache, I discovered a use-after-free problem
in vmap_remove_intersection where the node match could be deallocated before it gets used for 
unmapping.

Since I had to touch objcache.c, I updated it to spaces only in order to eliminate the mixture of tabs
and spaces. It makes the diff very noisy however, so I suggest viewing that file with leading/trailing whitespace
changes turned off.

There's also some minor cleanup in mem_debug.c

### DIFF
--- a/src/runtime/heap/objcache.c
+++ b/src/runtime/heap/objcache.c
@@ -20,27 +20,28 @@
 
 #include <runtime.h>
 
-#define FOOTER_MAGIC	(u16)(0xcafe)
+#define FOOTER_MAGIC    (u16)(0xcafe)
 
 typedef struct objcache *objcache;
 typedef struct footer {
-    u16 magic;			/* try to detect corruption by overruns */
-    u16 free;			/* next free (recycled) object in page */
-    u16 head;			/* next uninitialized object in page */
-    u16 avail;			/* # of free and uninit. objects in page */
-    objcache cache;		/* objcache to which this page belongs */
-    struct list list;		/* full list if avail == 0, free otherwise */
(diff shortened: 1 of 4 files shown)


src/runtime/heap/objcache.c
```diff
@@ -20,27 +20,28 @@
...
-    u64 objs_per_page;		/* objects per page */
-    u64 total_objs;		/* total objects in cache */
-    u64 alloced_objs;		/* total cache occupancy (of total_objs) */
+    struct list free;       /* pages with available objects */
+    struct list full;       /* fully-occupied pages */
+    bytes pagesize;         /* allocation size for parent heap */
+    u64 objs_per_page;      /* objects per page */
+    u64 total_objs;         /* total objects in cache */
+    u64 alloced_objs;       /* total cache occupancy (of total_objs) */
+    heap wrapper_heap;      /* heap wrapper */
```
francescolavra, line 44: This new struct member should be initialized in allocate_objcache(), otherwise it could get a random value when a (non-debug) objcache is created.

sanderssj: Yup, properly initialized now!

wjhun (approved):
LGTM

PR #1447, opened 2021-04-03
Fixes for Hyper-V Generation-2 instances

francescolavra:
In UEFI-based systems, the SMBIOS entry point can be retrieved from the UEFI configuration tables. With this PR, the UEFI bootloader retrieves the SMBIOS entry point and passes it to the kernel via a newly defined region type (REGION_SMBIOS); the kernel uses the SMBIOS entry point, if known, to find the location of the DMI tables.
This allows retrieving the DMI tables in UEFI-based systems, which is needed by the cloud-init klib to properly detect the cloud platform on which the kernel is running.
In addition, when running in a Generation 2 VM, after sending the "report ready" packet to the Azure platform, the guest receives from the hypervisor a VSTOR_OPERATION_ENUMERATE_BUS message in the Storvsc VMBus channel. Even though the driver doesn't support re-enumeration of the VMBus, it shouldn't halt the VM on reception of such message.


sanderssj:
LGTM!

PR #1449, opened 2021-04-09
Radar: add more robust retry mechanisms

francescolavra:
The radar klib now checks the server response to crash reports, and retries sending a crash report if the response is not received or does not contain an OK status code.
A retry mechanism is implemented for boot reports as well, so that a boot report is resent if no valid boot ID has been received.
Finally, a retry is triggered if connection to the server fails when sending either a crash or a boot report.

### DIFF
--- a/klib/radar.c
+++ b/klib/radar.c
@@ -52,6 +52,7 @@ static struct telemetry {
     tuple env;
     buffer auth_header;
     klog_dump dump;
+    boolean dump_done;
     u64 boot_id;
     boolean running;
     timestamp retry_backoff;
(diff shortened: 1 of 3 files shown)


klib/radar.c
```diff
@@ -156,21 +180,29 @@ closure_function(2, 1, status, telemetry_recv,
...
-            telemetry.running = true;
+            if (telemetry.boot_id) {
+                /* The boot event has been sent: start collecting usage metrics. */
+                for (int count = 0; count < RADAR_STATS_BATCH_SIZE; count++)
+                    telemetry.stats_mem_used[count] = heap_allocated(telemetry.phys);
+                telemetry_stats_send();
+                telemetry.stats_count = 0;
+                kfunc(register_timer)(CLOCK_ID_MONOTONIC, RADAR_STATS_INTERVAL, false,
+                        RADAR_STATS_INTERVAL, (timer_handler)&telemetry.stats_func);
+                telemetry.running = true;kfunc(timm_dealloc)(0);
```
sanderssj, line 202: These lines got joined together or maybe the timm_dealloc isn't supposed to be there?

sanderssj (commented):
Looks good to me!

wjhun (approved):
LGTM

klib/radar.c
francescolavra: Oops, the timm_dealloc wasn't supposed to be there, I put it there temporarily to do a test and then forgot to remove it

PR #1450, opened 2021-04-09
Implement TFS FUSE driver

sanderssj:
This PR implements a read/write FUSE driver for TFS that allows mounting the root filesystem on a raw disk image.
Since it requires the fuse kernel driver to run and fuse libraries to compile, it is not built by default and must be built
separately with 'make tfs-fuse'. Basic operation is to run it with a mount point and the disk image to mount, e.g.
'output/tools/bin/tfs-fuse /tmp/mnt output/image/disk.raw'.

The system-independent filesystem path related code found in filesystem.[ch] and syscall.c has been moved
into tfs.c and tfs.h. Since only system code knows about what the root filesystem is and how to cross mounts,
a couple of helper callbacks have been created which system code must initialize before using path resolution.

The fuse driver model closely imitates file-related syscalls so the tfs-fuse driver uses a lot of code from syscall.c
to implement the fuse callbacks, with modifications as necessary to accommodate fuse implementation requirements.


tools/tfs-fuse.c
```diff
@@ -0,0 +1,1013 @@
...
+
+static int tfs_write(const char *path, const char *buf, size_t size, off_t off, struct fuse_file_info *fi)
+{
+    tfs_fuse_debug("%s: path %s\n", __func__, path);
+    pthread_rwlock_wrlock(&rwlock);
+    int fd = fi->fh;
+    fdesc f = resolve_fd(fd);
+    int rv;
+    tfs_fuse_debug("  fd %d fdesc %p\n", fd, f);
+    if (!f->read){
```
francescolavra, line 561: if (!f->write) {

tools/tfs-fuse.c
```diff
@@ -0,0 +1,1013 @@
...
+    return seconds(t->tv_sec) + nanoseconds(t->tv_nsec);
+}
+
+static int tfs_utimens(const char *filename, const struct timespec tv[2])
+{
+    timestamp atime =
+        tv ? time_from_timespec(&tv[0]) : now(CLOCK_ID_REALTIME);
+    timestamp mtime =
+        tv ? time_from_timespec(&tv[1]) : now(CLOCK_ID_REALTIME);
+    return utime_internal(filename, atime, mtime);
```
francescolavra, line 920: I think before calling utime_internal() it should call pthread_rwlock_wrlock(&rwlock)

tools/tfs-fuse.c
```diff
@@ -0,0 +1,1013 @@
...
+}
+
+static u64 get_fs_offset(descriptor fd, int part, boolean by_index, u64 *length)
+{
+    char buf[512];
+
+    ssize_t nr = read(fd, buf, sizeof(buf));
+    if (nr < 0 || nr < sizeof(buf)) {
+        perror("read");
+	exit(EXIT_FAILURE);
```
wjhun, line 227: indent

wjhun (approved):
Minor comments, otherwise LGTM. Very useful!

src/tfs/tfs.h
```diff
@@ -110,3 +111,103 @@ u64 fs_usedblocks(filesystem fs);
...
+
+/* Same as resolve_cstring(), except that if the entry is a symbolic link this
+ * function follows the link (recursively). */
+int filesystem_resolve_cstring_follow(filesystem *fs, tuple cwd, const char *f, tuple *entry,
+        tuple *parent);
+
+int filesystem_follow_links(filesystem *fs, tuple link, tuple parent,
+                            tuple *target);
+
+int filesystem_add_tuple(const char *path, tuple t);
```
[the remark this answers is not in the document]
sanderssj: Agreed!

tools/tfs-fuse.c
sanderssj: Yes, good catch.

src/tfs/tfs.h
```diff
@@ -110,3 +111,103 @@ u64 fs_usedblocks(filesystem fs);
...
+    if (!filename) {
+        filename = path;
+    } else {
+        filename++;
+    }
+    return filename;
+}
+
+/* Expects an empty buffer, and never resizes the buffer. */
+static inline boolean dirname_from_path(buffer dest, const char *path)
```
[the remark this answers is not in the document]
sanderssj: Done!

tools/tfs-fuse.c
```diff
@@ -0,0 +1,1013 @@
...
+    u64 fs_offset = rootfs_part->lba_start * SECTOR_SIZE;
+    if (length)
+        *length = rootfs_part->nsectors * SECTOR_SIZE;
+    tfs_fuse_debug("detected filesystem at 0x%llx\n", fs_offset);
+    return fs_offset;
+}
+
+static void set_flush_timeout()
+{
+    alarm(0);
```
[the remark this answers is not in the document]
sanderssj: Right, it is unnecessary.

tools/tfs-fuse.c
sanderssj: A tab snuck in!

tools/tfs-fuse.c
sanderssj: Yes, good catch. The modification of the metadata needs to be protected.

tools/tfs-fuse.c
```diff
@@ -0,0 +1,1013 @@
...
+}
+
+static int tfs_open(const char *name, struct fuse_file_info *fi)
+{
+    pthread_rwlock_wrlock(&rwlock);
+    int fd = open_internal(name, fi->flags, 0);
+    pthread_rwlock_unlock(&rwlock);
+    if (fd < 0)
+        return fd;
+    fi->fh = fd;
```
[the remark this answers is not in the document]
sanderssj: The fuse header file says that the open callback will not be passed creation or truncation flags as it first calls other appropriate callbacks before open, so we don't need to worry about modification here.

PR #1452, opened 2021-04-13
User address space: replace id heaps with rangemap-based heap

francescolavra:
Id heaps use an underlying bitmap for storing information on what pages are allocated (1 bit for each page). This doesn't scale when using 4KB pages if the allocated address space reaches a few terabytes.
This change replaces the virtual and virtual_page id heaps of the user process with a rangemap-based heap. An address range allocated on the virtual heap but not mapped is represented by a struct vmap whose flags have the VMAP_FLAG_MMAP bit cleared; as a result, it cannot be mapped on-demand when a page fault occurs (unless the user process requests a mapping later on). Memory areas being mapped with VMAP_FLAG_PREALLOC are not released upon unmapping, so that the underlying address range is not released in the virtual heap until an explicit deallocation is requested.

Handling of userspace tagged memory areas on x86_64 is now done by putting the tag value just before an allocated area, instead of reserving virtual address space areas in the kernel to be mmapped by userspace.
This allows the virtual heap upper limit to be increased to USER_LIMIT, i.e. the virtual heap is now 8 times larger.

A simple fix in mremap() prevents an existing vmap from being removed when the mremap() call encounters an error.

Finally, there are a couple of fixes to the Aarch64 code that solve issues found when running on CI the kernel with the new rangemap-based heap.


src/unix_process/unix_process_runtime.c
```diff
@@ -85,25 +85,33 @@ void halt(char *format, ...)
...
 struct tagheap {
     struct heap h;
     heap mh;
-    u64 vtag;
+    u64 tag;
 };
 
+#ifndef __aarch64__
+
+static inline u64 tag_alloc(heap h, bytes length)
```
wjhun, line 96: I suggest that malloc_alloc and malloc_free are also made to allocate an additional (zeroed) byte like the tagged methods here - in essence making every alloc a tagged alloc. Existing code expects to be able to check the tag of any value which may include values that weren't allocated from a tagged region, like string buffers. This works by default with address tagging, but without a padded alloc here it could result in a dereference outside the allocated area, leading either to a fault or an incorrect tag being read.

wjhun (approved):
One issue regarding unix process general allocs needing to reserve tag space, otherwise LGTM.

src/unix_process/unix_process_runtime.c
```diff
@@ -85,25 +85,33 @@ void halt(char *format, ...)
...
+{
+    struct tagheap *th = (struct tagheap *)h;
+    u64 base = allocate_u64(th->mh, length + 1);
+    return base + 1;
```
wjhun, line 100: The tag is not being inserted here; need *(u8*)pointer_from_u64(base) = th->tag.

I found this same bug in src/boot/boot.h:tagged_allocate.

src/unix_process/unix_process_runtime.c
francescolavra: OK, I made malloc_alloc() and malloc_free() arch-specific (with the x864_64 variants allocating extra space for the tag), while tag_alloc() and tag_dealloc() are now arch-independent (the arch-specific part is in the tag() inline function).
The extra space for the tag has to be 8 bytes, not 1, otherwise mks segfaults somewhere within filesystem_write_linear(). While I haven't identified the root cause of this, I think it's wise to ensure in any case that the pointer returned by the alloc function is aligned to 64 bits (even though Intel CPUs support unaligned accesses, there may be a performance penalty).

src/unix_process/unix_process_runtime.c
francescolavra: Right, this is now fixed in https://github.com/nanovms/nanos/pull/1452/commits/27f12b82d5401d9acdd53bc44ae753bcdb376e14. (I also changed the extra space for the tag to be 8 bytes there too, so that bootloaders running in long mode don't have to do unaligned memory accesses.)

PR #1454, opened 2021-04-19
function-backed tuples and support for remote management of value space

wjhun:
This change introduces a new type of tuple, called function_tuple, that allows tuple attribute access to be handled by a backing function. Arbitrary actions can take place when a value is retrieved (via get or iterate) or stored (with set). Live statistics and running state, polled on retrieval, can now exist within the value space along with static values. State outside of the value space may be affected by setting values. These dynamic properties are exploited by the tuple_notifier type, one such use of a function_tuple, which is used to invoke notify actions on a configuration change (e.g. change in debug flags like "trace" and "debugsyscalls") as well as poll for stats on inquiry (e.g. the management interface to the heap allocators).

Tuples are no longer synonymous with the table type and are now represented by a union of struct table and struct function_table, with the variant identified by tag type. Correspondingly, the use of "table_find" to retrieve values has been replaced by "get", "table_set" by "set", and instances of table_foreach operating on a tuple have been refactored to use the "iterate" interface.

A very simple telnet-based tuple interface, enabled only when MANAGEMENT=telnet is specified in the make commandline, is included here as a simple interface for debugging. Requests are handled by the tuple parser and responses are returned in tuple format (with a "result" attribute representing an error, just as with status tuple within the kernel). Requests may also be attached to a periodic timer, making a very simple way to poll for changes in attributes over time.

To support this interface, several advances have been made to tuple (value) formatting. print_tuple and the '%t' format have been superceded by print_value and the existing '%v' formatter as well as a new '%V' formatter which accepts an additional tuple of formatting attributes, including options to prune output at a certain depth or pretty-print tuple output with newlines and indentation. Previously, tuple printing would traverse blindly into cycles, leading to a stack overflow and crash when printing tuples with back references (found in filesystem metadata with "." and ".."). To avoid this runaway traversal, print_value now maintains a table of nodes visited during the print, outputting `"<visited>"` for a node that has already been printed.


src/unix_process/unix_process_runtime.c
```diff
@@ -157,7 +158,8 @@ static heap allocate_tagged_region(heap h, u64 tag)
     th->h.pagesize = 32; // XXX
     th->h.allocated = 0;
     th->h.total = 0;
-    tag_debug("tag %d, heap %p\n", tag, th);
+    th->h.management = 0;
+    tag_debug("tag %d, bits 0x%lx, heap %p\n", tag, th->vtag, th);
```
francescolavra, line 162: This probably slipped through the rebase merge, there is no th->vtag anymore.

test/unit/parser_test.c
```diff
@@ -125,18 +125,20 @@ PARSE_TEST(partial_comment_test, "#[]\n()")
 {
     test_no_errors();
     test_assert(root != NULL);
-    test_assert(root->count == 0);
+    test_assert(tuple_count(root) == 0);
     return true;
 }
 
+/* XXX Most of these calls to get should be to get_string; but they currently
+   fail due to the user va tags issue. Fix once user tags are changed over to rewind type. */
```
francescolavra, line 133: I guess now this can be changed to use get_string()

src/tfs/tfs.h
```diff
@@ -162,6 +151,8 @@ static inline const char *filename_from_path(const char *path)
     return filename;
 }
 
+symbol lookup_sym(tuple parent, tuple t);
+
```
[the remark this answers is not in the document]
wjhun: Thanks - this was an inline that I needed to move to filesystem.c, but should instead go to tfs.c. I missed some other fuse-related changes in the rebase and am resolving those now.

src/kernel/storage.c
```diff
@@ -147,22 +148,40 @@ void storage_set_root_fs(filesystem root_fs)
...
 }
 
+closure_function(1, 2, boolean, volume_add_mount_each,
+                 volume, v,
+                 value, k, value, path)
+{
+    assert(is_symbol(k));
+    assert(!is_tuple(path)); // XXX is_string
+    if (volume_match(k, bound(v)))
+        volume_mount(bound(v), path);
```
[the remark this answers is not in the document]
wjhun: Ok, I added the is_string asserts above and a return false after the volume_mount.

src/kernel/locking_heap.c
```diff
@@ -1,10 +1,13 @@
 #include <kernel.h>
+#include <management.h>
 
 typedef struct heaplock {
     struct heap h;
     struct spinlock lock;
     heap parent;
     heap meta;
+    tuple mgmt;
+    tuple parent_mgmt;
```
sanderssj, line 10: parent_mgmt doesn't get initialized, although it doesn't seem like it would be accessed before initialization.

src/runtime/heap/mcache.c
```diff
@@ -25,6 +26,7 @@ typedef struct mcache {
     u64 pagesize;
     u64 allocated;
     u64 parent_threshold;
+    tuple mgmt;
```
sanderssj, line 29: Also doesn't get initialized

src/runtime/heap/objcache.c
```diff
@@ -42,6 +44,7 @@ typedef struct objcache {
     u64 total_objs;         /* total objects in cache */
     u64 alloced_objs;       /* total cache occupancy (of total_objs) */
     heap wrapper_heap;      /* heap wrapper */
+    tuple mgmt;
```
sanderssj, line 47: Another not initialized (unless meta is guaranteed to zero alloc?)

sanderssj (commented):
Looks good to me! I noticed a couple of new fields in some heap structs that weren't initialized on heap allocation.

src/kernel/locking_heap.c
wjhun: Thanks for catching these. I pushed a fix.

francescolavra:
LGTM

src/runtime/format.c
```diff
@@ -172,7 +172,7 @@ KLIB_EXPORT(rsnprintf);
 void rprintf(const char *format, ...)
 {
     /* What's a reasonable limit here? This needs to be reentrant. */
-    buffer b = little_stack_buffer(1024);
+    buffer b = little_stack_buffer(32768);
```
francescolavra, line 175: This causes a stack overflow when executing in an exception handler on x86, where exception handlers have a 32 KB stack, which means that when an unhandled fault occurs, the register dump and frame and stack traces that are supposed to be printed on the console are often replaced by those associated to this stack overflow, with a generic "frame xxxx already full" message that hides the original fault.
Would it be OK to decrease the rprintf buffer size here to, say, 16 KB?

wjhun: > Would it be OK to decrease the rprintf buffer size here to, say, 16 KB?

Sure. This change shouldn't have been in this merge - but I guess I let it slip as it seemed rather frequent that a "!buffer_is_wrapped" assertion would be hit while debugging and printing >1KB of output (e.g. when printing the full content of a tuple or buffer). 32KB was an arbitrary choice.

While we're at it, we may as well increase the size of the exception handler stack - I can't think of any reason why it needs to be constrained to 32KB. All these numbers are arbitrary - we just need to make sure things don't fall over if someone breathes on it the wrong way.

francescolavra: #1478

PR #1455, opened 2021-04-20
Fix the setting of origin time in ntp query packet

sanderssj:
This got accidentally messed up when fixing some of the compiler warnings around alignment.


francescolavra (commented):
Looks good to me

wjhun (approved):
Oops. Thanks for catching that.

PR #1456, opened 2021-04-21
Handle large error adjustments in ntp

sanderssj:
This PR adds the ability for the ntp klib to handle large errors in local vs ntp server time by resetting the 
rtc_offset, normally only set at boot, to an offset from the new time value. A large time reset will also
reset any current in-progress time adjustments. This change does not affect the monotonic time in
any way, so the only adjustments to existing timers is to update any based on the RTC with the difference
between the old and new offsets.  As a side effect of needing to adjust the timers, this also adds a way
to walk through the entries in a pqueue. Since ntp slewing is slow, the default threshold to reset the rtc
is any time difference greater than one minute.

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -15,6 +15,8 @@
 
 #define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
 
+#define NTP_CLOCK_RESET_THRESHOLD   60
+
 struct ntp_ts {
     u32 seconds;
     u32 fraction;
@@ -168,9 +170,9 @@ static void ntp_input(void *z, struct udp_pcb *pcb, struct pbuf *p,
     ntp.runtime_memcpy(&t2, &pkt->receive_ts, sizeof(t2));
     timestamp rtd = wallclock_now - origin - ntptime_diff(&t1, &t2);
     s64 offset = ntptime_to_timestamp(&t1) - wallclock_now + rtd / 2;
-    s64 sec = (offset < 0 ? (-offset) : offset)>>32;
-    if (sec > 60) {
-        ntp.clock_reset(ntptime_to_timestamp(&t2));
+    s64 sec = sec_from_timestamp(offset < 0 ? -offset : offset);
+    if (sec > NTP_CLOCK_RESET_THRESHOLD) {
+        ntp.clock_reset(wallclock_now + offset);
         ntp.last_offset = 0;
         ntp.last_raw = 0;
         success = true;
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -15,7 +15,7 @@
 
 #define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
 
-#define NTP_CLOCK_RESET_THRESHOLD   60
+#define NTP_RESET_THRESHOLD_MIN 60  /* 60 seconds */
 
 struct ntp_ts {
     u32 seconds;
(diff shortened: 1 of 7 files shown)


src/kernel/clock.c
```diff
@@ -35,3 +35,19 @@ void clock_adjust(timestamp wallclock_now, s64 temp_cal, timestamp sync_complete
     rtc_settimeofday(sec_from_timestamp(wallclock_now));
 }
 KLIB_EXPORT(clock_adjust);
+
+closure_function(0, 1, boolean, timer_id_rtc,
+                timer, t)
+{
+    return t->id == CLOCK_ID_REALTIME || t->id == CLOCK_ID_REALTIME_ALARM;
```
francescolavra, line 42: Should include also CLOCK_ID_REALTIME_COARSE

sanderssj: Thanks, got it

src/kernel/clock.c
```diff
@@ -35,3 +35,26 @@ void clock_adjust(timestamp wallclock_now, s64 temp_cal, timestamp sync_complete
     rtc_settimeofday(sec_from_timestamp(wallclock_now));
 }
 KLIB_EXPORT(clock_adjust);
+
+closure_function(0, 1, boolean, timer_id_rtc,
+                timer, t)
+{
+    switch (t->id) {
```
wjhun, line 42: Really a nit, but it seems like it would be more straightforward to do the timer adjustment right here, and the timer_adjust below replaced with the pqueue_walk followed by a pqueue_reorder. But maybe not if timer_adjust is likely to be used elsewhere.

klib/ntp.c
```diff
@@ -335,6 +348,16 @@ int init(void *md, klib_get_sym get_sym, klib_add_sym add_sym)
             ntp.pollmin = interval;
         }
     }
+    ntp.reset_threshold = 0;
+    value reset_thresh = get(root, sym_intern(ntp_reset_threshold, intern));
```
wjhun, line 352: These next several lines can now be simplified using get_u64.

src/kernel/clock.c
```diff
@@ -35,3 +35,26 @@ void clock_adjust(timestamp wallclock_now, s64 temp_cal, timestamp sync_complete
...
+        return false;
+    }
+}
+
+void clock_reset_rtc(timestamp wallclock_now)
+{
+    clock_debug("%s: now %T, wallclock_now %T\n",
+                __func__, now(CLOCK_ID_REALTIME), wallclock_now);
+    rtc_settimeofday(sec_from_timestamp(wallclock_now));
+    timer_adjust(runloop_timers, stack_closure(timer_id_rtc), wallclock_now - now(CLOCK_ID_REALTIME));
```
wjhun, line 57: It might be a little more accurate to call now prior to the rtc_settimeofday, as rtc_settimeofday will take up some time doing port I/O and re-reading the RTC (which could include spinning for a ready status). Sampling the clock after the RTC set will cause this additional time to be included in the delta as skew.

wjhun (approved):
Some minor comments, otherwise looks good to go.

src/kernel/clock.c
sanderssj: Good point!

src/kernel/clock.c
sanderssj: Yeah, it doesn't need to be this elaborate. Fixed, thanks!

klib/ntp.c
sanderssj: I'm not sure it's worth it in this case--get_u64 doesn't let you make a distinction between a missing value (valid here) vs malformed input (which should be reported as an error). Also, since this is a klib, it's another new symbol to pull in.

wjhun: Yes, that makes sense.

PR #1457, opened 2021-04-22
sigaltstack: do not validate ss->ss_sp if SS_DISABLE is specified

wjhun:
According to sigaltstack(2):

"To disable an existing stack, specify ss.ss_flags as SS_DISABLE.  In this case, the kernel ignores any other flags in ss.ss_flags and the remaining fields in ss."

The nanos implementation, however, was calling validate_user_memory on ss->ss_sp regardless of the presence of SS_DISABLE, which was causing an -EFAULT return on valid arguments. With this fix, sigaltstack only validates ss->ss_sp if SS_DISABLE is not set.


sanderssj (commented):
LGTM!

PR #1458, opened 2021-04-22
optimize for space when building stage2 bootloader

wjhun:
The stage2 build was previously being built with -O3, which could, on occasion, result in a stage2 image that exceeds 64kB. Sizes larger than 64kB are problematic for Xen-based VMs such as AWS t2 instances (for more detail on this issue, see #1383). Building the stage2 bootloader with -Os results in a much smaller executable (e.g. 46kB vs 64kB - a 27% reduction - on a Debian 9 / gcc 8.3.0 build).

Additionally, a bug is resolved in run64 (where stage2 transitions control to the 64-bit kernel) where %rsi, the second function argument in the ABI calling convention, which is used in init_service to detect parameters on direct loading (e.g. AWS firecracker), was not being zeroed.

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -135,7 +135,7 @@ DEFINES= \
 	-DKLIB \
 	-DMBEDTLS_USER_CONFIG_FILE=\"mbedtls_conf.h\" \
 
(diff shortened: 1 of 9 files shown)


sanderssj (commented):
LGTM!

PR #1459, opened 2021-04-23
Resume original thread after running blockq action

sanderssj:
This change removes some unnecessary thread_resume calls in poll.c and
adds a thread_resume with the original thread in blockq_apply_bi_locked.
This is necessary because a call to wake from the blockq in a syscall would
switch current to the waking thread and any references to current in the
original syscall (such as to set the return value) would use the new thread
instead of the old one and thus return the wrong value from a syscall.


wjhun (approved):
LGTM

PR #1461, opened 2021-04-26
support static configuration and DHCP on multiple network interfaces

wjhun:
While multiple network interfaces are detected and enabled by Nanos, static and dynamic IP configuration have been limited to the first interface. This adds support for configuration of any interface by moving its configuration into a root-level tuple under the interface name. For example, `en2:(ipaddr:192.168.43.2 netmask:255.255.255.0 gateway:192.168.43.1)` gives static configuration for the second virtio-net interface. Other supported attributes in the interface tuple are 'mtu', used to specify the interface MTU size, and 'default', for selecting an interface as the default egress interface when no route is found.

DHCP is run on detected interfaces that don't have static IP configuration.

To ease transition for ops and existing installations, interface configuration for en1 may still be specified on the root tuple, though this usage is now deprecated.

Note that multiple interfaces may be tested under QEMU by specifying "ENABLE_SECOND_IFACE=1" on the make commandline.

This should hopefully address configuration of the second interface in https://github.com/nanovms/ops/issues/1009.

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -336,7 +336,7 @@ STORAGE=	virtio-scsi
 STORAGE_BUS=	,bus=pci.2,addr=0x0
 NETWORK=	virtio-net
 NETWORK_BUS=	,bus=pci.3,addr=0x0
-
+NETWORK_BUS_2=  ,bus=pci.3,addr=0x1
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
LGTM

PR #1462, opened 2021-04-29
Network: add support for DHCPv6

francescolavra:
The DHCPv6 client is enabled on a given network interface if no static IPv6 address is provided for that interface in the manifest.
Depends on https://github.com/nanovms/lwip/pull/1, which adds stateful DHCPv6 support to lwIP.


wjhun (approved):
LGTM

PR #1464, opened 2021-05-01
EPOLL_CTL_ADD: fix setting of event mask and data

francescolavra:
When a file descriptor is added to an epoll file, the corresponding epollfd structure must be initialized with the user-supplied event mask and data.
The existing code was failing to do the above in cases where the file descriptor being added was already present in the fd vector of the epoll file. This change fixes the issue by defining a new function called reset_epollfd() that properly resets an already-allocated epollfd structure with new values for the event mask and data; this function is now used for both epoll_ctl(EPOLL_CTL_ADD) and poll().

### DIFF
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -121,6 +121,14 @@ define_closure_function(1, 0, void, epollfd_free,
     unix_cache_free(get_unix_heaps(), epollfd, efd);
 }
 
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
Looks good

PR #1467, opened 2021-05-04
Temporarily revert wrapping the root tuple

sanderssj:
This change reverts wrapping the root tuple because tfs code will call
destroy on the wrapped tuple and thus the root tuple when switching to
a new log and cleaning up.


wjhun (approved):

sanderssj:
I realized that settings like syscall debug and others require the wrapped_root notification to be activated from the manifest, so I added wrapped_root back in to handle only that.

PR #1469, opened 2021-05-04
connect(2): honor non-blocking flag

wjhun:
With this change, the connect syscall now supports non-blocking mode (SOCK_NONBLOCK/ O_NONBLOCK). The netsock runtime test verifies that an attempt to connect over the loopback device gives an EINPROGRESS result. It also checks that EPOLLOUT is subsequently reported on behalf of the socket fd once the connection is established.


sanderssj (commented):
LGTM!

francescolavra (commented):
Looks good to me

PR #1470, opened 2021-05-06
Add epoll custom edge trigger handling and customize eventfd handling

sanderssj:
Linux has different definitions for what constitutes an edge with EPOLLET in epoll,
which can vary by file descriptor type. This change allows each file type to implement
an optional handler to determine where edges are, otherwise the original edge tracking
code gets used. This change also implements a handler for eventfd, which has edges
for each read or write of an eventfd file, not just when the counter changes the
read or write ability.

### DIFF
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -14,6 +14,7 @@ declare_closure_struct(1, 0, void, epollfd_free,
 
 typedef struct epollfd {
     int fd;
+    fdesc f;
     u32 eventmask;  /* epoll events registered - XXX need lock */
     u32 lastevents; /* retain last received events; for edge trigger */
     u64 data;	    /* may be multiple versions of data? */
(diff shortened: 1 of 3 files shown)


src/unix/poll.c
```diff
@@ -273,9 +277,13 @@ static inline u32 report_from_notify_events(epollfd efd, u64 notify_events)
 
     /* catch falling edges for EPOLLET */
     if (edge_detect) {
-        u32 falling = efd->lastevents & ~events;
-        if (falling)
-            efd->lastevents &= ~falling;
+        if (efd->f->edge_trigger_handler) {
+            efd->lastevents = apply(efd->f->edge_trigger_handler, events, efd->lastevents);
+        } else {
+            u32 falling = efd->lastevents & ~events;
```
wjhun, line 283: Nit: comment above can move here

wjhun (approved):
LGTM!

src/unix/poll.c
sanderssj: Moved!

PR #1472, opened 2021-05-07
Add length assertions to basic buffer operations

sanderssj:
I've run into a couple situations where I got kernel page faults with strange addresses that were caused by moving
and accessing a buffer start pointer beyond the end or the length. This change adds assertions to enforce a
start<=end<=length invariant to avoid harder to debug crashes from out of bounds memory access. It also uses
a different macro name for assert so that klibs can skip the asserts which would require them to import at least
two additional symbols.

### DIFF
--- a/src/runtime/buffer.h
+++ b/src/runtime/buffer.h
@@ -10,8 +10,16 @@ struct buffer {
     void *contents;
 };
 
+/* don't force klibs import symbols for assert */
+#ifdef KLIB
+#define buffer_assert(x)
+#else
+#define buffer_assert(x) assert(x)
+#endif
+
 static inline void *buffer_ref(buffer b, bytes offset)
 {
+    buffer_assert(b->start + offset <= b->length);
     // alignment?
     return((void *)b->contents + (b->start + offset));
 }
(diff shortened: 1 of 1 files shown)


src/runtime/buffer.h
```diff
@@ -10,8 +10,16 @@ struct buffer {
...
+/* don't force klibs import symbols for assert */
+#ifdef KLIB
+#define buffer_assert(x)
+#else
+#define buffer_assert(x) assert(x)
+#endif
+
 static inline void *buffer_ref(buffer b, bytes offset)
 {
+    buffer_assert(b->start + offset <= b->length);
```
francescolavra, line 22: This should be `<` rather than `<=`

src/runtime/buffer.h
```diff
@@ -391,13 +405,16 @@ static inline u8 pop_u8(buffer b)
 
 static inline void push_u8(buffer b, u8 x)
 {
-    assert(buffer_extend(b, 1));
+    buffer_assert(buffer_extend(b, 1));
+    buffer_assert(b->start <= b->end);
+    buffer_assert(b->end <= b->length);
```
francescolavra, line 410: This should be `<` rather than `<=` (or alternatively this assert can be moved to the end of the function, after `b->end++`)

src/runtime/buffer.h
sanderssj: The problem here is that there is at least foreach_character macro which uses the comma operator in the test to use buffer_ref to set the next value before the test is performed. Let me see if I can rewrite the macro to avoid the buffer_ref if the test will fail.

sanderssj: I fixed the foreach_character, but I found another pattern used in a quite a few places is 'buffer_ref(b, buffer_length(b))' to get an end pointer for an iterator. I'm not sure what the best course of action is. Should we make a new method for getting an end pointer, or keep it as is where it could miss off-by-one errors at the end?

sanderssj: Sounds good

wjhun (approved):
LGTM

PR #1475, opened 2021-05-08
Use locked heap for queued_signal structs

sanderssj:
The queued_signal struct may be allocated outside of the kernel lock and
needs to use the locked heap instead of general.


francescolavra (commented):
LGTM

wjhun (approved):

PR #1476, opened 2021-05-10
TFS log: prevent log deallocation while a log extension is open

francescolavra:
A log extension keeps a pointer to the log structure, and needs to safely access this pointer. Thus, it should prevent the log from being deallocated as long as the extension is open, and not only during a flush to disk.
This change fixes a page fault that occurs during log compaction if the log being retired is deallocated (by the log_switch_complete() function) while its current extension is still being flushed to disk. This fault was caused by log_ext_free() accessing the heap member of an already deallocated log struct.
Since open_log_extension() now bumps the log refcount, the call to this function in log_new() is being moved to after the initialization of the log refcount.


sanderssj (commented):
LGTM!

wjhun (approved):
LGTM

PR #1478, opened 2021-05-10
x86 exception stack: change size from 32 KB to 64 KB

francescolavra:
The exception stack is being overflowed in the rprintf() function when executing an exception handler on x86, which means that when an unhandled fault occurs, the register dump and frame and stack traces that are supposed to be printed on the console are often replaced by those associated to this stack overflow, with a generic "frame xxxx already full" message that hides the original fault.
This commit increase the exception stack size from 32 KB to 64 KB, and in addition decreases the stack size requirements of the rprintf() function.


wjhun (approved):
Looks good

PR #1479, opened 2021-05-12
Use iterative unicast of discovered cpus for IPI broadcasts

sanderssj:
Using the "all but self" (TARGET_EXCLUSIVE_BROADCAST) address in an IPI on x86
leads to a significant performance penalty. The broadcast may target disabled
cpus or delay waiting for a response that doesn't come. This patch changes
the broadcast to an iterative approach that unicasts to discovered cpus.

Prior to this patch, the fs_full test took twice as long with a single cpu but SMP_ENABLED
as with a single cpu without it enabled. After the patch, the runtime is identical.

### DIFF
--- a/src/x86_64/apic.c
+++ b/src/x86_64/apic.c
@@ -41,9 +41,17 @@ static inline u32 apic_read(int reg)
 
 void apic_ipi(u32 target, u64 flags, u8 vector)
 {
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

Issue #1473, opened 2021-05-07
fs_full test sometimes fails on ARM with strange read result and a misaligned buffer error


sanderssj:
This failed in CI, but only once in quite a few CI runs. Failed at commit https://github.com/nanovms/nanos/commit/7b1bfa6692d30c78a6da6d6a9226561c8f997c54
```
MKFS	image
UUID: df8c5f0f-ebec-6018-181c-8874ec4a7f22
metadata
(children:(kernel:()))
UUID: 895b723d-4e6b-85c0-cf3f-341631692e0e
metadata
(arguments:(0:fs_full)
 children:(fs_full:())
 environment:(PWD:/
              USER:bobby)
 fault:t
 program:/fs_full)
qemu-system-aarch64 -machine virt -machine gic-version=2 -machine highmem=off -m 1G  -kernel /home/circleci/project/output/platform/virt/bin/kernel.img -display none  -serial stdio -drive if=none,id=hd0,format=raw,file=/home/circleci/project/output/image/disk.raw -device virtio-blk-pci,drive=hd0 -no-reboot -semihosting  -device virtio-net,netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostfwd=tcp::9090-:9090,hostfwd=udp::5309-:5309 -object filter-dump,id=filter0,netdev=n0,file=/tmp/nanos.pcap -cpu max
en1: assigned 10.0.2.15
total bytes: 121634816 free bytes: 120520704 blocksize: 512
en1: assigned FE80::5054:FF:FE12:3456
failed to read written block 40155 err 'Success'
fs_full: /home/circleci/project/test/runtime/fs_full.c:93: main: Assertion `statbuf.f_bfree < MAX_FREE_BYTES/BLOCKSIZE' failed.
signal 6 received by tid 1, errno 0, code -6
   core dump (unimplemented)
storage_rw_internal error: misaligned buf: 0xffffffff802b8228
storage_rw_internal error: write buffer not properly aligned
make[2]: *** [Makefile:337: run-noaccel] Error 127
make[2]: Leaving directory '/home/circleci/project/platform/virt'
make[1]: *** [Makefile:103: run-noaccel] Error 2
make[1]: Leaving directory '/home/circleci/project'
```

wjhun:
[...]

PR #1480, opened 2021-05-13
storage_rw_internal: remove gratuitous buffer alignment check

wjhun:
The "misaligned buf" error found in https://github.com/nanovms/nanos/issues/1473 results from a klog write of the klog.dump buffer (src/kernel/log.c) tripping an alignment check in virtio_storage.c:storage_rw_internal. This check and its comment seem dubious, however, and such an alignment requirement doesn't seem to be backed up by anything in the virtio spec. (Possibly the 16 byte virtqueue descriptor alignment was mistakenly interpreted as a buffer alignment.) Removing the check allows the write to continue with no apparent errors from qemu.


sanderssj (commented):
LGTM

PR #1481, opened 2021-05-13
extents_range_handler: properly handle extend failure when allocating

wjhun:
This PR resolves a crash that occurred while trying to reproduce https://github.com/nanovms/nanos/issues/1473 on a Raspberry Pi 4. If extents_range_handler was called during the allocation stage of a tfs write and a subsequent call to update_extent_length returned a failure, an apply_merge would be called with a null pointer. This changes the extend call to mirror fill_gap, returning a failure status instead of trying to apply it directly to a merge. Also, since the presence of a merge indicates extent writing is in process, eliminate the redundant write flag parameter in extent_range_handler, fill_gap and extend.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -569,7 +569,7 @@ static u64 write_extent(fsfile f, extent ex, sg_list sg, range blocks, merge m)
     return i.end;
 }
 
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

francescolavra (commented):
LGTM

Issue #1483, opened 2021-05-13
virtio-blk appears to suffer from poor write performance


wjhun:
https://github.com/nanovms/nanos/issues/1473#issuecomment-840823845

wjhun:
Resolved with #1485

PR #1485, opened 2021-05-14
add flush method to storage interfaces with implementations for virtio_storage and virtio_scsi

### LINKED ISSUE #1484: add flush method to storage devices
Without a flush method on block devices, we can't be certain that data written to storage is consistent after a sync or shutdown.

wjhun:
This extends the block device interface to add a "flush" method. A flush is issued as a final step in the chain of operations initiated by filesystem_flush, used by the various paths that synchronize storage devices. Flush implementations are included for virtio-storage (which issues a VIRTIO_BLK_T_FLUSH command) and virtio-scsi (which issues a SYNCHRONIZE_CACHE SCSI command) drivers; other storage drivers will require flush implementations to be added later.

As part of the virtio-storage implementation, the VIRTIO_BLK_F_FLUSH and VIRTIO_BLK_F_CONFIG_WCE features are negotiated. A flush method is created and writeback cache mode is set depending on negotiation of respective features. This improves the virtio-storage performance drastically, which now out-performs the virtio-scsi driver in the bulk write test by over 10% in one Linux/KVM-based test setup. This speedup is to be expected, as noted in the OASIS virtio spec, because the virtio-blk device involves much less overhead than the virtio-scsi device.

(Note that the overall performance for the write test will appear to drop slightly when compared to that of previous commits. This is because the storage flush operation is now included as part of the measured duration.)

Resolves #1484, #1483 and alleviates the symptoms found in #1473 (although another factor requires more work on tlog).

### DIFF
--- a/platform/pc/boot/stage2.c
+++ b/platform/pc/boot/stage2.c
@@ -282,6 +282,7 @@ void newstack()
                       infinity,
                       get_stage2_disk_read(h, fs_offset),
                       closure(h, stage2_empty_write),
+                      0 /* no flush */,
                       false,
                       closure(h, filesystem_initialized, h, backed, bh));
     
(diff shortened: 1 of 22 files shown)


sanderssj (commented):
LGTM

PR #1487, opened 2021-05-16
TCP receive: poll loopback interface after calling tcp_recved()

francescolavra:
When a high amount of TCP data is sent to the loopback interface, the TCP transmit window length may become zero, which blocks the transmitting thread. When later the receiving thread reads the incoming data from the interface, the transmit window length becomes non-zero again as a consequence of the receiving thread calling tcp_recved(), but in order for this to be detected by the transmitting thread, the loopback interface needs to be polled.
This PR adds a call to netsock_check_loop() in sock_read_bh_internal(), so that any thread blocked while transmitting to the loopback interfce can be unblocked.
The netsock runtime tests now include a new test that would block indefinitely if the above issue was not addressed.


wjhun (approved):
LGTM

PR #1489, opened 2021-05-18
Add option to print out names of files that are not found

sanderssj:
A manifest option 'missing_files' has been added that will keep a list of
filenames that are not found by open_internal and then print out the list
when nanos exits. This can be used to help find dynamic libraries or etc
files that need to be included in the filesystem for the application to
run properly. File names only (not path) are added to the list so that if
a file is opened later with a different path it gets removed from the missing
files list. This change also slightly refactors the way shutdown completions
are added as well as adds a couple of new buffer helper functions for fs
path processing.

### DIFF
--- a/src/runtime/buffer.h
+++ b/src/runtime/buffer.h
@@ -361,6 +361,15 @@ static inline int buffer_strchr(buffer b, int c)
     return -1;
 }
 
+static inline int buffer_strrchr(buffer b, int c)
+{
+    for (s64 len = buffer_length(b) - 1; len >= 0; len--) {
+        if (byte(b, len) == c)
+            return len;
+    }
+    return -1;
+}
+
 int buffer_strstr(buffer b, const char *str);
 
 // the ascii subset..utf8 me
@@ -469,3 +478,21 @@ static inline key fnv64(void *z)
 }
 
 void buffer_print(buffer b);
+
+/* modifies the original buffer, buffer must be a string */
+static inline buffer buffer_basename(buffer b)
+{
+    int p;
+
+    if (buffer_length(b) <= 1)
+        return b;
+    if ((p = buffer_strrchr(b, '/')) != -1) {
+        if (p == buffer_length(b) - 1) {
+            b->end--;
+            if ((p = buffer_strrchr(b, '/')) == -1 || buffer_length(b) == 1)
+                return b;
+        }
+        buffer_consume(b, p + 1);
+    }
+    return b;
+}
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -21,12 +21,15 @@ typedef struct syscall_stat {
 
 static struct syscall_stat stats[SYS_MAX];
 boolean do_syscall_stats;
+static boolean do_missing_files;
+static vector missing_files;
 
 sysreturn close(int fd);
 
 io_completion syscall_io_complete;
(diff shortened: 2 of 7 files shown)


src/runtime/buffer.h
```diff
@@ -361,6 +361,16 @@ static inline int buffer_strchr(buffer b, int c)
     return -1;
 }
 
+static inline int buffer_strrchr(buffer b, int c)
+{
+    bytes len = buffer_length(b);
+    for (bytes i = 0; i < len; i++) {
```
wjhun, line 367: Total nit, but could just loop from len - 1 down to zero

src/unix/syscall.c
```diff
@@ -834,6 +857,17 @@ sysreturn open_internal(filesystem fs, tuple cwd, const char *name, int flags,
         }
     }
 
+    if (do_missing_files) {
```
wjhun, line 860: I gather that this is a feature that would only be switched on in the rare case that some libraries / support files can't be found, and thus does not need to be efficient in any way, correct? If it were ever enabled by default, this implementation might (?) cause a significant performance penalty, given that this is adding an entry for every file - whether found or not - and then removing it from the missing_files vector for found files. Also, this implementation requires a linear search (vs, for example, a table-based implemenation).

Is it possible to make this add files only on the condition that a file is not found? Or is it written this way because a file might be not found and then later created?

src/unix/syscall.c
```diff
@@ -2433,15 +2467,47 @@ static void syscall_schedule(context f)
     syscall_debug(f);
 }
 
-void init_syscalls()
+static char *missing_files_exclude[] = {
+    "ld.so.cache",
```
wjhun, line 2471: Just curious - why is it necessary to exclude ld.so.cache - or any files for that matter?

wjhun (approved):

src/unix/syscall.c
sanderssj: You are correct, it's only intended to be a debugging feature. I had to implement like this because glibc tries several different paths for each dylib (e.g. /lib/x86_64-linux-gnu, /lib/x86_64-linux-gnu/x86_64, /usr/lib/x86_64-linux-gnu, etc) so if a file isn't in the first path then it's added to the list and if it's found later in a different directory then it's not really missing. It would definitely be more efficient to use a table. I originally thought it would be useful to preserve the order in which files are attempted to be open, but now I've tried it a few times I don't know that order is really that important.

src/unix/syscall.c
sanderssj: The intent was to reduce confusion for a user who may not understand what a file does, especially since ld.so.cache is generated for the host and I was concerned it might be a source of more trouble if a copied ld.so.cache had references to files or paths which did not exist on the nanos image. I'm happy to take it out if you think it's better to just show everything.

src/unix/syscall.c
wjhun: Yes, this makes sense and it looks good as it is. I don't think there's any need to change it to make a debug-only feature slightly more efficient.

src/unix/syscall.c
wjhun: No need to change - just curious!

src/runtime/buffer.h
sanderssj: I simplified it!

Issue #1299, opened 2020-09-28
investigate zero-copy socket writes


wjhun:
From a cursory look it appears that we could potentially implement zero-copy on socket writes by eliminating the TCP_WRITE_FLAG_COPY flag on calls to tcp_write when SO_ZEROCOPY / MSG_ZEROCOPY is specified. User pages not under the domain of the pagecache are, in a sense, pinned already, and pages within the pagecache could be pinned by taking an extra refcount on the pagecache_page. Implementation of socket error queues would also be necessary to allow completion notification to the application.

This could potentially yield a significant performance benefit in cases such as large static page loads when the service supports zero copy (which requires that user buffers remain unmodified until after sent TCP data is acknowledged), but some further exploration might be necessary to verify that in fact the zero copy path - from lwIP through our existing PV nic drivers - will work as expected. Furthermore, note that SO_ZEROCOPY is a hint to the kernel to use zero-copy if available - with a guarantee that completion notifications will be returned - and not a guarantee that copying will be avoided (so a non-compliant driver could result in use of TCP_WRITE_FLAG_COPY with completion notifications).

https://www.kernel.org/doc/html/v4.15/networking/msg_zerocopy.html
https://blogs.oracle.com/linux/zero-copy-networking-in-uek6

francescolavra:
Did some testing without the TCP_WRITE_FLAG_COPY flag on calls to tcp_write. The zero copy path from lwIP through the nic driver does work as expected (at least with the virtio net driver), in the sense that data from the user buffer are correctly sent to the nic, but doesn't result in an overall performance gain; in fact, I could see a slight degradation in performance (in the order of a few percentage points); and that is without the socket error queue messaging, which when added would likely contribute to a decrease in performance.
The reason for the zero-copy path not bringing performance benefits is that the savings from avoiding memory copying are outweighed by the overhead associated with handling an additional buffer in each network packet: when data is copied, each network packet can be sent as a single contiguous buffer, but when the data is not copied, the packet headers need to be allocated in a separate buffer and then chained to the user data buffer. At the nic driver level, the physical address needs to be retrieved for each buffer of a network packet, and this is relatively expensive.

Below is an annotated ftrace plot obtained when streaming TCP data from Nanos to the host machine, without zero-copy:
![image](https://user-images.githubusercontent.com/10710577/118389145-b960ad80-b628-11eb-94ea-44e3b410b155.png)
[...]

PR #1492, opened 2021-05-21
lwIP performance optimization

francescolavra:
This PR changes a few lwIP configuration options, following the hints at https://www.nongnu.org/lwip/2_1_x/optimization.html.
The lwip_htons() and lwip_htonl() macros have been defined so that byte order inversion operations are executed inline instead of as function calls (in #1299, the lwip_htons() function shows up as taking a non-negligible share of CPU time).
Instead of the default checksum algorithm # 2, algorithm # 3 is now being used, because it is faster on 64-bit platforms (checksum calculation on 20-byte data is around 35% faster).

### DIFF
--- a/src/net/lwipopts.h
+++ b/src/net/lwipopts.h
@@ -48,6 +48,8 @@
 #define LWIP_NO_LIMITS_H 1
 #define LWIP_NO_CTYPE_H 1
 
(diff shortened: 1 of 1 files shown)


sanderssj:
LGTM

wjhun (approved):

PR #1495, opened 2021-06-01
Add tun network interface implementation

francescolavra:
The tun interface driver is implemented as a klib.
In order to allow adding a new type of special files and implementing blocking I/O on this file, I/O operations on special files have been refactored using custom fdesc closures, instead of relying on the fdesc closures used for regular files.

### DIFF
--- a/klib/tun.c
+++ b/klib/tun.c
@@ -1,0 +1,349 @@
+#include <unix_internal.h>
+#include <lwip.h>
+#include <socket.h>
+
+/* ioctl requests */
+#define TUNSETIFF   0x400454ca
+#define TUNGETIFF   0x800454d2
+
+/* ifreq flags */
+#define IFF_TUN         0x0001
+#define TUN_TYPE_MASK   0x000f
+#define IFF_NO_PI       0x1000
+
+/* Packet information flags */
+#define TUN_PKT_STRIP   0x0001
+
+#define TUN_QUEUE_LEN   512
+
+typedef struct tun_pi { /* packet information */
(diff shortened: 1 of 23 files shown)


sanderssj (commented):
LGTM

klib/tun.c
```diff
@@ -0,0 +1,344 @@
...
+closure_function(1, 6, sysreturn, tun_write,
+                 tun, tun,
+                 void *, src, u64, len, u64, offset, thread, t, boolean, bh, io_completion, completion)
+{
+    tun tun = bound(tun);
+    if (!tun->netif)
+        return io_complete(completion, t, -EBADFD);
+    if (!(tun->flags & IFF_NO_PI)) {
+        if (len < sizeof(struct tun_pi))
+            return io_complete(completion, t, -EINVAL);
```
wjhun, line 151: Should len == sizeof(struct tun_pi) just return io_complete(completion, t, len) here?

wjhun (approved):
Everything looks good...just one question about pi write with zero len payload

klib/tun.c
francescolavra: Indeed, calling pbuf_alloc() with len=0 is wrong. Fixed.
If the IFF_NO_PI flag is not present, a write() call with len == sizeof(struct tun_pi) returns sizeof(struct tun_pi), while if the IFF_NO_PI flag is present, a write() call with len == 0 returns -EINVAL. This matches the behavior of the Linux kernel.

PR #1496, opened 2021-06-03
TFS log_read_complete(): avoid multiple stack closure allocations

francescolavra:
Allocating the log_read_ingest_extent closure for each extent found in a given filesystem is not necessary and can lead to a stack
overflow when reading a filesystem with a large number of extents.
This PR moves the closure allocation outside the loop that is executed for each extent, and uses the closure_member() macro inside the loop to update the fsfile closure member.


wjhun (approved):
LGTM

PR #1497, opened 2021-06-07
Add syslog klib

francescolavra:
This klib implements a console driver, which can be configured via the manifest to output console messages to a file (with a simple log rotation algorithm) and/or to an UDP syslog server.
The `syslog` manifest tuple is used to configure syslog behavior.
Supported configuration options:
- file: if present, syslog output is written to a file (whose path is indicated by the value of this option)
- file_max_size: file size at which log rotation occurs (default value: 8 MB); its value contains a byte count, optionally followed by a multiplier suffix ('k' or 'K' for KB, 'm' or 'M' for MB, 'g or 'G' for GB)
- file_rotate: number of rotated files to keep in the filesystem (default value: 1); if 0, old logs are deleted instead of being rotated
- server: if present, syslog output is sent to an UDP server (whose hostname or IP address is indicated by the value of this option)
- server port: UDP port number (default value: 514) at which syslog output is sent

Example manifest contents for file output:
```
syslog:(
  file:var/log/syslog
)
```
Example manifest contents for UDP output:
```
syslog:(
  server:example.com
)
```

To make log rotation actually useful, the filesystem_rename() function has been fixed to deallocate the extents associated to a file being overwritten, otherwise the storage space occupied by an overwritten file is leaked.

The console_write() function has been made slightly more efficient by sorting the attached console drivers based on their enabled status.

### DIFF
--- a/klib/syslog.c
+++ b/klib/syslog.c
@@ -1,0 +1,460 @@
+#include <unix_internal.h>
+#include <drivers/console.h>
+#include <filesystem.h>
+#include <lwip.h>
+#include <mktime.h>
+
+#define __STRING(x)     #x
+#define __XSTRING(x)    __STRING(x)
+
+#define SYSLOG_BUF_LEN  (8 * KB)
+
+#define SYSLOG_FLUSH_INTERVAL   seconds(1)
+
+#define SYSLOG_FILE_MAXSIZE_DEFAULT (8 * MB)
+
+#define SYSLOG_FILE_ROTATE_DEFAULT  1
+#define SYSLOG_FILE_ROTATE_MAX      9
+
+/* facility: 1 (user-level messages)
+ * severity: 6 (informational messages)
+ * priority = facility * 8 + severity
+ */
+#define SYSLOG_PRIORITY 14
+
+#define SYSLOG_VERSION  "1"
+
+#define SYSLOG_UDP_PORT_DEFAULT 514
+
+declare_closure_struct(0, 1, void, syslog_timer_func,
+    u64, overruns);
+
+#define syslog_lock()   u64 _irqflags = spin_lock_irq(&syslog.lock)
+#define syslog_unlock() spin_unlock_irq(&syslog.lock, _irqflags)
+
+static struct {
+    struct console_driver driver;
+    heap h;
+    buffer file_path;
+    u64 file_max_size;
+    u64 file_rotate;
+    fsfile fsf;
+    sg_io fs_write;
+    u64 file_offset;
+    sg_list file_sg;
+    sg_buf file_sgb;
+    timer file_flush_timer;
+    closure_struct(syslog_timer_func, file_flush);
+    buffer program;
+    char *server;
+    boolean dns_in_progress;
+    timestamp dns_backoff;
+    timestamp dns_req_next;
+    ip_addr_t server_ip;
+    u16 server_port;
+    struct udp_pcb *udp_pcb;
+    char local_ip[40];
+    bytes hdr_len;
+    struct spinlock lock;
+} syslog;
+
+static struct {
+    fsfile (*fsfile_open_or_create)(buffer file_path);
+    sg_io (*fsfile_get_writer)(fsfile f);
+    fs_status (*fsfile_truncate)(fsfile f, u64 len);
+    sg_list (*allocate_sg_list)(void);
+    sg_buf (*sg_list_tail_add)(sg_list sg, word length);
+    void (*deallocate_sg_list)(sg_list sg);
+    void (*runtime_memcpy)(void *a, const void *b, bytes len);
+    timer (*register_timer)(clock_id id, timestamp val, boolean absolute, timestamp interval,
+            timer_handler n);
+    void (*timm_dealloc)(tuple t);
+    sysreturn (*fs_rename)(buffer oldpath, buffer newpath);
+    err_t (*dns_gethostbyname)(const char *hostname, ip_addr_t *addr,
+            dns_found_callback found, void *callback_arg);
+    struct netif *(*netif_get_default)(void);
+    char *(*ipaddr_ntoa_r)(const ip_addr_t *addr, char *buf, int buflen);
+    struct pbuf *(*pbuf_alloc)(pbuf_layer layer, u16_t length, pbuf_type type);
+    u8 (*pbuf_free)(struct pbuf *p);
+    timestamp (*now)(clock_id id);
+    struct tm *(*gmtime_r)(u64 *timep, struct tm *result);
+    int (*rsnprintf)(char *str, u64 size, const char *fmt, ...);
+    err_t(*udp_sendto)(struct udp_pcb *pcb, struct pbuf *p,
+            const ip_addr_t *dst_ip, u16_t dst_port);
+} kfuncs;
+
+static void syslog_file_rotate(void)
+{
+    bytes path_len = buffer_length(syslog.file_path);
+    buffer old_file = alloca_wrap_buffer(stack_allocate(path_len + 2), path_len + 2);
+    buffer new_file = alloca_wrap_buffer(stack_allocate(path_len + 2), path_len + 2);
+    kfuncs.runtime_memcpy(buffer_ref(old_file, 0), buffer_ref(syslog.file_path, 0), path_len);
+    kfuncs.runtime_memcpy(buffer_ref(new_file, 0), buffer_ref(syslog.file_path, 0), path_len);
+    byte(old_file, path_len) = byte(new_file, path_len) = '.';
+
+    /* Rename rotated log files by replacing the ".<n-1>" name extension with ".<n>". */
+    for (u64 i = syslog.file_rotate; i > 1; i--) {
+        byte(new_file, path_len + 1) = '0' + i - 1;
+        byte(old_file, path_len + 1) = '0' + i;
+        kfuncs.fs_rename(new_file, old_file);
+    }
+
+    /* Rename the current log file by adding a ".1" extension to the file name. */
+    byte(old_file, path_len + 1) = '1';
+    sysreturn ret = kfuncs.fs_rename(syslog.file_path, old_file);
+
+    if (ret == 0) {
+        /* Continue logging on a new file. */
+        fsfile file = kfuncs.fsfile_open_or_create(syslog.file_path);
+        syslog.file_offset = 0;
+        syslog.fs_write = kfuncs.fsfile_get_writer(file);
+    } else {
+        syslog.fs_write = 0;    /* stop logging */
+    }
+}
+
+closure_function(3, 1, void, syslog_file_write_complete,
+                 void *, buf, u64, len, sg_list, sg,
+                 status, s)
+{
+    if (is_ok(s)) {
+        fetch_and_add(&syslog.file_offset, bound(len));
+        if (syslog.file_offset >= syslog.file_max_size) {
+            /* Avoid deadlock in case a write completion occurs synchronously. */
+            boolean unlock = spin_try(&syslog.lock);
+
+            if (syslog.file_rotate > 0) {
+                syslog_file_rotate();
+            } else {
+                /* Delete old logs instead of rotating. */
+                if (kfuncs.fsfile_truncate(syslog.fsf, 0) == FS_STATUS_OK)
+                    syslog.file_offset = 0;
+                else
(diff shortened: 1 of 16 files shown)


sanderssj (commented):
Looks good to me! I only had one question on whether or not the udp port should be configurable rather than a compiled constant.

src/drivers/console.c
```diff
@@ -16,18 +16,56 @@ struct console_driver serial_console_driver = {
...
 };
 
-struct console_driver *console_drivers[4] = {
+struct console_driver *console_drivers[8] = {
     &serial_console_driver,
 };
 
 static struct spinlock write_lock;
 
+static void sort_console_drivers(void)
```
wjhun, line 25: This is not critical, but I think it would be simpler to keep the console_drivers on a list, with list heads for both enabled and disabled. Or use a single list and always move disabled drivers to the end (and move to head on enable). Then use list_foreach in console_write and break once a disabled entry is found.

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+    buffer output = get(root, sym(syslog_output));
+    if (!output) {
+        rprintf("syslog output not specified\n");
+        return KLIB_INIT_FAILED;
+    } else if (!is_string(output)) {
+        rprintf("invalid syslog output\n");
+        return KLIB_INIT_FAILED;
+    }
+    syslog.h = heap_general(get_kernel_heaps());
+    if ((buffer_length(output) == 4) && !runtime_memcmp(buffer_ref(output, 0), "file", 4)) {
```
wjhun, line 282: Can use buffer_compare_with_cstring for these conditionals

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+        if (!file) {
+            rprintf("cannot create syslog output file\n");
+            return KLIB_INIT_FAILED;
+        }
+        syslog.fs_write = kfuncs.fsfile_get_writer(file);
+        syslog.file_offset = fsfile_get_length(file);   /* append to existing contents */
+        syslog.file_flush_timer = INVALID_ADDRESS;
+        init_closure(&syslog.file_flush, syslog_timer_func);
+        syslog.driver.write = syslog_file_write;
+    } else if ((buffer_length(output) == 3) && !runtime_memcmp(buffer_ref(output, 0), "udp", 3)) {
```
wjhun, line 301: Is there any particular reason syslog output has to be either file or udp but not both?

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+            buffer archive_file = stack_allocate(sizeof(struct buffer));
+            bytes file_path_len = buffer_length(syslog.output_file);
+            archive_file->contents = stack_allocate(file_path_len + 2);
+            archive_file->start = 0;
+            kfuncs.runtime_memcpy(archive_file->contents, buffer_ref(syslog.output_file, 0),
+                file_path_len);
+            byte(archive_file, file_path_len) = '.';
+            byte(archive_file, file_path_len + 1) = '1';
+            archive_file->end = archive_file->length = file_path_len + 2;
+            boolean unlock = spin_try(&syslog.lock);
```
wjhun, line 94: I guess this is a try in case a write completion occurs synchronously?

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+    syslog.hdr_len = 1 + sizeof(__XSTRING(SYSLOG_PRIORITY)) + 2 +
+            sizeof("YYYY-MM-ddThh:mm:ss.uuuuuuZ") + runtime_strlen(syslog.local_ip) + 1 +
+            buffer_length(syslog.program) + 5;
+}
+
+static void syslog_udp_write(void *d, const char *s, bytes count)
+{
+    if (ip_addr_isany_val(syslog.server_ip)) {
+        if (!syslog.dns_in_progress)
+            syslog_server_resolve();
```
wjhun, line 211: I think some handling of DNS resolution failures would be in order here. Or, at least, some sort of exponential backoff on initiating requests, as it seems that an immediate failure / completion on a lookup could result in possibly many more requests being initiated if there is significant console output.

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+            !(kfuncs.pbuf_free = get_sym("pbuf_free")) ||
+            !(kfuncs.now = get_sym("now")) ||
+            !(kfuncs.gmtime_r = get_sym("gmtime_r")) ||
+            !(kfuncs.rsnprintf = get_sym("rsnprintf")) ||
+            !(kfuncs.udp_sendto = get_sym("udp_sendto"))) {
+        rprintf("syslog: kernel symbols not found\n");
+        return KLIB_INIT_FAILED;
+    }
+    tuple root = get_root_tuple();
+    buffer output = get(root, sym(syslog_output));
```
wjhun, line 273: can use get_string unless you prefer the distinct error messages below

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+
+static void syslog_file_write(void *d, const char *s, bytes count)
+{
+    if (!syslog.fs_write || (count > SYSLOG_BUF_LEN))
+        return;
+    if (!syslog.file_sg) {
+        syslog.file_sg = kfuncs.allocate_sg_list();
+        if (syslog.file_sg == INVALID_ADDRESS)
+            return;
+        syslog.file_sgb = kfuncs.sg_list_tail_add(syslog.file_sg, SYSLOG_BUF_LEN);
```
wjhun, line 150: This isn't critical, and if it even makes sense to do it could be done in another PR, but I would suggest doing this kind of buffering on the front end of the syslog console write. There are a few reasons for this. There is quite a lot of work to do for every console write, and In situations where there are frequent and especially short writes, this overhead could add up and have a performance impact. With buffering on the front end, the actual file and/or udp write could be deferred somewhat by a timer, reducing the latency of the console writes and amortizing the cost of the write operation over a larger amount of output. It would also reduce the number of file writes / udp frames sent for a given amount of output if the output is bursty. Also, allowance for a backlog (up to a configurable limit) would allow the capture of output that occurs before the default interface is up and name resolution has finished.

wjhun (approved):
Looks good - mostly just comments about minor and future improvements

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+    timestamp t = kfuncs.now(CLOCK_ID_REALTIME);
+    u64 seconds = sec_from_timestamp(t);
+    struct tm tm;
+    kfuncs.gmtime_r(&seconds, &tm);
+    kfuncs.rsnprintf(pb->payload, syslog.hdr_len + 1,
+        "<" __XSTRING(SYSLOG_PRIORITY) ">1 %d-%02d-%02dT%02d:%02d:%02d.%06dZ %s %b - - - ",
+        1900 + tm.tm_year, 1 + tm.tm_mon, tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec,
+        usec_from_timestamp(t) - seconds * MILLION, syslog.local_ip, syslog.program);
+    kfuncs.runtime_memcpy(pb->payload + syslog.hdr_len, s, count);
+    kfuncs.udp_sendto(syslog.udp_pcb, pb, &syslog.server_ip, SYSLOG_UDP_PORT);
```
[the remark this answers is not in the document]
francescolavra: OK, I made the UDP port configurable, via the `syslog_port` manifest option

src/drivers/console.c
francescolavra: OK, I changed the console code to use a single list, with disabled drivers at the end.

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+closure_function(3, 1, void, syslog_file_write_complete,
+                 void *, buf, u64, len, sg_list, sg,
+                 status, s)
+{
+    if (is_ok(s)) {
+        fetch_and_add(&syslog.file_offset, bound(len));
+        if (syslog.file_offset >= SYSLOG_FILE_MAXSIZE) {
+            /* Do log rotation: rename the current log file by adding a ".1" extension to the file
+             * name, and continue logging on a new file. */
+            buffer archive_file = stack_allocate(sizeof(struct buffer));
```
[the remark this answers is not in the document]
francescolavra: alloca_wrap_buffer cannot be used because we have to append a suffix to the file name, so we would be writing past the end of the original buffer contents.

klib/syslog.c
```diff
@@ -0,0 +1,338 @@
...
+            bytes file_path_len = buffer_length(syslog.output_file);
+            archive_file->contents = stack_allocate(file_path_len + 2);
+            archive_file->start = 0;
+            kfuncs.runtime_memcpy(archive_file->contents, buffer_ref(syslog.output_file, 0),
+                file_path_len);
+            byte(archive_file, file_path_len) = '.';
+            byte(archive_file, file_path_len + 1) = '1';
```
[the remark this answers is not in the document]
francescolavra: Not overwriting an existing .1 file would defeat the main purpose of log rotation, which is to avoid filling up the disk with logs. Or did I misunderstand your comment?

klib/syslog.c
francescolavra: I changed the code so that file and udp output ca be enabled at the same time. Now, the `syslog_output` manifest option takes as value an array, e.g. `[udp]` to enable udp output and `[file udp]` to enable both file and udp output.

klib/syslog.c
francescolavra: Yes, spin_try() is there in case a write completion occurs synchronously. I added a comment to the code

klib/syslog.c
francescolavra: OK, I added an exponential backoff to handle DNS failures.

klib/syslog.c
francescolavra: get_string() cannot be used from klib code, because it's an inline function that calls a non-inline function.

klib/syslog.c
francescolavra: The syslog protocol requires each UDP packet to convey exactly one message, so we can't group different messages in a single packet (well, we could do it, but I assume most users would expect each syslog message to correspond to a printf() or similar function call in their program, especially if they are interested in the timestamp associated to each message). So I think buffering shouldn't be done for UDP.

klib/syslog.c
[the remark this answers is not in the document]
wjhun: alloca_wrap_buffer does the same thing you're doing here...it's just eliminating a few lines of code. You're only overwriting the end of the buffer contents, not actually appending. So something like alloca_wrap_buffer(stack_allocate(file_path_len + 2), file_path_len + 2) and then just the runtime_memcpy and writing of last two chars should be equivalent.

klib/syslog.c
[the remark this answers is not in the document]
wjhun: You can still have log rotation with >2 files; see logrotate. My only issue with this is that the amount of log data that can be buffered on disk is fixed at 16MB. logrotate allows both the maximum number of files (rotate parameter) and log size (size parameter) to be specified. I think it would be good to support the configuration of either (or both) to allow the user to control how much syslog data gets saved to disk.

klib/syslog.c
wjhun: I see. I guess the only real advantages to buffering, then, (while still sending one packet per queued message) would be to gather the UDP writes together as a deferred operation rather than going through the whole stack on each console output, thereby reducing the latency of writes, and also to be able to maintain some degree of backlog to capture early messages before the server becomes reachable.

klib/syslog.c
wjhun: Is syslog_output even necessary? The presence of either syslog_file or syslog_server should be sufficient to indicate which outputs are being used.

wjhun: Don't mean to be a nuisance, but maybe it would be preferable to have just a "syslog" tuple that has one or more of "file", "server" and "port" attributes.

francescolavra: Ok, makes sense, I will change the code to use a single syslog tuple with one or more attributes, among which there can be log rotation parameters

francescolavra: Configuration settings for the syslog klib are now grouped in a `syslog` manifest option, as explained in the updated description of this PR.

klib/syslog.c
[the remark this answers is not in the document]
francescolavra: Now log rotation behavior can be customized with the `file_max_size` and `file_rotate` configuration settings, that have the same semantics as the `size` and `rotate` parameters of the logrotate Linux program.

PR #1500, opened 2021-06-10
add "cwd" manifest option for setting current working directory

### LINKED ISSUE #1498: allow CWD env var or manifest option to be set before exec
this is a little bit related to https://github.com/nanovms/ops/issues/1076

wjhun:
The current working directory for an application may now be set by including a
path string under the "cwd" attribute in the program manifest. A test for this
feature is added to the mkdir runtime test. To allow parity testing with
Linux, the mkdir test has also been updated to use a test directory under
/tmp (rather than the root directory) and create files with legitimate
permissions.

Resolves #1498

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -27,6 +27,28 @@ static inline void report_sha256(buffer b)
 #define report_sha256(b)
 #endif
 
+const char *string_from_fs_status(fs_status s)
+{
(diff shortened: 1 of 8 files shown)


sanderssj (commented):
LGTM!

francescolavra (commented):
LGTM

PR #1501, opened 2021-06-14
Fix various unix socket bugs and behaviors

sanderssj:
This change fixes a number of bugs in the AF_UNIX family of sockets and in
general brings unix socket behavior more in lin with Linux behavior. The
biggest part of the change fixes the handling of SOCK_DGRAM unix
sockets: the source address is now carried in the sharedbuf struct so
that sendto and recvfrom could be implemented, listen/accept were updated
to not support datagrams, connect was updated to only stash the destination
and return for datagram type, and a SOCK_DGRAM socket can now have
datagrams from multiple other sockets. Other fixes include matching event
notifications to more closely imitate Linux, update some errno returns
to more closely imitate Linux, no overwriting existing flags on accept
and check valid flags on accept, and fix the handling of the connecting
state which caused some socket functions to erroneously return errors.

This change also updates the existing unixsocket runtime test to allow it
to run on Linux and reflect Linux behaviors in places. It also adds the
unixsocket test to the set of default runtime tests.

### DIFF
--- a/src/unix/socket.c
+++ b/src/unix/socket.c
@@ -6,19 +6,20 @@
 declare_closure_struct(1, 0, void, sharedbuf_free,
     struct sharedbuf *, shb);
 
-typedef struct sharedbuf {
-    buffer b;
-    struct refcount refcount;
-    closure_struct(sharedbuf_free, free);
-} *sharedbuf;
-
-#define UNIXSOCK_BUF_MAX_SIZE   PAGESIZE
-#define UNIXSOCK_QUEUE_MAX_LEN  64
-
 struct sockaddr_un {
     u16 sun_family;
     char sun_path[108];
 };
+
+typedef struct sharedbuf {
+    buffer b;
+    struct refcount refcount;
+    closure_struct(sharedbuf_free, free);
+    struct sockaddr_un from_addr;
+} *sharedbuf;
+
+#define UNIXSOCK_BUF_MAX_SIZE   PAGESIZE
+#define UNIXSOCK_QUEUE_MAX_LEN  64
 
 typedef struct unixsock {
     struct sock sock; /* must be first */
@@ -105,8 +106,8 @@ static inline void unixsock_notify_writer(unixsock s)
     notify_dispatch(s->sock.f.ns, EPOLLOUT);
 }
 
-closure_function(6, 1, sysreturn, unixsock_read_bh,
-                 unixsock, s, thread, t, void *, dest, sg_list, sg, u64, length, io_completion, completion,
+closure_function(8, 1, sysreturn, unixsock_read_bh,
+                 unixsock, s, thread, t, void *, dest, sg_list, sg, u64, length, io_completion, completion, struct sockaddr_un *, from_addr, socklen_t *, from_length,
                  u64, flags)
 {
     unixsock s = bound(s);
@@ -122,11 +123,10 @@ closure_function(6, 1, sysreturn, unixsock_read_bh,
     }
     shb = queue_peek(s->data);
     if (shb == INVALID_ADDRESS) {
-        if (!s->peer) {
+        if (s->sock.type == SOCK_STREAM && !s->peer) {
             rv = 0;
             goto out;
-        }
-        else if (s->sock.f.flags & SOCK_NONBLOCK) {
+        } else if (s->sock.f.flags & SOCK_NONBLOCK) {
             rv = -EAGAIN;
             goto out;
         }
@@ -154,6 +154,15 @@ closure_function(6, 1, sysreturn, unixsock_read_bh,
         length -= xfer;
         if (!buffer_length(b) || (s->sock.type == SOCK_DGRAM)) {
             assert(dequeue(s->data) == shb);
+            if (s->sock.type == SOCK_DGRAM) {
+                struct sockaddr_un *from_addr = bound(from_addr);
+                socklen_t *from_length = bound(from_length);
+                if (from_addr && from_length) {
+                    if (from_length > 0)
+                        runtime_memcpy(from_addr, &shb->from_addr, MIN(*from_length, sizeof(shb->from_addr)));
+                    *from_length = __builtin_offsetof(struct sockaddr_un, sun_path) + runtime_strlen(from_addr->sun_path) + 1;
+                }
+            }
             sharedbuf_release(shb);
             shb = queue_peek(s->data);
             if (shb == INVALID_ADDRESS) { /* no more data available to read */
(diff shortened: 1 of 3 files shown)


src/unix/socket.c
```diff
@@ -482,19 +538,17 @@ static sysreturn unixsock_connect(struct sock *sock, struct sockaddr *addr,
...
-        return -ECONNREFUSED;
-    }
-    listener = pointer_from_u64(*((u64 *) buffer_ref(b, 0)));
-    assert(listener);
+    int rv = lookup_socket(&listener, unixaddr->sun_path);
+    if (rv != 0)
+        return rv;
     if (!s->connecting) {
+        if (s->sock.type & SOCK_DGRAM) {
+            if (!(listener->sock.type & SOCK_DGRAM))
```
francescolavra, line 547: Here and in the previous line, `&` should be replaced by `==`

src/unix/socket.c
```diff
@@ -571,21 +631,36 @@ static sysreturn unixsock_accept4(struct sock *sock, struct sockaddr *addr,
 sysreturn unixsock_sendto(struct sock *sock, void *buf, u64 len, int flags,
         struct sockaddr *dest_addr, socklen_t addrlen)
 {
-    /* Non-connected sockets are not supported, so destination address is
-     * ignored. */
-    return apply(sock->f.write, buf, len, 0, current, false,
-            syscall_io_complete);
+    unixsock s = (unixsock) sock;
+    if (dest_addr || addrlen) {
+        if ((sock->type & SOCK_STREAM)) {
```
francescolavra, line 636: `sock->type == SOCK_STREAM`; also (nit), unnecessary parentheses

src/unix/socket.c
```diff
@@ -571,21 +631,36 @@ static sysreturn unixsock_accept4(struct sock *sock, struct sockaddr *addr,
...
+        struct sockaddr *src_addr, socklen_t *addrlen)
 {
-    /* Non-connected sockets are not supported, so source address is not set. */
-    if (addrlen) {
-        *addrlen = 0;
+    if (src_addr || addrlen) {
+        if (!(src_addr && addrlen))
+            return -EFAULT;
+        if (!validate_user_memory(addrlen, sizeof(socklen_t), true) ||
+            !validate_user_memory(src_addr, *addrlen, true))
```
francescolavra, line 659: These validate_user_memory() calls are already in the generic recvfrom() function in net/netsyscall.c, so they shouldn't be necessary here.

test/runtime/unixsocket.c
```diff
@@ -173,28 +175,23 @@ static void *uds_dgram_server(void *arg)
...
     uint8_t readBuf[SMALLBUF_SIZE];
 
-    client_fd = accept(fd, (struct sockaddr *) &addr, &addr_len);
-    test_assert(client_fd >= 0);
+    test_assert(recvfrom(fd, readBuf, SMALLBUF_SIZE / 2, 0, (struct sockaddr *)&addr, &addr_len) ==
+            SMALLBUF_SIZE / 2);
     test_assert(addr_len > sizeof(addr.sun_family));
     test_assert(addr_len <= sizeof(addr));
     test_assert(addr.sun_family == AF_UNIX);
-    test_assert(!strcmp(addr.sun_path, CLIENT_SOCKET_PATH));
```
francescolavra, line 184: I think this should be kept, it's a valid check on both Nanos and Linux.

src/unix/socket.c
```diff
@@ -154,6 +154,15 @@ closure_function(6, 1, sysreturn, unixsock_read_bh,
         length -= xfer;
         if (!buffer_length(b) || (s->sock.type == SOCK_DGRAM)) {
             assert(dequeue(s->data) == shb);
+            if (s->sock.type == SOCK_DGRAM) {
+                struct sockaddr_un *from_addr = bound(from_addr);
+                socklen_t *from_length = bound(from_length);
+                if (from_addr && from_length) {
+                    if (from_length > 0)
```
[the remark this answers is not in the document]
sanderssj: Yes, this was intended to skip the memcpy when *from_length was zero. I removed the check, thanks!

src/unix/socket.c
```diff
@@ -230,30 +244,65 @@ static sysreturn unixsock_write_to(void *src, sg_list sg, u64 length,
...
                  u64, flags)
 {
     unixsock s = bound(s);
     void *src = bound(src);
     u64 length = bound(length);
+    unixsock dest;
 
     sysreturn rv;
 
     if ((flags & BLOCKQ_ACTION_NULLIFY) && s->peer) {
```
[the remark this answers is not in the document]
sanderssj: Good catch, thanks

src/unix/socket.c
```diff
@@ -230,30 +244,65 @@ static sysreturn unixsock_write_to(void *src, sg_list sg, u64 length,
...
+            goto out;
+        }
+    }
+    dest = s->peer;
+    if (s->sock.type == SOCK_DGRAM) {
+        struct sockaddr_un *addr = bound(addr);
+        if (addr && bound(addrlen)) {
+            if (bound(addrlen) < sizeof(struct sockaddr_un) ||
+                addr->sun_family != AF_UNIX)
+                return -EINVAL;
```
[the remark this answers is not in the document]
sanderssj: Yeah, I messed up this whole section by returning before the blockq_handle_completion was called.

src/unix/socket.c
sanderssj: Yup, I got these mixed up with the flags argument.

src/unix/socket.c
sanderssj: Fixed

test/runtime/unixsocket.c
sanderssj: Yes, this should definitely be there. I think I had removed it in the middle of changes, before recvfrom was implemented in this new way and did not restore it.

src/unix/socket.c
```diff
@@ -571,21 +633,33 @@ static sysreturn unixsock_accept4(struct sock *sock, struct sockaddr *addr,
...
+            if (s->peer)
+                return -EISCONN;
+            else
+                return -ENOTCONN;
+        }
+        if (!(dest_addr && addrlen))
+            return -EFAULT;
+        if (addrlen < sizeof(struct sockaddr_un))
+            return -EINVAL;
+        if (dest_addr && !validate_user_memory(dest_addr, addrlen, true))
```
wjhun, line 648: change true to false here if you end up removing the null terminator write

wjhun (approved):
One concern about writing to a const user buffer, but otherwise looks good to me.

src/unix/socket.c
```diff
@@ -230,30 +244,65 @@ static sysreturn unixsock_write_to(void *src, sg_list sg, u64 length,
...
+        }
+    }
+    dest = s->peer;
+    if (s->sock.type == SOCK_DGRAM) {
+        struct sockaddr_un *addr = bound(addr);
+        if (addr && bound(addrlen)) {
+            if (bound(addrlen) < sizeof(struct sockaddr_un) ||
+                addr->sun_family != AF_UNIX)
+                return -EINVAL;
+            addr->sun_path[sizeof(addr->sun_path)-1] = 0;
```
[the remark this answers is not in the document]
sanderssj: Good point! I just went ahead and made a copy as I don't think we need to optimize that for now.

src/unix/socket.c
sanderssj: I actually removed it completely because it is already correctly checked in the main sendto function.

PR #1506, opened 2021-06-17
Add support for xsave and enable AVX instructions

sanderssj:
This change enables AVX instructions in XCR0 if available, which also requires using
xsave for saving the newly enabled registers in context switches. If xsave is not
available then it will continue to use fxsave instead.

### DIFF
--- a/src/x86_64/crt0.s
+++ b/src/x86_64/crt0.s
@@ -47,8 +47,6 @@ extern  init_service
 
 extern use_xsave
 
-;; XXX - stick with fx until we can choose according to capabilities -
-;; otherwise existing stuff breaks with ops, etc.
 %macro load_extended_registers 1
         mov al, [use_xsave]
         test al, al
--- a/src/x86_64/crt0.s
+++ b/src/x86_64/crt0.s
@@ -45,19 +45,34 @@ extern  init_service
         wrmsr
 %endmacro
 
+extern use_xsave
 
 ;; XXX - stick with fx until we can choose according to capabilities -
 ;; otherwise existing stuff breaks with ops, etc.
 %macro load_extended_registers 1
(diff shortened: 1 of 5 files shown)


src/x86_64/kernel_machine.h
```diff
@@ -239,15 +249,16 @@ static inline cpuinfo current_cpu(void)
...
-#if 0
-    u32 v[4];
-    cpuid(0xd, 0, v);
-    return v[1];
-#else
-    return 512;                 /* XXX fx only right now */
-#endif
+    if (use_xsave) {
+        u32 v[4];
+        cpuid(0xd, 0, v);
```
wjhun, line 258: I guess this won't be called too frequently - just when a new frame is allocated or clone(2) is called - but it might save some vm exits if the xsave size is probed once on init and stashed away.

src/x86_64/crt0.s
```diff
@@ -45,19 +45,34 @@ extern  init_service
         wrmsr
 %endmacro
 
+extern use_xsave
 
 ;; XXX - stick with fx until we can choose according to capabilities -
```
wjhun, line 50: can nuke this comment

wjhun (approved):
LGTM!

src/x86_64/crt0.s
sanderssj: Removed

src/x86_64/kernel_machine.h
sanderssj: I realized that the kernel frames are allocated before the xsave or avx gets discovered, so I had to rework that part a bit to initialize the cr0/cr4/xcr0 settings on cpu 0 before entering the kernel runtime init so that those frames would be properly allocated, but now the size is only calculated once at init.

PR #1507, opened 2021-06-18
linear_backed_heap and page table work

wjhun:
This adds the linear_backed_heap to the kernel, which, along with the original backed_heap (now page_backed_heap) is a variant of a new backed_heap base type. It is meant to replace most uses of the old backed heap, with the advantage of making virtual-to-physical address conversion a simple mask instead of a page table lookup. Compared to the page_backed_heap, allocations and deallocations require only an operation on the physical heap (versus both physical and virtual), and no per-page mapping and unmapping is necessary. This is made possible by taking advantage of the copious address space available and mapping all* of physical memory at a known virtual base. By using the largest mapping size available (1GB on aarch64 and x86_64, if available), kernel memory usage should now cause a much smaller footprint on the TLB than before.

(* = On Intel, only the lowest 64TB of physical address space can be used for linear_backed allocations; it is assumed that platforms don't paint all their memory on the ceiling - and, just in case, the linear_backed_heap does a range allocate to be sure that only mapped memory is allocated.) Mappings are set up on initialization, so there is no concern that a linear_backed_heap allocate could lead to a map (and thus, the unified page table code - see below - can rely on this property, allowing heap allocates directly from page mapping code - while holding the page table lock - and getting rid of p2v lookup tables to access page table memory).

To facilitate such large mappings, support for 1GB pages needed to be added to the x86_64 page table code. This code was in a rather poor state, so instead the page table code form the aarch64 implementation has been made architecture-independent (at least, compatible 4-level page tables available on x86_64, aarch64, and even RISC-V should that ever get support) and the arch-specific code minimized to definitions and helpers to interpret and set up page table entries. Should 1GB pages not be available on a particular x86_64 implementation, as determined by cpuid on init, the page table code will simply resort to using 2MB and 4KB pages only. 2MB-based linear backed mappings costs only a single 4K page of table memory per 1GB of physical memory.

The page_backed_heap is still needed for uses where page protections of individual 4K pages must be modified (and update_map_flags will assert that the pages are not within the linear backed address space). Also, at least on x86_64, allocations from the linear_backed heap cannot contain tags, so tagged memory must still come from a page_backed_heap.

Some other changes which were necessary along the way:

- For the time being, all user mappings should be 4K - or the minimum system page size. This would naturally be the case for faulted-in pages, but large mappings could be used on exec when segments that exceed 2MB are loaded. This adds the "pageflags_minpage()" modifier as a way to signal to the page table code to only used minimum-sized pages. This ameliorates some crashes that arose on page permission changes previously, though it may be a topic to revisit as taking advantage of larger user mappings can have significant performance implications for the system, and the need to effect protections for a min-sized page area of memory could be achieved through re-mapping and factoring down mappings as necessary.

- I have gone through and converted uses of "heap_backed" to "heap_page_backed" or, preferably, "heap_linear_backed." If I wasn't sure about a particular use, I would default to heap_page_backed to be safe. These probably need some more scrutiny.


src/kernel/page.h
```diff
@@ -0,0 +1,44 @@
+/* Though page flags are just a u64, we hide it behind this type to
+   emphasize that page flags should be composed using helpers with
+   clear semantics, not architecture bits. This is to avoid mistakes
+   due to a union of PAGE_* constants on one architecture meaning
+   something entirely different on another. */
+
+typedef struct pageflags {
+    u64 w;                      /* _PAGE_* flags, keep private to page.[hc] */
```
francescolavra, line 8: Could remove the initial underscore from `_PAGE_*` in the comment, since these constants have been renamed without the underscore.

src/unix/syscall.c
```diff
@@ -1591,24 +1591,29 @@ static sysreturn brk(void *x)
...
+                u64 phys = physical_from_virtual(pointer_from_u64(new_end));
+                write_barrier();
+                unmap(new_end, len);
+                id_heap_set_area(heap_physical(kh), phys, len, true, false);
+            }
         } else if (p->brk < x) {
-            // I guess assuming we're aligned
-            u64 alloc = pad(u64_from_pointer(x), PAGESIZE) - pad(u64_from_pointer(p->brk), PAGESIZE);
+            u64 alloc = new_end - old_end;
+            assert(alloc > 0);
```
francescolavra, line 1611: Userspace should be allowed to call brk() with a non-page-aligned address, and the kernel should handle page alignment transparently for the program (meaning brk() should return the address requested by the program even if unaligned).

src/virtio/virtio_balloon.c
```diff
@@ -424,5 +424,5 @@ void init_virtio_balloon(kernel_heaps kh)
 {
     virtio_balloon_debug("%s\n", __func__);
     heap h = heap_locked(kh);
-    register_pci_driver(closure(h, vtpci_balloon_probe, h, kh->backed, heap_physical(kh)));
+    register_pci_driver(closure(h, vtpci_balloon_probe, h, (backed_heap)heap_huge_backed(kh), heap_physical(kh)));
```
francescolavra, line 427: heap_huge_backed() returns a backed_heap, no need to cast the return value

src/virtio/virtio_net.c
```diff
@@ -339,7 +339,7 @@ closure_function(2, 1, void, vtmmio_net_probe,
 void init_virtio_network(kernel_heaps kh)
 {
     heap h = heap_locked(kh);
-    backed_heap page_allocator = kh->backed;
+    backed_heap page_allocator = (backed_heap)heap_huge_backed(kh);
```
francescolavra, line 342: ditto

src/virtio/virtio_scsi.c
```diff
@@ -531,5 +531,5 @@ closure_function(3, 1, boolean, virtio_scsi_probe,
 void init_virtio_scsi(kernel_heaps kh, storage_attach a)
 {
     heap h = heap_locked(kh);
-    register_pci_driver(closure(h, virtio_scsi_probe, h, a, kh->backed));
+    register_pci_driver(closure(h, virtio_scsi_probe, h, a, (backed_heap)heap_huge_backed(kh)));
```
francescolavra, line 534: ditto

src/virtio/virtio_storage.c
```diff
@@ -257,7 +257,7 @@ void init_virtio_blk(kernel_heaps kh, storage_attach a)
 {
     virtio_blk_debug("%s\n", __func__);
     heap h = heap_locked(kh);
-    backed_heap page_allocator = kh->backed;
+    backed_heap page_allocator = (backed_heap)heap_huge_backed(kh);
```
francescolavra, line 260: ditto

src/x86_64/page_machine.h
```diff
@@ -0,0 +1,297 @@
...
+}
+
+static inline u64 pte_map_size(int level, pte entry)
+{
+    if (pte_is_present(entry)) {
+        // XXX take from arm
+        if (level == 4)
+            return PAGESIZE;
+        if ((entry & PAGE_PS) && level != 1)
+            return level == 2 ? HUGE_PAGESIZE : PAGESIZE_2M;
```
francescolavra, line 208: HUGE_PAGESIZE is defined as 0x100000000ull (4 GB), this function should return 1 GB instead.

francescolavra (commented):
With these changes, the kernel is unable to boot on an x86 VM with 4 GB physical memory: if I change QEMU_MEMORY in platform/pc/Makefile from 2G to 4G, the kernel fails on assert(pagemem.pageheap), at kernel/page.c:78; that's because the initial pagemem.current_phys range set up from REGION_INITIAL_PAGES is not enough to do the mappings requested before init_page_tables() is called.
Nanos master boots fine with 4 GB.

wjhun:
> With these changes, the kernel is unable to boot on an x86 VM with 4 GB physical memory: if I change QEMU_MEMORY in platform/pc/Makefile from 2G to 4G, the kernel fails on assert(pagemem.pageheap), at kernel/page.c:78; that's because the initial pagemem.current_phys range set up from REGION_INITIAL_PAGES is not enough to do the mappings requested before init_page_tables() is called.

I tested up to the maximum allowed on my system (64GB), but I did not think to test high memory configurations for systems that don't have 1GB pages. I suspect that is what is causing your failure. I can verify that startup fails for 4GB memory I manually disable 1GB pages.

Would you please see if the cpuid check in init_mmu() is detecting 1GB support on your setup? Out of curiosity, what hardware / qemu version are you testing on?

I think the solution here would be to simply raise INITIAL_PAGES_SIZE. Setting it to 1MB would allow the mapping of up to 256GB of physical memory for systems that don't support 1GB mappings.

francescolavra:
Yes, I verified that 1GB page support is not there according to what init_mmu() finds. v[3] from cpuid is 0x2193fbfd, so bit 26 is not set.
My cpuinfo is below:
```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 58
model name	: Intel(R) Pentium(R) CPU 2117U @ 1.80GHz
stepping	: 9
microcode	: 0x21
cpu MHz		: 896.976
cache size	: 2048 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 2
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic popcnt tsc_deadline_timer xsave lahf_lm epb ssbd ibrs ibpb stibp kaiser tpr_shadow vnmi flexpriority ept vpid fsgsbase smep erms xsaveopt dtherm arat pln pts md_clear flush_l1d
bugs		: cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit
bogomips	: 3592.13
clflush size	: 64
cache_alignment	: 64
address sizes	: 36 bits physical, 48 bits virtual
power management:
```
QEMU emulator version 5.0.50 (v5.0.0-1244-g7d3660e-dirty)

francescolavra:
> Getting back to the SMP issue, it appears that two different cores are faulting on a common (anonymous mmap) page, but through different threads. Again, this is a situation that would just fail silently before, as the old page table code silently overwrote entries (thus leaking some physical memory, too). I am working on a solution for this at the moment. [Edit: I confirmed that this is an issue that is occurring in master but is being silently ignored. A fix for that issue should precede this PR being merged.]

This issue seems to affect non-SMP systems too, because while a thread is blocked waiting for a page to be faulted in, a second thread could hit a page fault for the same page, thus triggering a second mapping request. In fact, I can reproduce the issue without SMP:
```
map_level error: would overwrite entry: level 4, v [0x1c1f4bb000 0x1c1f4bc000), pa 0x4cde000, flags 0x204, index 187, entry 0x4cde005
ra 0xffffffff8001bbbd
ffff80000009fea0:   ffffffff8001aa22	(map + 0000000000000332/0000000000000462)
ffff80000009ff30:   ffffffff8001bbbd	(map_page_finish + 000000000000007d/00000000000000a5)
ffff80000009ff50:   ffffffff8003a6d8	(merge_join + 0000000000000038/0000000000000088)
ffff80000009ff70:   ffffffff8001bfd2	(pagecache_service_completions + 00000000000000f2/00000000000001ea)
ffff80000009ffb0:   ffffffff80023898	(runloop_internal + 0000000000000178/000000000000055b)
map failed for v 0x1c1f4bb000, p 0x4cde000, len 0x1000, flags 0x204
```

wjhun:
> This issue seems to affect non-SMP systems too, because while a thread is blocked waiting for a page to be faulted in, a second thread could hit a page fault for the same page, thus triggering a second mapping request. In fact, I can reproduce the issue without SMP:

Yes, I found this as well. The pending fix will address SMP and non-SMP cases.

src/kernel/page.c
```diff
@@ -0,0 +1,531 @@
...
+                if (!map_level(tp, level + 1, irangel(v.start, len), p, flags, fe))
+                    return false;
+            }
+            page_init_debug("      pte @ ");
+            page_init_debug_u64(u64_from_pointer(&table_ptr[i]));
+            page_init_debug(" = ");
+            page_init_debug_u64(pte);
+            page_init_debug("\n");
+            table_ptr[i] = pte;
+            if (invalidate)
```
sanderssj, line 431: We might consider calling an invalidate on this address here or set invalidate to true (which I think is overkill) to cover an unusual case in smp. There's a scenario in which an address gets unmapped by one cpu and then mapped by another cpu but which hasn't run the shootdown yet so the old translation is still in the TLB. The page table is updated, but code run after the map would still have the old physical address causing problems. IIRC, this happened before with an interaction between the pagecache and storage drivers both of which were doing a lot of alloc/dealloc. I haven't tripped it yet though with this branch, so perhaps it's mitigated in a way I haven't found yet.

sanderssj (commented):
Looks good, only had a couple of comments!

src/unix/syscall.c
wjhun: Yes, I agree. I have further revised the brk(2) implementation and opened PR #1527 for it.

src/kernel/page.c
wjhun: I'm just circling back to this after the fault digression. I think this code would invalidate in the case you describe, though maybe I'm not understanding the scenario. The invalidate flag is just to only invalidate for actual page (or block / large page) addresses and not for intermediate tables; it's only set to false when installing a new table page.

francescolavra:
Works for me

src/kernel/page.c
sanderssj: Yes, it looks like you're right. I don't see a way that happens now, I think I was mistaken before.  Thanks!

PR #1508, opened 2021-06-18
Fix miscellaneous event behavior to match Linux

sanderssj:
This fixes some small event notification related bugs: EPOLLEXCLUSIVE was defined incorrectly. eventfd
edge notifications were fixed so that reads or writes are an edge for both reads and writes. Finally, tcp sockets
weren't reporting EPOLLHUP before listen/connect.

### DIFF
--- a/src/unix/eventfd.c
+++ b/src/unix/eventfd.c
@@ -15,10 +15,8 @@ closure_function(0, 2, u64, efd_edge_handler,
                 u64, events, u64, lastevents)
 {
     /* A read or a write acts as an edge */
-    if (events & EPOLLIN)
-        lastevents &= ~EPOLLIN;
-    if (events & EPOLLOUT)
-        lastevents &= ~EPOLLOUT;
+    if (events & (EPOLLIN|EPOLLOUT))
+        lastevents &= ~(EPOLLIN|EPOLLOUT);
     return lastevents;
 }
 
--- a/src/unix/eventfd.c
+++ b/src/unix/eventfd.c
@@ -15,8 +15,10 @@ closure_function(0, 2, u64, efd_edge_handler,
                 u64, events, u64, lastevents)
 {
     /* A read or a write acts as an edge */
-    if (events & (EPOLLIN|EPOLLOUT))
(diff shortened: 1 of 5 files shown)


src/unix/eventfd.c
```diff
@@ -15,10 +15,8 @@ closure_function(0, 2, u64, efd_edge_handler,
                 u64, events, u64, lastevents)
 {
     /* A read or a write acts as an edge */
-    if (events & EPOLLIN)
-        lastevents &= ~EPOLLIN;
-    if (events & EPOLLOUT)
-        lastevents &= ~EPOLLOUT;
+    if (events & (EPOLLIN|EPOLLOUT))
+        lastevents &= ~(EPOLLIN|EPOLLOUT);
```
wjhun, line 19: Just curious - is this behavior documented anywhere or is it just a peculiarity of Linux epoll with eventfd found through inspection?

wjhun (approved):
LGTM

src/unix/eventfd.c
francescolavra: It seems to me that the issue (or at least one of the issues) with the current eventfd code is that efd_read_bh() calls `notify_dispatch(efd->f.ns, EPOLLOUT)`, and analogously efd_write_bh() calls `notify_dispatch(efd->f.ns, EPOLLIN)`, but both should be calling `fdesc_notify_events(&efd->f)` instead, because otherwise an eventfd that has just been written but is still writable reports its readability but not its writeability (and the same goes for an eventfd that has just been read but is still readable).
@sanderssj does your test case behave correctly if you make these changes? If so, maybe the changes to efd_edge_handler are not needed.

sanderssj: The fix @francescolavra suggested was right, the problem is that the poll should be returning the current valid events, not only the one that just happened. I noticed there were a few other places where this was the case, and I have updated those as well. This also lead me to find a use-after-free problem with unix datagram sockets that requires refcounting, so I've added that as well.

francescolavra: LGTM

PR #1518, opened 2021-06-21
Network socket recv(): add support for MSG_PEEK flag

francescolavra:
This flag causes a receive operation to return data from the beginning of the receive queue without removing that data from the queue. Thus, a subsequent receive call will return the same data.

A new test case is being added to the netsock runtime test to exercise this feature.


wjhun (approved):
LGTM

PR #1522, opened 2021-06-27
virtIO SCSI: remove maximum queue size setting for GCP

### LINKED ISSUE #649: Remove GCE hacks
Currently two GCE hacks exist:
- virtqueue is limited to 1 request, if this limit is removed GCE kills nanos VM with hostError error in StackDriver log
- MAX_IO_SIZE (offset_block_io) is set to 256k (should be PAGESIZE)

I suppose hostError happens because of interrupt handling latency. Need to get back to this once we are able to run with interrupts enabled.

Some work for that is in feature/io-enable-interrupts branch but I was not able to get rid of hostError.

francescolavra:
The maximum number of pending virtIO SCSI requests was set to 1 for Google Cloud Platform because otherwise the SCSI device would return an "Overlapped commands attempted" error. This error was caused by the fact that the virtIO SCSI driver was failing to set the command identifier (corresponding to the "task tag" in the SCSI standard) in its requests.
The SCSI standard (SAM-3, section 5.9.3) states that "an overlapped command occurs when a task manager detects the use of a duplicate tag in a command before a task holding that tag completes its task lifetime. A task manager that detects an overlapped command shall abort all tasks for the faulted initiator port in the task set and the device server shall return CHECK CONDITION status for that command. The sense key shall be set to ABORTED COMMAND and the additional sense code shall be set to OVERLAPPED COMMANDS ATTEMPTED".
This commit adds setting of the command identifier to a unique value in virtio_scsi_alloc_request (similarly to what Linux does,
the identifier is set to the address of the SCSI request structure), and removes the maximum queue size setting, thereby
allowing multiple requests to be enqueued at the same time. Since the maximum queue size setting is now unused, this functionality is being removed from the virtqueue code.
With this change, the write runtime test shows a tenfold increase in performance for the filesystem stress test and a speed increase by a factor of 5.5 for the bulk write test.
Example test run before this change:
```
fs stress test
   write   1.659341256s
   delete   0.020281360s
   total   1.679622616s
bulk write test, size 20480 KB:
   write   0.039089554s
   fsync   8.994906062s
    read   0.022408186s
   total   9.056403802s (2261 KB/s)
```
Example test run after this change:
```
fs stress test
   write   0.139564050s
   delete   0.014996136s
   total   0.154560186s
bulk write test, size 20480 KB:
   write   0.042244509s
   fsync   1.608039334s
    read   0.021905470s
   total   1.672189313s (12247 KB/s)
```
Closes #649.

### DIFF
--- a/src/virtio/virtio_internal.h
+++ b/src/virtio/virtio_internal.h
@@ -109,8 +109,6 @@ status virtqueue_alloc(vtdev dev,
                        thunk *t,
                        queue sched_queue);
 
-void virtqueue_set_max_queued(virtqueue, int);
-
 /* The Host uses this in used->flags to advise the Guest: don't kick me
  * when you add a buffer.  It's unreliable, so it's simply an
  * optimization.  Guest will still kick if it's out of buffers. */
--- a/src/virtio/virtio_scsi.c
+++ b/src/virtio/virtio_scsi.c
@@ -177,6 +177,7 @@ static virtio_scsi_request virtio_scsi_alloc_request(virtio_scsi s, u16 target, 
     r->req.lun[1] = target;
     r->req.lun[2] = ((lun >> 8) & 0x3f) | 0x40;
     r->req.lun[3] = (lun & 0xff);
+    r->req.id = u64_from_pointer(r);
     r->alloc_len = alloc_len;
 
     return r;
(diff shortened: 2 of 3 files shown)


sanderssj (commented):
LGTM!

wjhun (approved):

PR #1523, opened 2021-06-29
Build with SMP_ENABLE defined

sanderssj:
This change defines SMP_ENABLE in the Makefile so nanos will be built with spinlocks and tlb shootdowns but
it still only uses one processor for tests. It also fixes a couple of race issues in runtime tests and expands the 
service queue in virtqueue because of runtime tests failures when running with multiple cpus.


wjhun (approved):
Looks good!

PR #1524, opened 2021-06-29
Release script: add optional smoke tests on cloud providers

### LINKED ISSUE #1015: release script should have an optional smoke test through gcloud && aws
gcloud is super quick, aws takes a bit longer - but worth rolling those into the release script as we pick up more velocity

francescolavra:
The release.sh script now takes an optional command line argument, which if present specifies a command to be executed before publishing the release artifacts.
Currently, the only supported command is `smoke-test`, which triggers execution of smoke tests on different cloud providers (currently, gcp and aws). A smoke test consists of creating a cloud image with the webg program, starting an instance from this image, sending 1000 HTTP requests to this instance, and checking that it responds correctly; if any of these steps fails, the release artifacts are not published. The created instance and image are then deleted as a final cleanup step.
The location of config.json files for Ops can be specified via environment variables:
- OPS_CONF_GCP specifies the location of the config file for gcp
- OPS_CONF_AWS specifies the location of the config file for aws
- OPS_CONF specifies the location of a generic config file, which is used for a given cloud provider if no provider-specific environment variable is set.

Example:
```
export OPS_CONF_GCP=~/config-gcp.json
export OPS_CONF_AWS=~/config-aws.json
./release.sh smoke-test
```
These changes are compatible with Linux and macOS shells.
Closes #1015.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,4 +1,94 @@
-#!/bin/sh
+#!/bin/bash
+
+smoke_test()
+{
+  cfg="$2"
+  cfg_binaries='"Boot": "output/platform/pc/boot/boot.img", "Kernel": "output/platform/pc/bin/kernel.img"'
+  image_name='webg-smoke-test'
+  instance_name="$image_name-`date +'%s'`"
+  tmp_file=output/test/smoke.json
+  server_port=8080
+  if [ -z $cfg ]; then # no provider-specific cloud configuration has ben supplied
+    cfg=${OPS_CONF}    # use generic cloud configuration, if supplied
+  fi
+  if [ ! -z "$cfg" ]; then
+    if [ ! -f "$cfg" ]; then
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -1,5 +1,95 @@
 #!/bin/sh
```
sanderssj, line 1: It may not matter so much for how we use it, but since this has some bashisms, we might consider changing this to something like #!/bin/bash or #!/usr/bin/env bash for maximum portability.

sanderssj (commented):
Looks fine to me, just had one comment about the interpreter.

wjhun (approved):

release.sh
francescolavra: OK, I changed it to #!/bin/bash

PR #1527, opened 2021-07-01
brk(2): unmap and deallocate pages when heap size is reduced, do not force alignment on brk value reported to userspace

wjhun:
The previous implementation of brk(2) was not unmapping and deallocating pages on reduction of heap size. The effects of this omission were masked due to the x86_64 page table code allowing entries to be overwritten. This would result in leaked pages if the process heap was lowered and then later raised again. This change unmaps and deallocates pages that become unassociated with the process heap after a call to brk(2).

This also remedies an existing bug which was forcing the reported process break to be page-aligned. The process should be able to set the break to an unaligned address, and the returned break value should equate to the value set.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -1585,35 +1585,41 @@ static sysreturn getcwd(char *buf, u64 length)
     return cwd_len;
 }
 
-static sysreturn brk(void *x)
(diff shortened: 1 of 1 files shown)


francescolavra (commented):
Looks good to me

PR #1529, opened 2021-07-06
x86 IPIs: set ASSERT bit in ICR flags

francescolavra:
The Intel software developer's manual (volume 3A, chapter 10 "Advanced Programmable Interrupt Controller", section 10.6.1 "Interrupt Command Register") states that the "Level" field of the ICR must be set to 1 for all IPI delivery modes (except INIT level de-assert).
This commit sets the ICR_ASSERT flag in calls to apic_ipi(), according to the above requirement. This fixes an issue observed on Hyper-V multi-processor instances where the kernel fails to start secondary cores.


wjhun (approved):
LGTM

PR #1542, opened 2021-07-18
Dump tool output changes

francescolavra:
The first commit fixes the change to the terminal foreground color made by dump when displaying filesystem contents in tree format (with the `-t` command line option).
The second commit changes the dump output to display filesystem metadata only when the user does not select the tree format option.


wjhun (approved):
LGTM

PR #1544, opened 2021-07-18
Implement O_TMPFLAG flag in the open syscall

sanderssj:
This change implements O_TMPFLAG by adding a way to create an unnamed file with no parent that
will not be saved in the filesystem. It was also necessary to add a refcount to fsfile so that the extents
are deallocated when the last file descriptor is closed. This new refcount also fixes a problem where an
opened file is unlinked causing existing extents to be deallocated immediately, even though they could
be referenced through the open fd. I think this problem was masked by the pagecache as I could not
easily create a test case to trip it. The unlink runtime test is updated to test O_TMPFILE as that is
the closest equivalent behavior (open file and then unlink it).


test/runtime/unlink.c
```diff
@@ -45,6 +48,53 @@ static void test_unlinkat(int dirfd)
...
+        perror("tmpfile rdonly open");
+        exit(EXIT_FAILURE);
+    }
+    fd = open("/tmp", O_RDWR|O_TMPFILE);
+    if (fd < 0) {
+        perror("tmpfile open");
+        exit(EXIT_FAILURE);
+    }
+    for (int i = 0; i < sizeof(buf); i++)
+        buf[i] = i % sizeof(buf[0]);
```
francescolavra, line 75: sizeof(buf[0]) is 1, so this is basically setting buf[i] to a constant value (0); did you mean something else?

test/runtime/unlink.c
```diff
@@ -45,6 +48,53 @@ static void test_unlinkat(int dirfd)
...
+    char buf[8192];
+    char vbuf[8192];
+
+    if (mkdir("/tmp", 0777) != 0 && errno != EEXIST) {
+        perror("mkdir /tmp");
+        exit(EXIT_FAILURE);
+    }
+    statfs("/", &statbuf);
+    unsigned long fsfree = statbuf.f_bfree;
+    fd = open("/tmp", O_RDONLY|O_TMPFILE);
```
francescolavra, line 64: With gcc 8, this fails to build: `/usr/include/x86_64-linux-gnu/bits/fcntl2.h:50:4: error: call to ‘__open_missing_mode’ declared with attribute error: open with O_CREAT or O_TMPFILE in second argument needs 3 arguments`.

test/runtime/unlink.c
```diff
@@ -45,6 +48,53 @@ static void test_unlinkat(int dirfd)
...
+    if (fd != -1) {
+        perror("tmpfile rdonly open");
+        exit(EXIT_FAILURE);
+    }
+    fd = open("/tmp", O_RDWR|O_TMPFILE);
```
francescolavra, line 69: Same gcc error here.

src/tfs/tfs.c
```diff
@@ -1193,6 +1216,14 @@ void filesystem_log_rebuild_done(filesystem fs, log new_tl)
     fs->temp_log = 0;
 }
 
+void file_extents_dealloc(filesystem fs, tuple t);
```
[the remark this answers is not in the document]
sanderssj: Yeah, that was leftover. I've removed it, thanks.

test/runtime/unlink.c
sanderssj: No, it just needs to be a straight assignment. I've fixed it.

test/runtime/unlink.c
sanderssj: Thanks, I hadn't seen this error yet.

src/unix/syscall.c
```diff
@@ -637,6 +637,9 @@ closure_function(2, 2, sysreturn, file_close,
                  thread, t, io_completion, completion)
 {
     file f = bound(f);
+    fsfile fsf = bound(fsf);
+    if (fsf && f->f.type == FDESC_TYPE_REGULAR)
```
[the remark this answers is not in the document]
sanderssj: Thanks, I wasn't totally sure so I was leaning towards being safe. I took out the pointless check.

wjhun (approved):
LGTM

PR #1545, opened 2021-07-19
Add gdb printing helpers for nanos structs

sanderssj:
This adds some python-based printing helpers for the buffer and table structs to aid
in examining memory.

### DIFF
--- a/tools/nanos_gdb.py
+++ b/tools/nanos_gdb.py
@@ -1,0 +1,69 @@
+import gdb
+
+# these definitions and tagof must be in sync with C source
+class Tag:
+    unknown = 0
+    string = 1
+    symbol = 2
+    table_tuple = 3
+    function_tuple = 4
+
+def tagof(p):
+    return ((p.cast(typ('u64')) >> 38) & 7)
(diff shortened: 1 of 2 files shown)


tools/nanos_gdb.py
```diff
@@ -0,0 +1,69 @@
...
+# these definitions and tagof must be in sync with C source
+class Tag:
+    unknown = 0
+    string = 1
+    symbol = 2
+    table_tuple = 3
+    function_tuple = 4
+
+def tagof(p):
+    return ((p.cast(typ('u64')) >> 38) & 7)
```
francescolavra, line 12: VA_TAG_OFFSET in src/x86_64/machine.h is defined as 39, shouldn't there be 39 here as well?

wjhun: VA_TAG_OFFSET changed from 39 to 38 with the merging of the linear_backed_heap code. This is due to bit 46 being used to indicate a linear-backed mapping.

wjhun (approved):
Nice!

PR #1547, opened 2021-07-23
fix race condition when setting up new user pages

wjhun:
The map_and_zero function, used by both demand_anonymous_page and brk to set up new user mappings with zeroed contents, allowed for a race condition whereby a mapping could be set up on one processor (typically as a result of a page fault, not an mmap itself) and then be used by a thread running on another processor before the function had completed zeroing the memory. With webg running in an SMP configuration and a large number of concurrent connections, a range of errors could occur as a result of some memory being zeroed after having been written to. In one such failure mode within the go runtime, a queueing function that used cas would loop indefinitely as the target value, having been zeroed, would never reach the expected value.

This solution takes advantage of the new linear_backed_heap by making allocations for physical memory through it, allowing the memory to be zeroed prior to setting up the desired user mapping. It also obviates the need to first map read-only memory as writable and then update the flags to read-only, thus simplifying the operation.

### DIFF
--- a/src/aarch64/page_machine.h
+++ b/src/aarch64/page_machine.h
@@ -404,18 +404,4 @@ static inline pageflags pageflags_from_pteptr(pteptr pp)
     return (pageflags){.w = PAGE_FLAGS_MASK & *pp};
 }
 
-static inline void map_and_zero(u64 v, physical p, u64 length, pageflags flags, status_handler complete)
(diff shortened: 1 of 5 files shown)


sanderssj:
LGTM! It's handy the new linear backed heap offered and easy alternative.

PR #1551, opened 2021-07-31
virtio: vqmsg_push(): fix length value passed to buffer_extend()

francescolavra:
buffer_extend() takes as length value the number of bytes that the caller wants to append to the buffer, not the number of total bytes in the buffer after appending.
This fix prevents an unnecessary buffer extension in all cases where VQMSG_DEFAULT_SIZE (currently 3) descriptors are associated to a virtqueue message, which happens at each I/O request made by virtio-block and virtio-scsi drivers.


sanderssj:
LGTM

wjhun (approved):
Looks good

PR #1552, opened 2021-08-01
brk(): fix deallocation of physical memory

francescolavra:
The process heap may consist of non-contiguous chunks of physical memory, thus when the heap is being shrunk the page table needs to be walked to retrieve the physical pages to be deallocated.


sanderssj:
LGTM

wjhun (approved):
Good catch...this bug could have had some nasty effects.

PR #1553, opened 2021-08-01
buffer_extend(): fix handling of consumed buffers

francescolavra:
When less than `len` bytes are available after the end of the current contents, buffer_extend() calculates empirically a new capacity value based on the length of the current contents and the `len` value.
If the buffer has been at least partially consumed from, the calculated capacity value may be less than or equal to the current capacity. If it is equal, then buffer_set_capacity() does not do anything, which leaves the buffer in a state not suitable for appending `len` bytes; if it is less, then the buffer memory area is reallocated with a smaller size, which is not necessary.
This PR fixes the above issues by modifying buffer_extend() so that buffer_set_capacity() is called only if the current capacity value has to be increased, otherwise the current buffer contents are moved to the beginning of the allocated memory to make room for appending new data.
The buffer unit tests have been enhanced with a new test case that would have failed without this fix.

### DIFF
--- a/src/runtime/buffer.h
+++ b/src/runtime/buffer.h
@@ -107,7 +107,14 @@ static inline boolean buffer_extend(buffer b, bytes len)
     // xxx - pad to pagesize
     if (b->length < (b->end + len)) {
         bytes new_len = 2 * (b->end - b->start + len);
-        return (buffer_set_capacity(b, new_len) == new_len);
+        if (new_len > b->length) {
+            return (buffer_set_capacity(b, new_len) == new_len);
+        } else {
+            /* no need to resize, move current contents to the beginning of the allocated memory */
+            runtime_memcpy(b->contents, b->contents + b->start, b->end - b->start);
+            b->end -= b->start;
+            b->start = 0;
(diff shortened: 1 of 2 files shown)


src/runtime/buffer.h
```diff
@@ -107,7 +107,14 @@ static inline boolean buffer_extend(buffer b, bytes len)
     // xxx - pad to pagesize
     if (b->length < (b->end + len)) {
         bytes new_len = 2 * (b->end - b->start + len);
-        return (buffer_set_capacity(b, new_len) == new_len);
+        if (new_len > b->length) {
+            return (buffer_set_capacity(b, new_len) == new_len);
+        } else {
+            /* no need to resize, move current contents to the beginning of the allocated memory */
+            runtime_memcpy(b->contents, b->contents + b->start, b->end - b->start);
```
wjhun, line 114: I think this is fine but just thinking this through: the source and destination ranges may overlap, but it should be no issue as data is being moved from a higher to lower location, right?

wjhun (approved):
LGTM

PR #1554, opened 2021-08-02
Add support for SO_REUSEADDR and TCP_NODELAY

sanderssj:
This implements the setsockopt to set the option or flags for the SO_REUSEADDR
and TCP_NODELAY options on sockets by setting their appropriate flags in
the lwip pcbs and enabling reuse address support in lwipopts.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1888,6 +1888,44 @@ sysreturn setsockopt(int sockfd,
             goto unimplemented;
         }
         break;
+    case SOL_SOCKET:
+        switch (optname) {
+        case SO_REUSEADDR:
+            if (optlen != sizeof(int))
+                return -EINVAL;
+            if (s->sock.type == SOCK_STREAM) {
+                if (*((int *)optval))
+                    ip_set_option(s->info.tcp.lw, SOF_REUSEADDR);
+                else
+                    ip_reset_option(s->info.tcp.lw, SOF_REUSEADDR);
+            } else if (s->sock.type == SOCK_DGRAM){
+                if (*((int *)optval))
+                    ip_set_option(s->info.udp.lw, SOF_REUSEADDR);
+                else
+                    ip_reset_option(s->info.udp.lw, SOF_REUSEADDR);
+            } else
+                return -EINVAL;
+            break;
+        default:
+            goto unimplemented;
+        }
+        break;
+    case SOL_TCP:
+        switch (optname) {
+        case TCP_NODELAY:
+            if (optlen != sizeof(int))
+                return -EINVAL;
+            if (s->sock.type != SOCK_STREAM)
+                return -EINVAL;
+            if (*((int *)optval))
+                tcp_nagle_enable(s->info.tcp.lw);
+            else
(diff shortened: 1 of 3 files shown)


src/net/netsyscall.c
```diff
@@ -1888,6 +1888,43 @@ sysreturn setsockopt(int sockfd,
...
+        switch (optname) {
+        case SO_REUSEADDR:
+            if (optlen != sizeof(int))
+                return -EINVAL;
+            if (s->sock.type == SOCK_STREAM) {
+                if (*((int *)optval))
+                    s->info.tcp.lw->so_options |= SOF_REUSEADDR;
+                else
+                    s->info.tcp.lw->so_options &= ~SOF_REUSEADDR;
+            } else {
```
wjhun, line 1901: I know that stream and dgram are the only options at the moment, but we're likely to at least add SOCK_RAW...maybe just explicitly check for SOCK_DGRAM here (or use switch - case) to mask any unexpected bad behavior in the future? I realize this is probably an issue in other places ... in general it may be safer to not assume any default going forward.

wjhun (approved):
Looks good

src/net/netsyscall.c
```diff
@@ -1888,6 +1888,43 @@ sysreturn setsockopt(int sockfd,
...
+    case SOL_SOCKET:
+        switch (optname) {
+        case SO_REUSEADDR:
+            if (optlen != sizeof(int))
+                return -EINVAL;
+            if (s->sock.type == SOCK_STREAM) {
+                if (*((int *)optval))
+                    s->info.tcp.lw->so_options |= SOF_REUSEADDR;
+                else
+                    s->info.tcp.lw->so_options &= ~SOF_REUSEADDR;
```
francescolavra, line 1900: Alternatively to setting and clearing directly the option bits, ip_set_option() and ip_reset_option() could be called (same for UDP below).

src/net/netsyscall.c
```diff
@@ -1888,6 +1888,43 @@ sysreturn setsockopt(int sockfd,
...
+        switch (optname) {
+        case TCP_NODELAY:
+            if (optlen != sizeof(int))
+                return -EINVAL;
+            if (s->sock.type != SOCK_STREAM)
+                return -EINVAL;
+            if (*((int *)optval))
+                s->info.tcp.lw->flags |= TF_NODELAY;
+            else
+                s->info.tcp.lw->flags &= ~TF_NODELAY;
```
francescolavra, line 1922: Here, tcp_set_flags() and tcp_clear_flags() could be used, or alternatively tcp_nagle_disable() and tcp_nagle_enable().

src/net/netsyscall.c
sanderssj: Good point, I've updated accordingly.

src/net/netsyscall.c
sanderssj: I switched over to use the lwip convenience functions!

src/net/netsyscall.c
sanderssj: Same as above

PR #1555, opened 2021-08-03
TFS: fix handling of files not linked to the filesystem

francescolavra:
When a file is not linked to the filesystem (e.g. as a result of an unlink or rename syscall), its metadata and extent tuples can potentially be destroyed at any time when the TFS log is compacted. Thus, they should not be accessed, even if the file may still exist due to open file descriptors.
This PR modifies the code that handles tuples associated to a file so that it accesses these tuples only if the `md` member of struct fsfile is non-zero (which indicates that the file is linked to the filesystem). filesystem_alloc() and filesystem_dealloc() now take an fsfile argument instead of a tuple, because they can be called for files without an associated metadata tuple.
filesystem_creat_unnamed() no longer creates a tuple (because files opened with O_TMPFILE are not linked to the filesystem), and returns an fsfile instead of a tuple.
filesystem_rename() is now fixed so that if a file is being deleted (because it's being overwritten by another file), its reference count is decreased instead of deallocating its extents, because the file could still be referenced by open file descriptors.
When a file is finally deallocated (i.e. when its reference count becomes zero), deallocate_fsfile() is now called instead of file_extents_dealloc() (which was leaking the extent rangemap).
deallocate_fsfile() has been modified to be able to use a generic extent destructor closure.
The fallocate and write runtime tests have been amended to exercise parts of the code that handles files not linked to the filesystem.

The first commit is not specifically related to files not linked to the filesystem, but is a fix for a bug that surfaced when writing test cases to exercise code that handles these files.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -491,35 +491,36 @@ static void destroy_extent(filesystem fs, extent ex)
 
 static fs_status add_extent_to_file(fsfile f, extent ex)
 {
-    heap h = f->fs->h;
-    tuple extents;
-    symbol a = sym(extents);
-    assert(f->md);
-    if (!(extents = get_tuple(f->md, a))) {
-        extents = allocate_tuple();
-        fs_status s = filesystem_write_eav(f->fs, f->md, a, extents);
+    if (f->md) {
+        heap h = f->fs->h;
+        tuple extents;
+        symbol a = sym(extents);
+        if (!(extents = get_tuple(f->md, a))) {
+            extents = allocate_tuple();
+            fs_status s = filesystem_write_eav(f->fs, f->md, a, extents);
+            if (s != FS_STATUS_OK) {
+                deallocate_value(extents);
+                return s;
+            }
+            set(f->md, a, extents);
+        }
+
+        // XXX encode this as an immediate bitstring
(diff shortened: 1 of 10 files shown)


sanderssj:
LGTM!

src/tfs/tfs.c
```diff
@@ -622,20 +627,22 @@ static fs_status fill_gap(fsfile f, sg_list sg, range blocks, merge m, u64 *edge
...
 static fs_status update_extent_length(fsfile f, extent ex, u64 new_length)
 {
-    value v = value_from_u64(f->fs->h, new_length);
-    fs_status s = filesystem_write_eav(f->fs, ex->md, sym(length), v);
-    if (s != FS_STATUS_OK)
-        return s;
+    if (f->md) {
+        assert(ex->md);
+        value v = value_from_u64(f->fs->h, new_length);
+        fs_status s = filesystem_write_eav(f->fs, ex->md, sym(length), v);
```
wjhun, line 633: Could stash sym(length) in a local var

wjhun (approved):
LGTM

src/tfs/tfs.c
francescolavra: Right, using a local var now

PR #1556, opened 2021-08-06
interrupt handlers: move thread enqueue to follow interrupt processing

wjhun:
The external interrupt dispatch routines were re-enqueueing the interrupted
user thread before setting the FRAME_FULL flag and calling interrupt
handlers. This could cause a "frame already full" fault to occur if such a
thread were to be migrated to another CPU and executed before the FRAME_FULL
set occurred in interrupt dispatch. The migrated thread would run and take an
interrupt or issue a syscall, at which point the FRAME_FULL flag would have
been set by the original interrupt dispatch, causing a terminal fault.

This fix simply moves the call to schedule_frame to follow setting of the
FRAME_FULL flag. Even if the thread is migrated to another CPU prior to
runloop service on the originating CPU, the set flag will be cleared upon
return to the user frame.

### DIFF
--- a/src/aarch64/interrupt.c
+++ b/src/aarch64/interrupt.c
@@ -267,13 +267,8 @@ void irq_handler(void)
 
     int_debug("%s: enter\n", __func__);
 
-    /* re-enqueue interrupted user thread */
-    if (ci->state == cpu_user && !shutting_down) {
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
Looks good to me

PR #1558, opened 2021-08-09
Set current thread to access user's epoll events array

sanderssj:
If the user events array passed in epoll_wait hasn't been paged in and the
epoll_wait_notify is triggered without a thread then the page fault
for that user memory will fail without the thread context.


francescolavra (commented):
LGTM

wjhun (approved):
Looks good

PR #1560, opened 2021-08-17
epoll: fix edge trigger behavior

francescolavra:
When epoll_wait() is called, the current event status is checked for all registered file descriptors by calling check_fdesc(), which notifies the current events to the epoll_wait_notify() handler. The events notified to the handler should reflect the current status of a file descriptor, regardless of the last events set on the epoll instance for that descriptor, because the handler uses the notified events to update (set and/or clear as appropriate) the last events. The existing code was removing the last events from the events reported to the handler, which causes the last events to toggle between successive invocations of epoll_wait() even if the status of the file descriptor does not change.
A new test case (test_edgetrigger() function) has been added to the epoll runtime tests to trigger the above issue: without this fix, the third call to epoll_wait() would report the EPOLLOUT event for both the first and the third file descriptor, even though the status of the first file descriptor does not change between the first and the third call to epoll_wait().
In order to avoid breaking the peculiar edge trigger behavior of eventfd, the eventfd edge handler has been modified so that both EPOLLIN and EPOLLOUT are reported after a read or write, to mirror the Linux implementation. This behavior is now tested in the epoll runtime tests.

### DIFF
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -365,10 +365,7 @@ static epoll_blocked alloc_epoll_blocked(epoll e)
 
 static void check_fdesc(epollfd efd, fdesc f, thread t)
 {
-    /* if edge-triggered, only notify the changes to events */
     u32 events = apply(f->events, t);
-    if (efd->eventmask & EPOLLET)
-        events &= ~efd->lastevents;
     notify_dispatch_for_thread(f->ns, events, t);
 }
 
--- a/src/unix/eventfd.c
+++ b/src/unix/eventfd.c
@@ -9,16 +9,20 @@ struct efd {
     int flags;
     blockq read_bq, write_bq;
     u64 counter;
+    boolean io_event;
 };
 
-closure_function(0, 2, u64, efd_edge_handler,
+closure_function(1, 2, u64, efd_edge_handler,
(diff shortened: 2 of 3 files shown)


src/unix/poll.c
```diff
@@ -365,10 +365,7 @@ static epoll_blocked alloc_epoll_blocked(epoll e)
 
 static void check_fdesc(epollfd efd, fdesc f, thread t)
 {
-    /* if edge-triggered, only notify the changes to events */
     u32 events = apply(f->events, t);
-    if (efd->eventmask & EPOLLET)
-        events &= ~efd->lastevents;
```
sanderssj, line 371: I think this currently breaks some epoll_wait behavior in other circumstances based on the hacky epoll test I have. I will write up some simplified test cases to help identify those areas. I think this particular change is right, but that we'll need some additional changes to not break existing behavior.

francescolavra: @sanderssj I modified the eventfd edge handler according to your suggestions, and now it mirrors the Linux behavior. I also added your eventfd-related tests to the epoll runtime tests. Thanks!

sanderssj:
LGTM

wjhun (approved):

PR #1561, opened 2021-08-18
Network sockets: fix syscall handling for shut down TCP sockets

francescolavra:
When both TX and RX directions of a TCP socket are shut down, the `info.tcp.lw` netsock struct member is set to NULL, because the lwIP PCB structure can be deallocated at any time and should not be referenced anymore. If other syscalls are invoked on a shut down socket, they should deal properly with this state and avoid accessing a NULL pointer.
A few more test cases have been added to the netsock runtime tests to exercise handling of sockets after shutdown.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1168,8 +1168,8 @@ static sysreturn netsock_bind(struct sock *sock, struct sockaddr *addr,
         IP_SET_TYPE(&ipaddr, IPADDR_TYPE_ANY);
     err_t err;
     if (sock->type == SOCK_STREAM) {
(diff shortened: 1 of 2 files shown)


sanderssj:
LGTM, just a have little nit. Although it would be a user error and very unlikely, I think you could currently crash the kernel with a bind after something frees the pcb.

francescolavra:
> I think you could currently crash the kernel with a bind after something frees the pcb

Good catch, I added a check in the netsock_bind() function so that it handles gracefully sockets with a NULL pcb.

wjhun (approved):

PR #1562, opened 2021-08-18
accept(): return socket file descriptor even if connection is reset

francescolavra:
The current code makes the accept() syscall return -ECONNRESET if the TCP socket just created has been disconnected already. This error code is not among the errors returned by other operating systems for this syscall (even Linux, which may return network errors on the new socket as error codes from accept(), never returns -ECONNRESET), and is likely to be considered as a fatal error by user programs (notably, it causes Golang web server functions such as http.ListenAndServe() and
http.ListenAndServeTLS() to exit with error).
This PR changes accept_bh() so that it returns the file descriptor of the newly created socket even if the socket is disconnected (in this case, the peer address cannot be retrieved and will be set to a dummy value). As a side effect, the newly created socket is no longer leaked, because it can be closed by the user program via the returned file descriptor, and the EPOLLIN flag
is correctly reset on the listening socket if there are no more incoming connections.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1759,22 +1759,24 @@ closure_function(5, 1, sysreturn, accept_bh,
         return BLOCKQ_BLOCK_REQUIRED;               /* block */
     }
 
-    err_t child_err = get_lwip_error(child);
-    if (child_err != ERR_OK) {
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

wjhun (approved):

PR #1567, opened 2021-08-28
Netlink read(): apply proper I/O completion

francescolavra:
The Netlink read() closure was applying the generic syscall I/O completion instead of the completion supplied as argument.


sanderssj:
LGTM

wjhun (approved):

PR #1568, opened 2021-08-28
futex(FUTEX_CMP_REQUEUE): fix return value

francescolavra:
The existing code is storing the number of requeued threads in a local variable that is out of scope when the syscall return value is calculated.
The futex runtime tests have been amended to check for the return value from futex(FUTEX_CMP_REQUEUE).


sanderssj (commented):
LGTM

wjhun (approved):

PR #1569, opened 2021-08-30
Revise blockq with locking and other fixes

wjhun:
To support syscall execution outside of the domain of the kernel lock, the
blockq code is revised to allow operations on a blockq to occur safely on
multiple processors. This essentially amounts to protecting the list of
waiting threads per blockq with a spinlock. Other improvements that have been
made include the embedding of the previous "blockq_item" struct directly into
the thread struct, which obviates the need to (de)allocate items on each
block, adding blockq_reserve() and blockq_release() calls to support safe
implementation of thread_attempt_interrupt(), and associating any
io_completions for a given blocking operation with each waiting thread instead
of with the blockq, thus allowing multiple waiting threads to have their own
io_completions, rather than only a single I/O completion for a given
resource (blockq).

The comments at the head of blockq.c were very outdated and have been updated
to reflect the current behavior and usage.

### DIFF
--- a/src/unix/blockq.c
+++ b/src/unix/blockq.c
@@ -2,64 +2,37 @@
 
 /* blockq - a facility for queueing up threads waiting for a resource
 
-   Given an action closure, blockq_check() will attempt the action
-   while holding the blockq's lock. If it succeeds (rv >= 0), it will
-   release the lock and return the result of the action to the
-   caller. If rv is negative, an error condition is presumed, the lock
-   is released and rv is returned at once. If rv ==
-   BLOCKQ_BLOCK_REQUIRED, the action gets added to the blockq's
-   waiters queue, an optional timeout is set, the lock is released and
+   Given an action closure, blockq_check() will attempt the action at once. If
+   it either succeeds (rv >= 0) or fails (rv < 0), it will return the result
+   of the action to the caller. If rv == BLOCKQ_BLOCK_REQUIRED, the action
+   gets added to the blockq's waiters queue, an optional timeout is set, and
    the thread is finally blocked.
 
-   blockq_wake_one(), meant to be safe for calling from interrupt context,
-   takes the blockq lock (really a no-op until SMP support) and
-   attempts to apply the action at the head of the waiters queue. If
-   it returns BLOCKQ_BLOCK_REQUIRED, it is left at the head of the
-   queue, the lock is released and waiting resumes, otherwise it
-   causes the action to be removed from the queue, and the lock
-   released. In either case, the call returns to the caller. It is up
-   to the action to apply results to the thread frame and call
-   thread_wakeup() as necessary.
+   blockq_wake_one() attempts to apply the action at the head of the waiters
+   queue. If it returns BLOCKQ_BLOCK_REQUIRED, the waiter is moved to the
+   bottom of the waiters queue and waiting resumes. The waiter is otherwise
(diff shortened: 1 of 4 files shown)


src/unix/signal.c
```diff
@@ -944,8 +944,12 @@ closure_function(1, 2, void, signalfd_notify,
         sig_debug("%d spurious notify\n", sfd->fd);
         return;
     }
-    blockq_wake_one_for_thread(sfd->bq, t);
-    notify_dispatch_for_thread(sfd->f.ns, EPOLLIN, t);
+
+    /* null thread on notify set release (thread dealloc) */
+    if (t) {
+        blockq_wake_one_for_thread(sfd->bq, t);
+        notify_dispatch_for_thread(sfd->f.ns, EPOLLIN, t);
```
francescolavra, line 951: To keep the current behavior, notify_dispatch_for_thread() should be called even if `t==0`, otherwise any threads doing polling on a signalfd won't be notified when the signalfd is closed (all notify handlers are safe to be called with a null thread argument).

src/unix/blockq.c
```diff
@@ -294,80 +213,80 @@ void kern_blockq_handle_completion(blockq bq, u64 bq_flags, io_completion comple
...
+    do {
+        u64 saved_flags = spin_lock_irq(&bq->lock);
+        list l = list_get_next(&bq->waiters_head);
+        if (!l) {
+            spin_unlock_irq(&bq->lock, saved_flags);
+            return;
+        }
+        list_delete(l);
+        spin_unlock_irq(&bq->lock, saved_flags);
+        blockq_flush_thread(bq, struct_from_list(l, thread, bq_l));
```
francescolavra, line 233: blockq_flush_thread() calls blockq_wake_thread_internal(), which asserts that `(t->bq_l.next && t->bq_l.prev)` (and also tries to remove the thread from the waiter list). I think we should just call blockq_apply() instead of blockq_flush_thread().

src/unix/signal.c
wjhun: Yes, I agree that shouldn't be conditional. Fix applied.

src/unix/blockq.c
wjhun: Ok, I changed the call to blockq_flush_thread to a call to blockq_apply. Thanks.

sanderssj:
LGTM - My only question is about the necessity of blockq_{reserve,release} (or maybe even refcounting?). Are you anticipating needing them? As it is, the one call to blockq_release could just be a refcount_release unless you wanted to keep the interface.

francescolavra:
blockq_{reserve,release} will be needed in thread_attempt_interrupt() when the kernel lock is removed, see https://github.com/nanovms/nanos/commit/e96c3dacea6503e00720b88822051eb3752e283c#diff-bccf980566899b9b58dc0228f7537ac1c2d1c996a075139baf298dafaae5efaaR279-R287

sanderssj:
> blockq_{reserve,release} will be needed in thread_attempt_interrupt() when the kernel lock is removed, see [e96c3da#diff-bccf980566899b9b58dc0228f7537ac1c2d1c996a075139baf298dafaae5efaaR279-R287](https://github.com/nanovms/nanos/commit/e96c3dacea6503e00720b88822051eb3752e283c#diff-bccf980566899b9b58dc0228f7537ac1c2d1c996a075139baf298dafaae5efaaR279-R287)

I see, thanks!

PR #1570, opened 2021-08-31
Unix domain sockets: fix setting of filesystem metadata

francescolavra:
The trick currently being used in unixsock_bind() to prevent the address of a bound socket from being stored in the filesystem log is flawed, because if the filesystem is compacted the newly built log will contain the socket address, which can cause invalid memory accesses if the program tries to connect to the socket after a reboot.
This commit changes the format of the filesystem metadata that encodes the address of a bound socket, using sym(no_encode) to prevent the tuple containing the address from being written to the TFS log.
Direct access to filesystem metadata from the Unix socket code (which will be unsafe when the kernel lock is removed) is replaced by the filesystem_mk_socket(), filesystem_get_socket() and filesystem_clear_socket() functions.

The first commit prevents leaking filesystem metadata if a socket file is unlinked from the filesystem.

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -1618,6 +1618,66 @@ int filesystem_follow_links(filesystem *fs, tuple link, tuple parent,
     }
 }
 
+fs_status filesystem_mk_socket(filesystem *fs, tuple cwd, const char *path, void *s, tuple *t)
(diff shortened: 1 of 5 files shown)


wjhun (approved):
LGTM

PR #1572, opened 2021-09-10
Thread and process structures: protect against concurrent access

francescolavra:
This change adds a spinlock to the thread and process structures, and uses this spinlock to protect structure members against concurrent access.
Also, thread_from_tid() now increments the thread reference count, to prevent a thread from being deallocated while it is being accessed; callers of thread_from_tid() need to call thread_release() when they are done with the thread.
The code that needs to retrieve the current working directory and its filesystem now calls process_get_cwd(), which retrieves this information atomically with respect to the chdir and fchdir syscalls.
For x86_64, the virtual32 heap has been switched from non-locking to locking.

### DIFF
--- a/src/unix/blockq.c
+++ b/src/unix/blockq.c
@@ -254,8 +254,10 @@ int blockq_transfer_waiters(blockq dest, blockq src, int n)
                                             dest, t));
         }
         list_delete(&t->bq_l);
+        thread_lock(t);
         assert(t->blocked_on == src);
         t->blocked_on = dest;
+        thread_unlock(t);
         list_insert_before(&dest->waiters_head, &t->bq_l);
         transferred++;
     }
--- a/src/unix/signal.c
+++ b/src/unix/signal.c
@@ -410,7 +410,7 @@ static void check_syscall_restart(thread t, sigaction sa)
 sysreturn rt_sigreturn(void)
 {
     struct rt_sigframe *frame;
-    sigaction sa;
+    struct sigaction sa;
     thread t = current;
 
     assert(t->dispatch_sigstate);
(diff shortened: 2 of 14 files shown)


src/unix/signal.c
```diff
@@ -360,11 +364,10 @@ void deliver_signal_to_process(process p, struct siginfo *info)
         /*  Attempt to deliver via signalfd notify. If the
             thread becomes runnable, we're done.  */
         signalfd_dispatch(can_wake, sigword);
-        if (thread_is_runnable(can_wake))
-            return;
         sig_debug("attempting to interrupt thread %d\n", can_wake->tid);
         if (!thread_attempt_interrupt(can_wake))
             sig_debug("failed to interrupt\n");
```
wjhun, line 369: This is only a debug message, so it should be no issue, but it seems that a thread could possibly wake up between the point that we select it above (being in interruptible sleep) and attempt to interrupt it here - thus making a false return valid here. For that matter, it may be necessary to place a thread_lock / unlock in thread_in_interruptible_sleep around the t->blocked_on checks.

wjhun (approved):
LGTM, just minor comments

src/unix/blockq.c
```diff
@@ -195,8 +195,11 @@ sysreturn blockq_check_timeout(blockq bq, thread t, blockq_action a, boolean in_
     u64 saved_flags = spin_lock_irq(&bq->lock);
     list_insert_before(&bq->waiters_head, &t->bq_l);
     spin_unlock_irq(&bq->lock, saved_flags);
-    if (!in_bh)
+    if (!in_bh) {
+        thread_lock(t);
```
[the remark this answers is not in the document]
francescolavra: Somehow I was under the impression that not locking the thread here could race with thread_attempt_interrupt(). But now I see that it should be fine to change t->blocked_on without the lock as long as the previous value is not a valid bq pointer (if the previous value is a bq pointer, we must take the lock in order to give a chance to concurrent code to reserve the bq before it disappears).
I removed the lock here.

src/unix/signal.c
francescolavra: thread_in_interruptible_sleep() is called by code (like thread_attempt_interrupt) with the thread lock already held, so we can't place thread_lock / unlock in this function.
Yes, it is possible that a thread wakes up between the point it's selected and the attempt to interrupt it; but this is the case even with the kernel lock, because a thread can be woken up by a bhqueue thunk running without the kernel lock. If that happens, the signal will be delivered at a later point, most likely the next time a thread (possibly a different one) is about to be scheduled again.
Anyway, I re-added the call to thread_is_runnable() so that if the thread is woken up via signalfd we don't attempt to interrupt it.

PR #1574, opened 2021-09-15
synchronization for lwIP and associated networking code

wjhun:
In support of the effort to eliminate dependence on the kernel lock, these changes propose to add the necessary synchronization to networking code within the kernel. Specifically, all entry points into lwIP are protected by an lwIP lock, and network sockets contain an additional lock to protect the dequeueing of received data. This also fixes a race in connect_tcp_bh whereby a connect syscall could be interrupted but the pending connection completed before the nullify action. In this case, the connection is simply closed and socket returned to TCP_SOCK_CREATED. In the direct net code, received pbufs are now put on a queue for deferred processing rather than handled directly within the lwIP receive callback. This avoids the possibility of a deadlock in the typical scenario that the receive buffer handler attempts to synchronously send data in response.

The AWS ena driver contained a bug whereby link status changes would be propagated to lwIP at interrupt level. To remedy this, link status changes are now performed from deferred tasks that are embedded into the ena_adapter struct.

To aid in the debugging of spinlock deadlocks, the debug spinlocks (enabled with SPIN_LOCK_DEBUG_NOSMP) now store the program address of the code taking the lock.


src/net/netsyscall.c
```diff
@@ -423,8 +436,13 @@ static sysreturn sock_read_bh_internal(netsock s, thread t, void * dest,
                 length -= xfer;
                 xfer_total += xfer;
                 dest = (char *) dest + xfer;
-                if ((s->sock.type == SOCK_STREAM) && !(flags & MSG_PEEK))
+                if ((s->sock.type == SOCK_STREAM) && !(flags & MSG_PEEK)) {
+                    if (!blocked)
+                        lwip_lock();
```
francescolavra, line 441: This could deadlock with a parallel thread where sock_read_bh_internal() is called with lwIP already locked and is trying to acquire s->consume_lock. To prevent this deadlock, the lwIP lock should be acquired before s->consume_lock.

francescolavra (commented):
In netsock_ioctl(), the socket consume_lock should be acquired when handling FIONREAD requests, because the code checks pbuf structs from the incoming queue.

klib/tun.c
```diff
@@ -126,7 +128,9 @@ static err_t tun_if_output(struct netif *netif, struct pbuf *p, const ip4_addr_t
         f = struct_from_list(next, tun_file, l);
     } while (f != t->next_tx);
     if (selected && kfuncs.enqueue(selected->pq, p)) {
+        lwip_lock();
```
francescolavra, line 131: kfuncs.lwip_lock()

francescolavra:
LGTM

klib/tun.c
```diff
@@ -126,7 +128,9 @@ static err_t tun_if_output(struct netif *netif, struct pbuf *p, const ip4_addr_t
...
         kfuncs.pbuf_ref(p);
+        lwip_unlock();
```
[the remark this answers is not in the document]
wjhun: Ok, I have addressed the klib build issue in the fix-klib-smp-build branch. Incidentally, I also made "KERNEL" defined for all kernel code builds, whether klib or the main kernel binary. This eliminates the need for macros to check both KERNEL and KLIB to determine if code is being compiled to run at kernel level.

As for the lock/unlock, it turns out that these aren't valid here anyhow, as the output function is called from lwIP, and thus with the lock already held. I also updated the blockq action to only conditionally take the lwIP lock if the action is not called with the blocking flag (similar to the bq actions in netsyscall). Also BLOCKQ_ACTION_NULLIFY was not being handled there, so I added that as well.

PR #1575, opened 2021-09-17
Unix special files: add /proc/mounts

francescolavra:
This file contains the list of currently mounted filesystems, using same syntax as used in the Linux /proc/mounts file.


wjhun (approved):
LGTM

PR #1576, opened 2021-09-17
Unix file descriptors: allocate after full initialization

francescolavra:
When allocate_fd() is called, the newly allocated file descriptor number can be accessed by userspace, thus the corresponding kernel structure must be fully initialized to properly handle any syscalls that may be invoked in the file descriptor.
This change makes file descriptor allocation the last step in the initialization sequence of various syscalls, to make it safe in SMP systems without the kernel lock. Most file descriptor allocation errors can now be handled by simply calling the close() function (which will properly clean up the just initialized data structure) and returning -EMFILE.
In the socket code, file descriptor allocation has been removed from socket_init(), and added to its callers after full initialization of socket data.

### DIFF
--- a/src/unix/socket.h
+++ b/src/unix/socket.h
@@ -82,11 +82,6 @@ static inline int socket_init(process p, heap h, int domain, int type, u32 flags
         struct sock *s)
 {
     runtime_memset((u8 *) s, 0, sizeof(*s));
-    s->fd = allocate_fd(p, s);
-    if (s->fd == INVALID_PHYSICAL) {
-        msg_err("failed to allocate fd\n");
-        goto err_fd;
-    }
     s->rxbq = allocate_blockq(h, "sock receive");
     if (s->rxbq == INVALID_ADDRESS) {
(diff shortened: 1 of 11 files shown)


sanderssj (commented):
LGTM

src/unix/socket.h
```diff
@@ -82,11 +82,6 @@ static inline int socket_init(process p, heap h, int domain, int type, u32 flags
         struct sock *s)
 {
     runtime_memset((u8 *) s, 0, sizeof(*s));
-    s->fd = allocate_fd(p, s);
```
wjhun, line 85: Nit: I realize s->fd will be zero after the memset, but maybe it's a bit clearer to just return 0 below on success.

wjhun (approved):
LGTM

src/unix/socket.h
francescolavra: Agreed, returning s->fd is misleading, I changed it to return 0

PR #1578, opened 2021-09-19
delay storage operations for an extent that is transitioning from uninited to initialized state

### LINKED ISSUE #1571: occasional failure of fallocate test
The fallocate runtime test will, on occasion (say, one out of every dozen runs), exit with the following assertion failure:

```Error: buf[i] == 0 -- failed at /home/wjhun/src/nanos/master4/test/runtime/fallocate.c:32```

This appears to happen in the test_tmpfile during the loop that validates that all data read after the written first byte appears as zero. These zeros are not written explicitly but rather assumed to be zero after the call to fallocate(). I have only seen this failure when running in qemu under Linux with KVM and under macOS with hvf, and have not been able to reproduce it on either platform using TCG.

After further inspection, it appears that the buffer byte location that throws the assertion contains 0xff, which is the pattern that is first used to fill the buffer. However, if I change the fill pattern to something else (and change the first assertion for the written byte accordingly), the failure mode still shows 0xff bytes. Incorporating a check in sg_zero_fill(), called by tfs.c:zero_hole() and tfs:read_extent(), validates that the pagecache buffer is filled with zeros and that its address and length are correct. Also note that the area filled with the 0xff pattern seems to always begin and end on sector boundaries, though the extent of this area seems to change with each failure.

wjhun:
When a file extent that is marked as uninited is written to, all sectors
corresponding to the extent are written, including both the data from the
write request and any zeroed area(s) surrounding it. This presents a problem
when subsequent storage requests are issued on behalf of the extent. Such
requests may be made without first waiting for completion of the initializing
write. This property of the pagecache - that requests are issued without first
waiting for completions for prior requests - is typically not an issue, for
read data is cached and not re-read after a write, and writes are posted from
the same pagecache buffer, making the order of write completions
irrelevant. But the areas of an extent outside of the initial write request,
which are written with zeros, could be subsequently read into the cache (or
pages overwritten as a whole) prior to the zero writes completing. This has
been found to lead to spurious inconsistent data being read - namely, areas
that should be zero are read and found to contain nonzero data.

To ameliorate this condition, the uninited flag has been changed into an
optional record that becomes allocated when a write is first posted to an
uninited extent. When subsequent storage requests are made to the extent
before the initializing writes complete, they are added to a queue within the
uninited record. On completion of the initializing writes, the queued
requests are then dispatched, and the uninited state is set to 'initialized',
indicating that further operations on the extent should occur normally, as if
uninited was not set. (Given that extent references are not tracked and that
the extents themselves may be deallocated at any time, the uninited completion
never refers to the extent itself, and the record exists for the lifetime of
the extent - thus the uninited 'initialized' state having the same semantics
as a null uninited field.)

Resolves #1571

### DIFF
--- a/src/tfs/tfs.c
+++ b/src/tfs/tfs.c
@@ -208,9 +208,17 @@ static inline extent allocate_extent(heap h, range file_blocks, range storage_bl
     rmnode_init(&e->node, file_blocks);
     e->start_block = storage_blocks.start;
     e->allocated = range_span(storage_blocks);
-    e->uninited = false;
+    e->uninited = 0;
     return e;
 }
+
+#ifdef KERNEL
+#define uninited_lock(u) spin_lock(&(u)->lock);
+#define uninited_unlock(u) spin_unlock(&(u)->lock);
+#else
+#define uninited_lock(u)
+#define uninited_unlock(u)
+#endif
 
 u64 filesystem_allocate_storage(filesystem fs, u64 nblocks)
 {
@@ -256,7 +264,7 @@ void ingest_extent(fsfile f, symbol off, tuple value)
         halt("out of memory\n");
     ex->md = value;
     if (get(value, sym(uninited)))
-        ex->uninited = true;
+        ex->uninited = INVALID_ADDRESS;
     assert(rangemap_insert(f->extentmap, &ex->node));
 }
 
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
Looks good to me

PR #1579, opened 2021-09-21
fix SMP build for klibs and correct process lock init in create_process()

wjhun:
As noted in https://github.com/nanovms/nanos/pull/1574#discussion_r710147747, the build flag SMP_ENABLE was not being set for klib builds, leading to compilation of no-op versions of inlined locking functions used for SMP builds. This PR fixes that issue and, for convenience, adds a "NOSMP" flag that can be specified on the make commandline to universally disable SMP support and enable spinlock debugging (albeit with asserts omitted for klib builds - see comments). Also, the semantics of the "KERNEL" define have changed to mean any code that is built for use at kernel level (excepting bootloaders); KERNEL is now defined for klib builds as well as the main kernel binary build.

In the process of testing using spinlock debugging, an error was found and corrected in create_process() whereby the process spinlock could be used (by allocate_fd()) prior to being initialized. The fix is to simply move the call to spin_lock_init() to the top of the function.


sanderssj (commented):
LGTM

PR #1580, opened 2021-09-22
TFS: add protection against concurrent execution

francescolavra:
This will allow filesystem-related syscalls and I/O completions to be run without the global kernel lock.
Syscalls that modify the filesystem and take a file path as argument have been changed so that file path resolution and handling of filesystem tuples is done inside the TFS code, where the filesystem is locked; in cases where reading of filesystem metadata is best done at the syscall level, the new filesystem_get_node() function is now called: this function resolves the supplied path to a tuple (possibly creating a new file in the filesystem) and returns that tuple (as well as an fsfile pointer, if applicable), which can be safely accessed from external code until filesystem_put_node() is called.
To correctly handle cases where a filesystem tuple reference held by a file descriptor becomes invalid, most filesystem functions have been modified to take inode numbers instead of tuple pointers, and validate the inode numbers to ensure that they refer to valid filesystem entries; to implement that, all directory entries (not just regular files) are put in the internal TFS file table. Regular files are distinguished from non-regular files by the fact that the latter have a file table entry value set to INVALID_ADDRESS (as opposed to an fsfile pointer). The new filesystem_get_meta() and filesystem_put_meta() functions have been added to allow safe access to filesystem tuples corresponding to inode numbers.

### DIFF
--- a/klib/radar.c
+++ b/klib/radar.c
@@ -396,7 +396,7 @@ static void telemetry_boot(void)
 
 closure_function(2, 4, void, telemetry_vh,
                  buffer, b, int, count,
-                 u8 *, uuid, const char *, label, filesystem, fs, tuple, mount_point)
+                 u8 *, uuid, const char *, label, filesystem, fs, inode, mount_point)
 {
     u64 block_size = kfunc(fs_blocksize)(fs);
     buffer b = bound(b);
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -248,7 +248,7 @@ static void __attribute__((noinline)) init_service_new_stack()
     init_debug("in init_service_new_stack");
     init_page_tables((heap)heap_linear_backed(kh));
     init_tuples(allocate_tagged_region(kh, tag_table_tuple));
-    init_symbols(allocate_tagged_region(kh, tag_symbol), heap_general(kh));
+    init_symbols(allocate_tagged_region(kh, tag_symbol), heap_locked(kh));
 
     for_regions(e) {
         if (e->type == REGION_SMBIOS) {
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -247,7 +247,8 @@ static void __attribute__((noinline)) init_service_new_stack()
     kernel_heaps kh = get_kernel_heaps();
     init_debug("in init_service_new_stack");
     init_page_tables((heap)heap_linear_backed(kh));
-    init_tuples(allocate_tagged_region(kh, tag_table_tuple));
+    init_tuples(locking_heap_wrapper(heap_general(kh),
+                allocate_tagged_region(kh, tag_table_tuple)));
     init_symbols(allocate_tagged_region(kh, tag_symbol), heap_locked(kh));
 
     for_regions(e) {
--- a/platform/virt/service.c
+++ b/platform/virt/service.c
@@ -226,7 +226,8 @@ static void __attribute__((noinline)) init_service_new_stack(void)
     init_page_tables((heap)heap_linear_backed(kh));
     /* mmu init complete; unmap temporary identity map */
     unmap(PHYSMEM_BASE, INIT_IDENTITY_SIZE);
-    init_tuples(allocate_tagged_region(kh, tag_table_tuple));
+    init_tuples(locking_heap_wrapper(heap_general(kh),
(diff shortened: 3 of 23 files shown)


francescolavra:
I added a commit that is needed to make allocation and deallocation of file descriptors (of any type, not just filesystem entries) safe without the kernel lock.

sanderssj:
LGTM. The tfs code has a lot of the filesystem_get_meta calls that are "undone" only with filesystem_unlock. The filesystem_resolve_cstring is similar but with filesystem_reserve/release. Both those functions are commented appropriately about that, but I was wondering if the names should also reflect it? The names would get awfully long. I'm not asking to change it, but wanted to put it out there. Is there also a danger that filesystem_get/put_meta get something additional in the future and these get_meta calls would only be mirrored by the unlock?

Also something I noticed that isn't related to the change, but should volume_link call filesystem_release if the mount fails?

francescolavra:
> LGTM. The tfs code has a lot of the filesystem_get_meta calls that are "undone" only with filesystem_unlock. The filesystem_resolve_cstring is similar but with filesystem_reserve/release. Both those functions are commented appropriately about that, but I was wondering if the names should also reflect it?

I've been struggling to find names that better reflect what the functions do, if anyone has suggestions they are welcome of course.

> Is there also a danger that filesystem_get/put_meta get something additional in the future and these get_meta calls would only be mirrored by the unlock?

In most cases, the calls to filesystem_get_meta in tfs.c are made to get the tuple of the working directory, but then the filesystem can change if the file path resolution crosses a mount point, so replacing the unlock at the end of a function with a call to filesystem_put_meta would be conceptually wrong (although it would work in practice), because we could be calling filesystem_put_meta with an `fs` argument value potentially different from the value passed to filesystem_get_meta.
In the future we might want to change filesystem_get_meta so that it doesn't lock the filesystem (to do that, we would need some sort of locking/refcounting for individual filesystem entries), but most of the code in tfs.c would still need to lock the filesystem to make changes to it, so I imagine the final calls to filesystem_unlock() would still be needed.

> Also something I noticed that isn't related to the change, but should volume_link call filesystem_release if the mount fails?

Indeed the filesystem should be released if a mount fails, otherwise it would be leaked because no code would hold a reference to it. I'll fix that in a separate PR.

sanderssj:
> I've been struggling to find names that better reflect what the functions do, if anyone has suggestions they are welcome of course.
I don't know that I have any great ideas either. I thought about something like "filesystem_lock_and_get_meta" but I think that's a little clumsy and makes the function intent harder to read.

> In most cases, the calls to filesystem_get_meta in tfs.c are made to get the tuple of the working directory, but then the filesystem can change if the file path resolution crosses a mount point, so replacing the unlock at the end of a function with a call to filesystem_put_meta would be conceptually wrong (although it would work in practice), because we could be calling filesystem_put_meta with an `fs` argument value potentially different from the value passed to filesystem_get_meta. In the future we might want to change filesystem_get_meta so that it doesn't lock the filesystem (to do that, we would need some sort of locking/refcounting for individual filesystem entries), but most of the code in tfs.c would still need to lock the filesystem to make changes to it, so I imagine the final calls to filesystem_unlock() would still be needed.

Makes sense, thanks!

francescolavra:
> Is the inode is really a sort of opaque type that is used to force the users to go through the get/put functions in order to add locking as well as validate the inode in the files table? 
> Also, my understanding (maybe dated?) is that an inode number is a kind of serial number that is consistent across multiple mounts of a filesystem.

In general I think of an inode as just a unique identifier of a filesystem entry, with no further constraints. It appears that usually it's a consistent number across mounts for disk filesystems, but not necessarily so for other types of filesystems such as network filesystems. So in general an application cannot rely on inode numbers to be fixed.
In our kernel we could reduce the usage of get_meta/put_meta if we stored more metadata in the file struct, but still we couldn't do everything without get_meta/put_meta, because for example when we read a directory contents with getdents() we have to rely on an inode number to fetch data from the filesystem, and I think all other OSes do the same (a directory could be removed after having been opened, and in that case getdents() returns -ENOENT).
Anyway, in the future we could convert inode numbers to be actual indices, that would require the TFS code to have (beside the existing table for regular files) an array where we store the tuple pointers for all filesystem entries.

PR #1584, opened 2021-09-24
Network sockets: replace non-locked heaps with locked heaps

francescolavra:
This is necessary when executing network socket syscalls without the global kernel lock.


wjhun (approved):
Looks good

PR #1585, opened 2021-09-28
Unix sockets: fix usage of transmission blockq

francescolavra:
When a socket write operation is carried out, it should be tied to the tx blockq of the destination socket, not the one of the sender, so that if the destination queue becomes full and a write operation blocks, the receiver can notify the sender via blockq_wake_one() when the queue becomes writable again, without needing a reference to the sender (for datagram sockets, there is no such reference).
In addition, for poll operations to be supported, the sender should add a notify handler to the receiver notify set.
Whenever a reference to a peer socket is being used (such as during a connect or a sendto), its reference count is now incremented, regardless of the socket type, so that it is safe to access the peer socket structure even after the peer socket has been closed (and it is possible to detect if the peer socket has been closed by checking the data field). Socket path resolution has been moved from unixsock_write_bh() to unixsock_sendto(). These changes will make it easier to implement multi-thread safety without the global kernel lock.
Handling of connect() and poll() operations, including various error conditions, has been revised to make it more consistent with the Linux implementation, and a number of test cases have been added to the runtime test.


src/unix/socket.c
```diff
@@ -411,19 +453,14 @@ closure_function(1, 2, sysreturn, unixsock_close,
...
-        if (s->peer->data) {
-            socket_flush_q(&s->peer->sock);
-            fdesc_notify_events(&s->peer->sock.f);
-        }
-    }
     if (s->conn_q) {
         /* Notify any connecting sockets that connection is being refused. */
         unixsock child;
         while ((child = dequeue(s->conn_q)) != INVALID_ADDRESS) {
-            child->connecting = false;
```
sanderssj, line 425: Should this (or rather child->peer connecting status) actually be removed? Unless I've missed it somewhere else, seems like it should go with the conn_status change below or else moved somewhere else like when returning the failed connection from the bottom half. Otherwise you can't call connect again with this socket.

sanderssj (commented):
LGTM other than one question about failed connections.

src/unix/socket.c
francescolavra: This child socket is not the one the requested the connection, but the one that was created as a result of a connection request and is about to be deallocated a few lines below, so setting child->connecting = false is useless (in fact, before these changes a socket blocked in a connect() would hang indefinitely if the listening socket was closed).
We are not setting connecting = false on the connecting socket because we need to keep track of the fact that the connection attempt is still in progress, in order for unixsock_events() to return the correct value if the connect is being done in a non-blocking fashion. `connecting` is set to false in accept_bh; if the listening socket is closed and as such there is no accept() call, `connecting` will be left with a true value, but connect() can still be called, because unixsock_is_connecting() now returns (s->connecting && (s->conn_status == -EINPROGRESS)).

sanderssj: Right, I follow that. I'm thinking of that situation you describe where a connect is attempted but the listening socket is closed before accept. The connection attempt syscall returns -ECONNREFUSED, as set by the closed listening socket. So unixsock_is_connecting() returns false since conn_status is different, but further in unixsock_connect() it checks the connecting boolean to lookup and queue a new connection request. Since connecting is still true, the new connect syscall would do nothing. I was thinking connect_bh could reset connecting when returning a non -EINPROGRESS value, but perhaps the second check for connecting should be changed to unixsock_is_connecting()?

PR #1587, opened 2021-10-02
Page cache: remove runqueue usage for I/O completions

francescolavra:
Now that the filesystem code is multithread-safe, there is no need to use the runqueue protected by the global kernel lock, and the bhqueue can be used for all I/O completions.


wjhun (approved):
LGTM

PR #1588, opened 2021-10-03
aio.c: make code safe to run without the global kernel lock

francescolavra:
The first commit makes looking up and removal of AIO context identifiers multi-thread safe.
The second commit revises locking and adds proper reference counting in the AIO code to make it safe when multiple syscalls and I/O completions can be invoked simultaneously.


sanderssj:
LGTM. Could you add a comment to io_getevents_bh noting the aio lock is usually already held on entry?

francescolavra:
Sure, I added a comment that says that io_getevents_bh() is called with the lock held, unless BLOCKQ_ACTION_BLOCKED is set in the BQ flags.

wjhun (approved):
LGTM

PR #1589, opened 2021-10-03
Volumes: fix filesystem leak on error condition

francescolavra:
The fist commit fixes a deadlock that would happen if a filesystem creation completion is invoked synchronously (e.g. if an error is encountered).
The second commit fixes a memory leak when a filesystem associated to a volume cannot be mounted.


sanderssj:
LGTM

wjhun (approved):
LGTM

PR #1590, opened 2021-10-05
Klibs: fix handling of lwIP lock

francescolavra:
A few changes have been made in net/direct.c to allow TCP clients implemented in the klibs to operate correctly:
- connection establishment and shutdown events are processed by the receive service routine instead of being processed directly by the respective lwIP callbacks
- the receive service routine dequeues all pbufs found in a given queue, instead of just the first one
- direct_conn structures created as a result of a client connection being accepted are added to the connection list of the direct structure, so that they are properly serviced
In addition, a missing call to pbuf_free() in the receive service routine has been added.

The radar, ntp  and cloud_init klibs have been fixed to avoid calling lwip_lock() from within the DNS callback, where the lock is already held.


wjhun (approved):
LGTM

PR #1592, opened 2021-10-05
Use 256 max interrupts on x86

sanderssj:
This change allows nanos to use the full range of interrupt vectors on x86. The idtr is now loaded
from the stack so that the IDT can use the entire page to fit 256 entries.


wjhun (approved):
LGTM

PR #1597, opened 2021-10-08
revised runtime timer interface with locking

wjhun:
In line with the effort to remove dependencies on the kernel lock, this merge
aims to make kernel timers safe to use outside of the domain of the kernel
lock. It consists of three main commits: revisions to the runtime timer interface
(now called timerqueue), the moving of timer service from the runloop to a task
run from the bhqueue, and the addition of locking and refcounting for unix timers
(posix, itimer and timerfd). More detail on these changes may be found in the
respective commit messages.

Incidentally, there seems to be a modest performance improvement when
running webg with SMP and a test of 'ab -n 10000 -c 100'. Taking an average of
multiple tests run on a setup with kvm, with -smp 8 and bridging, the average
number of requests per second increased from 7933 to 8618, or an 8.6%
improvement. The source of this improvement was not isolated, but it is a
reasonable expectation given that timer service may now occur on cores that
are not holding the kernel lock.

### DIFF
--- a/src/unix/blockq.c
+++ b/src/unix/blockq.c
@@ -42,6 +42,9 @@
 #define blockq_debug(x, ...)
 #endif
 
+#define blockq_lock(bq) spin_lock(&bq->lock)
+#define blockq_unlock(bq) spin_unlock(&bq->lock)
+
 /* This applies a blockq action after it has been removed from the
    waiters list. If the action indicates that waiting should continue,
    it re-adds the thread to the waiter list and returns false. If the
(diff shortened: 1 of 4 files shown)


wjhun:
I have pushed two more commits. The first one addresses a rather heinous bug caused by the use of heap_general in allocating blockq_action closures in nanosleep and clock_nanosleep. Now that timer handlers are run outside of the kernel lock, the closures were deallocating themselves from heap_general without the protection of the kernel lock, wreaking havoc in the objcache allocators and causing occasional crashes when running the webg test with multiple cores.

The second commit addresses a fundamental race issue in the blockq code that was found by inspection. The problematic scenario is described in the commit log. In a nutshell, the change attempts to correct usage of the thread lock in order to prevent concurrent but independent wakeup operations from interfering with each other.

PR #1614, opened 2021-10-26
Allow /proc/mounts to notify pollers of mount change

sanderssj:
The /proc/mounts file has special behavior where a change to the file
from a mount or unmount causes poll to return POLLPRI and POLLERR. This
can be used to monitor the mount status of the system. This change
implements this feature by adding a generation number to track changes
to mounts and calling a list of registered handlers whenever a mount
changes. The /proc/mounts special file registers a handler that checks
to see if the generation number has changed and calls notify_dispatch
to send the event to waiters on the file. In order to track the special
file state, a new pointer field is added to the fdesc struct for holding
arbitrary data.

There are also some small fixes to the mounts_handler that generates
the /proc/mounts contents.

### DIFF
--- a/src/unix/special.c
+++ b/src/unix/special.c
@@ -10,7 +10,13 @@ typedef struct special_file {
     sysreturn (*read)(file f, void *dest, u64 length, u64 offset);
     sysreturn (*write)(file f, void *dest, u64 length, u64 offset);
     u32 (*events)(file f);
+    u64 alloc_size;
 } special_file;
+
+typedef struct special_file_wrapper {
+    struct file f;
+    u64 alloc_size;
+} *special_file_wrapper;
 
 static sysreturn urandom_read(file f, void *dest, u64 length, u64 offset)
 {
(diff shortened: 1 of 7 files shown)


src/unix/unix_internal.h
```diff
@@ -390,6 +390,7 @@ typedef struct fdesc {
     int type;
     int flags;                  /* F_GETFD/F_SETFD flags */
     notify_set ns;
+    void *private;
```
francescolavra, line 393: Wouldn't it be better to put this field in struct file instead? fdesc is used as a base struct for all types of file descriptors, so I think private data specific for a given descriptor type should be embedded in the struct that extends this base struct for that descriptor type.

src/unix/special.c
```diff
@@ -100,20 +100,24 @@ closure_function(1, 4, void, mounts_handler,
...
-        while (true) {
-            int rv = file_get_path(fs, mount_point, buffer_end(b), buffer_space(b));
+        for (int count = 2; count > 0; count--) {
+            int rv = file_get_path(get_root_fs(), mount_point, buffer_end(b), buffer_space(b));
             if (rv > 0) {
                 buffer_produce(b, rv - 1);  /* drop the string terminator character */
-                break;
-            } else if (!buffer_extend(b, 2 * buffer_space(b))) {
+                goto out;
+            } else if (!buffer_extend(b, 2 * buffer_length(b))) {
```
francescolavra, line 108: If file_get_path() fails on the first non-root volume due to insufficient buffer space, this code would call buffer_extend() when buffer_length() is strlen("root / tfs rw 0 0\n", i.e. 18, and since the available buffer space is (64 - 18) = 46 a buffer_extend() call with 2 * buffer_length() = 36 as argument value would be a no-op, so the second attempt at file_get_path() will also fail and this volume would be skipped. So doing just two iterations wouldn't be enough. Maybe we could iterate until the available buffer space is PATH_MAX, or alternatively call buffer_extend() with PATH_MAX as argument.

src/unix/unix_internal.h
sanderssj: You're right, in fact I decided to rework the special file allocation so that no new fields needed to be added to file or fdesc structs, so special files can now have their own struct containing a file struct.

src/unix/special.c
sanderssj: I went with calling buffer_extend with PATH_MAX to avoid having to deal with multiple extend allocations.

wjhun (approved):
LGTM

PR #1615, opened 2021-10-30
klibs: implement dynamic linking to kernel

francescolavra:
This PR implements dynamic linking of klibs to the kernel, which uses the symbol definitions found in the kernel ELF file and removes the need for klibs to look up or define symbols explicitly.
Symbols defined in a klib are added to the main symbol table when the klib has been loaded and successfully initialized (so that they can be looked up by any klib that needs them), and are removed from the symbol table when a klib is unloaded.

In order to prevent the linker from discarding functions that are defined in the kernel but called exclusively by klibs, the list of dynamic relocation entries found in each klib is retrieved via objdump and transformed (using sed scripts) into a set of EXTERN linker commands, which are put in an auto-generated file that is included by the kernel linker script.

The now() kernel function cannot be dynamically linked to, because it's an inline function that references VDSO variables, which are declared with hidden visibility; for this reason, klibs have to call the kern_now() wrapper function instead of now().

Since loading a klib requires the kernel symbol table to be populated, if the kernel is not booted by stage2 (e.g. with Firecracker) klib initialization is now done after loading the symbols from the kernel ELF file, thus the boot filesystem needs to be read regardless of whether the klibs are in that filesystem.

The first commit is a fix needed for the cloud_init klib to work correctly when multiple downloads are specified in the configuration.
The third commit adds stack smashing protection to klibs, which is now possible because the symbols used by the stack protector can be resolved when a klib is loaded.

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -155,8 +155,6 @@ endif
 
 CFLAGS+=	$(KERNCFLAGS) -O3 $(INCLUDES) -fPIC $(DEFINES)
 
-# TODO should add stack protection to klibs...
-CFLAGS+=	-fno-stack-protector
 LDFLAGS+=	-shared -Bsymbolic -nostdlib -T$(ARCHDIR)/klib.lds
 
 CLEANFILES+=	$(KLIB_SYMS)
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -116,6 +116,8 @@ SRCS-mbedtls-tls= \
 
 all: $(PROGRAMS)
 
+KLIB_SYMS= $(OBJDIR)/klib-syms.lds
+
 include ../rules.mk
 
 ifeq ($(UNAME_s),Darwin)
@@ -155,6 +157,7 @@ CFLAGS+=	$(KERNCFLAGS) -O3 $(INCLUDES) -fPIC $(DEFINES)
 
 # TODO should add stack protection to klibs...
 CFLAGS+=	-fno-stack-protector
-LDFLAGS+=	-pie -nostdlib -T$(ARCHDIR)/klib.lds
+LDFLAGS+=	-shared -Bsymbolic -nostdlib -T$(ARCHDIR)/klib.lds
 
+CLEANFILES+=	$(KLIB_SYMS)
 CLEANDIRS+=	$(OUTDIR)/klib/vendor $(OUTDIR)/klib/vendor/mbedtls
--- a/rules.mk
+++ b/rules.mk
@@ -106,7 +106,7 @@ KERNCFLAGS+=    -mno-sse \
 		-mno-sse2
 endif
 KERNCFLAGS+=	-fno-omit-frame-pointer
-KERNLDFLAGS=	--gc-sections -n
+KERNLDFLAGS=	--gc-sections -n -L $(OUTDIR)/klib
 
 ifeq ($(MEMDEBUG),mcache)
 CFLAGS+= -DMEMDEBUG_MCACHE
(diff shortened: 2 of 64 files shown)


klib/Makefile
```diff
@@ -155,6 +155,6 @@ CFLAGS+=	$(KERNCFLAGS) -O3 $(INCLUDES) -fPIC $(DEFINES)
 
 # TODO should add stack protection to klibs...
 CFLAGS+=	-fno-stack-protector
-LDFLAGS+=	-pie -nostdlib -T$(ARCHDIR)/klib.lds
+LDFLAGS+=	-shared -Bsymbolic -nostdlib -T$(ARCHDIR)/klib.lds
```
sanderssj, line 156: It looks like -Bsymbolic flag forces resolution of global symbols to try to come from the shared library. Is this necessary because there are some global name conflicts between some of the klibs and the kernel?

rules.mk
```diff
@@ -106,7 +106,9 @@ KERNCFLAGS+=    -mno-sse \
 		-mno-sse2
 endif
 KERNCFLAGS+=	-fno-omit-frame-pointer
-KERNLDFLAGS=	--gc-sections -n
+KERNLDFLAGS=	--gc-sections -n \
+		-u ipaddr_ntoa_r -u dns_gethostbyname -u netif_input -u netif_remove \
```
sanderssj, line 110: This could be way too much hassle so feel free to ignore, but do you think there's a way to autogenerate these? That might even remove the necessity of the special macro and text section. For example, 'nm -u' on the klib could be used to generate a list of names which could be dumped to a file as an EXTERN linker script command (which is supposed to be equivalent of -u) that could be included by the kernel linker script. Provided klibs are built before the kernel, of course.

sanderssj (commented):
LGTM! Just have one question about a flag and a comment.

klib/Makefile
francescolavra: No, I didn't find any global name conflicts between the klibs and the kernel. -Bsymbolic is not strictly necessary, even if there are name conflicts, but removes the need to apply relocations to symbols defined within a klib. Without this flag, when loading a klib, for each symbol defined in the klib we have to put the relocated symbol address in the PLT (Procedure Linkage Table), and there is also a runtime performance penalty because each call to global functions must go through the PLT (much like using a table of function pointers); with -Bsymbolic, all calls to functions defined in a klib are direct calls using position-relative addressing, and the PLT is only used for calls to kernel functions.

wjhun (approved):
LGTM

rules.mk
francescolavra: Now the list of dynamic relocation entries found in each klib is retrieved via objdump and transformed (using sed scripts) into a set of EXTERN linker commands, which are put in an auto-generated file (output/klib/klib-syms.lds) that is included by the kernel linker scripts. so we no longer need to manually add `-u` linker options, nor to use the KLIB_EXPORT #define and the .text.klib linker section.

wjhun: > Now the list of dynamic relocation entries found in each klib is retrieved via objdump and transformed (using sed scripts) into a set of EXTERN linker commands, which are put in an auto-generated file (output/klib/klib-syms.lds) that is included by the kernel linker scripts. so we no longer need to manually add `-u` linker options, nor to use the KLIB_EXPORT #define and the .text.klib linker section.

This is nice. As long as klibs are always built at the same time as the kernel (and within the same tree), I see no problem with this. And if there might ever be an exception to that, we could include a file with a list of symbols to always EXTERN.

sanderssj: Great, thanks!

Issue #1618, opened 2021-11-03
nested signal handling is not supported


wjhun:
For the same cause that is given in #1617, nested signal handling is currently unsupported in Nanos.

wjhun:
Resolved in #1619

Issue #1617, opened 2021-11-03
siglongjmp out of signal handlers is unsupported


wjhun:
The existing signal handling construction involves switching to an alternate frame (the sighandler_frame) when dispatching a signal handler. As such, a siglongjmp out of a signal handler leads to undefined behavior, as rt_sigreturn is bypassed, preventing signal state stored in the thread struct from being reset, and thread execution continues on the sighandler_frame. 

A test using sigsetjmp to save a program frame and siglongjmp to restore the saved frame from a signal handler shows that this sequence will work one time, but another attempt results in the second test signal never being delivered.

This should be solved by moving all thread execution to a single thread frame and reorganizing signal handler dispatch such that no signal-related thread state needs to be reset after signal delivery (aside from restoring a saved signal mask, which sigsetjmp and siglongjmp perform from within userspace).

wjhun:
Resolved in #1619

Issue #1616, opened 2021-11-03
syscall restart logic in rt_sigreturn() assumes a syscall was interrupted


wjhun:
In rt_sigreturn(), check_syscall_noreturn() checks for a -ERESTARTSYS syscall return value in the thread frame and subsequently initiates a syscall restart. This assumes that the signal was dispatched after an interrupted syscall, but a signal could also be dispatched during normal thread entry, e.g. after interrupt service. This should be corrected so that the syscall restart logic is applied only if a syscall was interrupted by a delivered signal.

wjhun:
Resolved in #1619

PR #1619, opened 2021-11-03
fix numerous signal handling issues, use a single context frame per thread

### LINKED ISSUE #1396: complete save/restore for extended mcontext (fp / simd) for aarch64

wjhun:
The previous signal handling construction required switching to an alternate frame (the "sighandler_frame") when dispatching a signal handler. As such, any signal handling pattern which bypasses rt_sigreturn (such as the use of siglongjmp) will not work, as doing so would prevent signal state stored in the thread struct from being reverted, and threads execution would continue on the sighandler_frame rather than reverting to running on to the default_frame. For the same reasons, nested signal handling was also not supported.

This behavior is corrected by moving all thread execution to a single frame (which, incidentally, is also needed for pending context-switching work) and making use of the sigcontext stored on the thread stack to save and restore program state. Special state held for the duration of signal handling has been eliminated from the thread struct, thus enabling the use of siglongjmp.

An issue in syscall restarting has also been resolved. Previously, rt_sigreturn would check if a siginfo contained the SA_RESTART flag, applying a check for -ERESTARTSYS in the position of the thread frame where a syscall return value would be stored. This was done unconditionally, whether a syscall was interrupted or not. This could pose a problem if a thread was interrupted while running and by chance contained the value -ERESTARTSYS in the register used for syscall return values. The solution here is to have blockq_wake_internal_locked() set the "interrupting_syscall" flag in the thread struct, which in turn later triggers the call to check_syscall_restart() from dispatch_signals().

To allow reliance on the stored sigcontext in restoring program state, extended frame data (FP & SIMD state) is now being saved and restored for both x86 and aarch64 architectures.

Tests for sigsetjmp/siglongjmp use in signal handling and nested signal handling have been added to the signal runtime test.

Resolves #1396, #1616, #1617, and #1618

### DIFF
--- a/src/aarch64/crt0.S
+++ b/src/aarch64/crt0.S
@@ -45,10 +45,12 @@
         mrs     x0, spsr_el1
         mrs     x1, esr_el1
         mrs     x2, elr_el1
+        mrs     x3, far_el1
 
         add     x17, x17, #256
         stp     w0, w1, [x17, #((FRAME_ESR_SPSR - 32) * 8)]
         str     x2, [x17, #((FRAME_ELR - 32) * 8)]
+        str     x3, [x17, #((FRAME_FAULT_ADDRESS - 32) * 8)]
         .endm
 
         .text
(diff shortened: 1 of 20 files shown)


sanderssj:
LGTM!

PR #1620, opened 2021-11-04
Initialize atomic_test_*_bit variable for older compilers

sanderssj:
When the compiler does not support setting the carry flag value directly into
a variable for the atomic test bit functions, it uses a second instruction, setc,
to set the variable. This instruction doesn't appear to set the variable when
the carry flag is zero, although the intel manual suggests otherwise. Therefore
the variable must be initialized to zero since it's on the stack to let the
function return the correct value when the bit being checked is zero.

### DIFF
--- a/src/aarch64/machine.h
+++ b/src/aarch64/machine.h
@@ -164,7 +164,7 @@ static inline __attribute__((always_inline)) void memory_barrier(void)
 }
 
 /* XXX something b0rked with ldset here...ordering, or asm constraints? */
-static inline __attribute__((always_inline)) int atomic_test_and_set_bit(u64 *target, u64 bit)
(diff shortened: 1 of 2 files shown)


sanderssj:
> Initializing the variable should be harmless, but I wonder if there is another issue being masked here, as both the Intel and AMD manuals state that SETcc sets the byte regardless.
> 
> Does this work if oldbit is a u8? Also - and I don't think this should be the case as the carry flag is not being modified but only tested - but I wonder if the compiler would behave differently if a "cc" clobber was added in the last argument.

You're right, the variable size is the real problem! I verified that it works properly with an uninitialized u8, and updated the commit

wjhun (approved):
LGTM

PR #1624, opened 2021-11-10
Update xen timers to per-cpu one-shot for better smp support

sanderssj:
The existing timer is unreliable when running with multiple cpus and the
recent timer changes. The timer would not fire predictably on all cpus. The
xen interrupt handler was also assuming the interrupt would run only on cpu
0, which was not guaranteed even before the timer change. CPU initialization
has been refactored out to run for every cpu and the timer is now set by
set_singleshot_timer which is configured to fire on the cpu that set it.
This PR also moves idle_cpu_mask allocation to before the APs get started up,
using the present_processors value which is also now calculated before AP
startup. Otherwise if an AP wakes out of sleep and proceeds to runloop it will
try to access the unallocated mask.

### DIFF
--- a/src/xen/xen.c
+++ b/src/xen/xen.c
@@ -66,9 +66,6 @@ typedef struct xen_platform_info {
     struct list xenbus_list;
     struct list driver_list;
 
-    /* timer */
-    evtchn_port_t timer_evtchn;
-
     /* XXX could make generalized status */
     boolean initialized;
 } *xen_platform_info;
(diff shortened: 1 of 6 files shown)


src/xen/xen.c
```diff
@@ -460,44 +511,13 @@ boolean xen_detect(kernel_heaps kh)
...
-    rv = HYPERVISOR_event_channel_op(&eop);
-    if (rv < 0) {
-        msg_err("failed to bind virtual timer IRQ (rv %d)\n", rv);
-        goto out_unregister_irq;
-    }
-    xen_info.timer_evtchn = eop.u.bind_virq.port;
-    xen_debug("timer event channel %d", xen_info.timer_evtchn);
-    xen_register_evtchn_handler(xen_info.timer_evtchn, closure(xen_info.h, xen_runloop_timer_handler));
-    assert(xen_unmask_evtchn(xen_info.timer_evtchn) == 0);
+    register_platform_clock_timer(closure(xen_info.h, xen_runloop_timer), closure(heap_locked(kh), xen_per_cpu_init, shared_info_phys));
```
wjhun, line 520: xen_info.h is the locked heap

wjhun (approved):
LGTM

PR #1625, opened 2021-11-10
fix altsigstack(2) failure due to an uninitialized variable

wjhun:
Running the fst runtime test with MEMDEBUG=mcache revealed that the
signal_stack_length field in the thread struct was being used without being
initialized in thread_is_on_altstack(). This could cause sigaltstack() to fail
with an -EPERM error for a thread that was not running on an alternate signal
stack. The fix here is to make thread_is_on_altsigstack() first check if
t->signal_stack is set before testing if the thread stack pointer is on the
signal stack.


sanderssj:
LGTM!

PR #1627, opened 2021-11-12
Track frame syscall status via CS register instead of context field

sanderssj:
In x86 arch, a user context can return to user mode by either iret or sysret,
with the latter having less state to reload. Therefore, the user frame must
track whether or not it entered the kernel from a syscall or not. There was a
field in the context struct to track this state, but properly updating when
changing the frame, such as with sighandling, is error prone. If the sysret
path is taken instead of iret, some user frame register values may be
incorrect causing crashes or other bad behavior.
This commit removes the field from the context struct and instead tracks
syscall state through the CPL portion of the CS register. This way the syscall
state moves with the registers through context switches or sighandling
frames. The syscall entry point clears bit 0 of the CPL in the saved frame,
and in frame_return the CPL value is checked. A value of 2 uses the
sysret path while any other value uses the normal iret path. The proper CPL
value of 3 is then restored to the user frame while on the sysret path.

### DIFF
--- a/src/unix/thread.c
+++ b/src/unix/thread.c
@@ -200,8 +200,7 @@ static void setup_thread_frame(heap h, context frame, thread t)
     frame[FRAME_FAULT_HANDLER] = u64_from_pointer(&t->fault_handler);
     frame[FRAME_QUEUE] = u64_from_pointer(current_cpu()->thread_queue);
 #ifdef __x86_64__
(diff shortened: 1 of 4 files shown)


wjhun (approved):
This seems to fix it. Thanks for getting to the root cause!

francescolavra (commented):
LGTM

PR #1628, opened 2021-11-13
Add volume event handling to virtio scsi driver

sanderssj:
Virtio scsi uses the event queue to pass asynchronous information about
volumes added or removed. This change implements handling of these events.
Virtio scsi disks for a scsi device are now stored in a vector for lookup
by the event handler. The rescan event sends an inquiry to the lun specified
in the event, unless than lun is zero, which triggers a report luns for
the target, as indicated by the specification. LUNs that are detected in this
rescan are not re-added if they already exist in the disk list.

I had implemented handling of unit attention sense codes but I could not trigger
them on GCP, so I left it out of this PR as I was unable to test it and that handling
is more complicated.

### DIFF
--- a/src/virtio/virtio_scsi.c
+++ b/src/virtio/virtio_scsi.c
@@ -1,5 +1,6 @@
 #include <kernel.h>
 #include <virtio/scsi.h>
+#include <storage.h>
 
 #include "virtio_internal.h"
 #include "virtio_pci.h"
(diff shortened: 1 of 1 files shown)


src/virtio/virtio_scsi.c
```diff
@@ -105,29 +116,95 @@ struct virtio_scsi {
...
-    virtio_scsi_debug("%s: event 0x%x\n", __func__, bound(e)->event);
+    virtio_scsi_event e = bound(e);
+    virtio_scsi s = bound(s);
+    virtio_scsi_debug("%s: event 0x%x\n", __func__, e->event);
+    if (e->event & VIRTIO_SCSI_EVENT_LOST)
+        msg_err("scsi event reported lost due to missing buffers\n");
+    switch (e->event & ~VIRTIO_SCSI_EVENT_LOST) {
+    case VIRTIO_SCSI_T_TRANSPORT_RESET: {
+        u16 target = e->lun[1];
+        u16 lun = ((e->lun[2] & 0x3f) << 8) | e->lun[3];
```
francescolavra, line 191: e->lun[2] is an u8, so doing `<< 8` will always yield 0.

wjhun (approved):
LGTM

PR #1630, opened 2021-11-18
Make kernel.dis objdump output a separate Makefile target

sanderssj:
Generating the objdump output, kernel.dis, can be a significant portion of time
for incremental compiles and is not necessary for running images. This change
makes a new target to generate the kernel.dis and has the image target generate
only the kernel binary. However, if the variable BUILD_KERNEL_DIS is set then
it will use the old behavior, forcing generation of kernel.dis even if it's not
specified as a make argument. In any case, whenever a new kernel binary is
generated, any existing kernel.dis file is moved to kernel.dis.old so there is no
confusion as to whether or not the disassembly reflects the current binary.


wjhun (approved):
LGTM

PR #1631, opened 2021-12-01
Improve ACPI support with acpica; add PCI hotplug support

francescolavra:
This PR adds full support for ACPI table parsing and ACPI object evaluation and method execution by importing the acpica
library; this brings support for shutdown events and for powering down an instance to platforms that implement ACPI but don't emulate a PIIX4 controller: for example, it is now possible to power down gracefully a Qemu instance by enabling the Qemu monitor and sending a `system_powerdown` command on the monitor interface, and similarly a VirtualBox instance reacts to a shutdown event and is able to power down itself.
Support for ACPI control methods allowed implementing PCI hotplug support, which is used in Nitro-based AWS instances when volumes are attached and detached on a running instance.


wjhun:
This is a lot of vendor code to commit directly into the Nanos tree. Would it instead be possible/practical to pull a tagged acpica as a vendor source on build (like lwIP) and then apply a patchset to it?

francescolavra:
> This is a lot of vendor code to commit directly into the Nanos tree. Would it instead be possible/practical to pull a tagged acpica as a vendor source on build (like lwIP) and then apply a patchset to it?

Yes, I think it's practical to pull a tagged version from the acpica repo (a shallow clone with depth=1 will pull in some 58 MB of stuff) and apply a patchset to it. I will do that and update this PR.

francescolavra:
I removed the acpica code from the kernel source and added it as vendor code, to which a patch is applied (from the  patches/acpica.patch files) after a git clone.

src/kernel/pci.c
```diff
@@ -280,6 +381,22 @@ pci_probe_bus(int bus)
     }
 }
 
+void pci_bus_set_iomem(int bus, id_heap iomem)
+{
+    pci_bus pbus = allocate(pci_buses->h, sizeof(*pbus));
+    assert(pbus != INVALID_ADDRESS);
+    pbus->iomem = iomem;
+    assert(vector_set(pci_buses, bus, pbus));
```
wjhun, line 389: Just to confirm: Does pci_buses does not need to be locked because it is only ever modified on init (scanning of PCI root bridges)?

src/kernel/pci.c
```diff
@@ -259,6 +309,57 @@ static void pci_probe_device(pci_dev dev)
...
+        d->driver_data = 0;
+    } else {
+        driver = 0;
+    }
+    spin_unlock(&pci_lock);
+    if (driver) {
+        if (driver->remove && driver_data)
+            apply(driver->remove, driver_data, completion);
+    } else {
+        apply(completion);
```
wjhun, line 361: Would it be useful to make completion a status_handler and pass some indication of whether the removal actually happened to acpi_eject?

wjhun (approved):
LGTM, just minor comments / questions

src/drivers/acpi.c
```diff
@@ -0,0 +1,518 @@
...
+    length = pad(length + page_offset, PAGESIZE);
+    acpi_debug("  unmapping 0x%lx(%ld)", where, length);
+    unmap(u64_from_pointer(where), length);
+    heap vh = (heap)heap_virtual_page(get_kernel_heaps());
+    deallocate(vh, where, length);
+}
+
+ACPI_STATUS AcpiOsReadMemory(ACPI_PHYSICAL_ADDRESS address, UINT64 *value, UINT32 width)
+{
+    void *v = AcpiOsMapMemory(address, sizeof(UINT64));
```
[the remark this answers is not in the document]
francescolavra: The ACPICA reference manual says that "these interfaces are used by the ACPICA for small amounts of data transfer only, such as memory mapped I/O. For large transfers (such as reading the ACPI tables), the ACPICA code will call AcpiOsMapMemory instead".
In my tests I have never seen AcpiOsReadMemory() or AcpiOsWriteMemory() being called, not even when enabling ACPI events or when such events are generated.

src/kernel/pci.c
francescolavra: Correct, pci_buses is only ever modified when init_acpi_tables() is called during boot

src/kernel/pci.c
francescolavra: As far as I can see, the only case where the completion is invoked after not having removed the device from the list of probed devices is if pci_remove_device() is called more than once for the same device (perhaps the hypervisor might do some sort of retry if the eject doesn't happen after a timeout), and one call occurs simultaneously with the invocation of the completion from a previous call. As such, this seems a pretty unlikely edge case, I would guess the call to the _EJ0 ACPI method would fail but nothing else would happen except a msg_err().
The other case where `driver` is null is when the kernel doesn'tt bind any driver to a given device; in that case, we simply eject the device right away and nothing should change for acpi_eject.
So I don't see an advantage in adding a status to the completion closure.

PR #1633, opened 2021-12-03
Decrease memory utilization

francescolavra:
The PR improves efficiency in utilization of physical memory, which makes the kernel able to run in machines with low amounts of available memory, while not affecting performance in high-memory machines.


wjhun (approved):
LGTM

PR #1634, opened 2021-12-06
C web server: fix memory leaks

francescolavra:
This change set enhances the HTTP parser and network socket implementation so that the C web server (as run with e.g. `make run TARGET=webs`) does not leak memory when handling client requests.


wjhun (approved):
LGTM

PR #1635, opened 2021-12-06
new context type, support for simultaneous blocking page faults, removal of kernel lock

### LINKED ISSUE #1082: explore alternatives to global kernel lock
With the introduction of #1078, the kernel uses a global lock to provide synchronization for kernel data and resources. While this allows simultaneous scheduling of unix threads across cores, it allows only one core to execute kernel code (syscalls and deferred operations off of the bhqueue and runqueue). Resources that require finer-grained locking (e.g. process memory map, physical page allocation, page table, driver ring populating) use spinlocks.

Explore the ground between these two extremes in an effort to increase parallelism of kernel code execution. For example, we could consider locks protecting certain classes of data, e.g. one for lwIP processing, one filesystem operations, etc.

wjhun:
This work introduces a redefined context type that associates a running frame with the
following: methods used when switching contexts (pause, resume, pre_suspend,
schedule_return), a fault handler, and a transient heap that can be used for allocations
confined to the lifetime of the context. kernel_contexts and unix threads now extend this
base type, and the syscall_context type has been created to encompass all syscall
execution. Syscall operations which block awaiting continuation now stamp their respective
context in the continuation closure using contextual_closure(). Areas of the kernel which
service such continuations now invoke them using async_apply*() functions. These
continuations are then invoked directly from the runloop, under the stamped context,
effectively decoupling this context from the one that scheduled the continuation.

Now that each syscall in progress is represented by its own syscall_context, it is
possible to service concurrent page faults for multiple syscalls. unix_fault_handler now
handles faults in a way more orthogonal to the context type, relying on the context
methods to effectively suspend and resume the context when blocking is necessary
(currently just for faults on file-backed memory). With use_fault_handler() and
clear_fault_handler(), handlers may also be temporarily "borrowed" for situations where
user memory may be accessed while running in a kernel context.

The kernel lock has been finally removed now that simultaneous blocking page faults are
supported. In addition, the new context-switching scheme allows for several future
developments:

- Rather than relying on spinlocks to protect very coarse-grained critical sections, a
  mutex type (or other context-blocking lock) may be used to suspend a context while
  waiting for a resource. This can mitigate situations where this is a large degree of
  contention for a single protected resource (e.g. lwIP).

- While instances of the per-context "transient" heap currently point to the locked kernel
  heap, it is possible to use a kind of ephemeral heap that exists only for the lifetime
  of the context. In such an arrangement, allocations are carved out of free pages which
  are later returned at the end of the context lifetime. This allows a simplified
  allocation method (when compared to the mcache allocate) and makes object deallocation
  a no-op, potentially mitigating the performance cost of allocating memory on behalf of
  the context. This would be especially useful for syscall contexts.

- A kernel thread context type is now possible. This would allow tasks running in kernel
  space to be written as free-running threads. Optional services (via klibs) could be
  ported into kernel space without the need to convert them to the asynchronous,
  run-to-completion programming style that is used within the core kernel and syscalls.

- Now that any context may be paused and resumed in a general way, we can revisit enabling
  interrupt handling during kernel execution as well as explore making all kernel paths
  preemptable (should real time considerations arise in the future).

I'm opening this PR now to get a head start on the review process, but there are still a
few items that need addressing before merging:

- Ftrace support needs to be fixed in thread_return: ftrace_thread_switch() expects to
  have the last running thread be specified as "out". With the way context switching works
  now, this could be split into two separate calls that are invoked from thread resume and
  pause methods.

- gdbstub support requires updating / validating

- There is an occasional hang in the signal runtime test when being run with multiple
  processors.

- The inotify test also occasionally crashes or hangs when running with multiple
  processors.

- If a page fault fails for a non-user context (syscall, or kernel context if the fault
  handler is being borrowed), the existing paths will try to deliver a fault signal to the
  thread, but the corresponding syscall context still needs to be terminated and released.

Resolves #1082


francescolavra (commented):
A general observation is that a context switch may require a memory allocation, which could fail in low-memory situations. Ideally the code that suspends a running context (due to a page fault, or a mutex lock operation) should not be allowed to fail.
Perhaps in the future we could revisit this so that each cpu has a default context that is activated any time kernel mode is entered (via interrupt or syscall), and before doing anything that could potentially suspend (like servicing a syscall or a bottom half handler) the code gets/allocates a new context and switches to it (while keeping the default context available); then, any time a context is suspended, control passes back to the default context.

wjhun:
> A general observation is that a context switch may require a memory allocation, which could fail in low-memory situations.

Are you referring to the possible context alloc (via check_{kernel,syscall}_context_replace()) when a context's pre_suspend method is called? That could cause a context to be allocated if none is on the free list - for syscall and kernel contexts. I suppose these free lists could be proactively primed and kept populated.

But if a context is truly going to be suspended (e.g. file-backed fault), other allocations are imminent, and currently there will not be anything like a graceful path to recovery should those paths fail (e.g. on allocating pagecache pages). There are attempts to deliver SIGSEGV to a process (which needs some work as noted in the description), but I am doubtful that even a successful shutdown / flush to disk can succeed in such dire straits.

> Ideally the code that suspends a running context (due to a page fault, or a mutex lock operation) should not be allowed to fail. 

Though I agree in principle, I think that if we get to the point where free memory for kernel use is so low that minor allocations are failing, there really is no hope for a graceful recovery. The approach thus far has been inconsistent to say the least. Throughout the kernel, we try to catch and pass allocation errors up the stack in some places (often in vain, I think) and have asserts in others. I sometimes wonder if we should not just default to asserting and instead work proactively to flush caches, deflate the balloon, if any, and otherwise try to gain resources in order to maintain a minimum level of kernel memory. Below this watermark we really enter the danger zone, at which point we prepare to shut down - kill the process, send logs/metrics via radar and syslog, expedite the writing of dirty pages, flushing the tlog, etc. That should happen long before getting to a point where a context allocation - or really any kernel allocation - could fail.

> Perhaps in the future we could revisit this so that each cpu has a default context that is activated any time kernel mode is entered (via interrupt or syscall)

It is already the current scheme to keep both kernel and syscall contexts at the ready in each cpuinfo. The context allocates occur only during initialization and when a default kernel or syscall context is being replaced (prior to blocking). Each call to runloop() installs the kernel_context stored in the cpuinfo and switches to the top of the corresponding stack. Similarly, on syscall entry, the stored syscall_context is installed.

Interrupt and exception handling don't necessitate a context switch - yet - with the thinking being that both entries are given their own stacks to operate on and all such execution should not cause any faults whatsoever. I suppose installing one could be helpful to catch (error) exceptions that may occur in handlers without clobbering the underlying context frames. This is part of the reason why I left the context argument to the fault handler - to leave the door open should we ever install decide to install a new kernel context before entering the fault handler. (Incidentally, this turned out to be useful for the fault handler 'borrowing' scheme, as the handler does not assume that the faulting context is the enclosed one (the thread) and instead acts on the context argument _with reference to_ the enclosed thread and associated process vmap.)

> and before doing anything that could potentially suspend (like servicing a syscall or a bottom half handler) the code gets/allocates a new context and switches to it (while keeping the default context available); then, any time a context is suspended, control passes back to the default context.

The only difference I see between that and the existing scheme is that the get/allocate is done unconditionally prior to executing a possibly suspending path versus only replacing the default one before a suspend. It would seem that ensuring that the per-cpu syscall and kernel context free lists are at least one level deep prior to executing such paths would be accomplishing the same thing (that is, avoiding context allocation failures from the fault handler). I can take a stab at that - my question now is where it would be appropriate to do such a check (perhaps in the context resumes - during kernel (not exception) execution only).

I still think this is somewhat arbitrary considering that there are other allocations that occur from the fault handler (e.g. allocating new anonymous pages from linear_backed, allocating new pending_fault records - which are only checked with an assert). I think it sort of presses us to develop a coherent strategy for handling low memory in the kernel. Beyond trying to maintain some baseline amount of free memory, is it really worth trying to recover from small allocation failures at all?

francescolavra:
OK, makes sense. Yes, I was referring to the possible context alloc (via check_{kernel,syscall}_context_replace()) when a context's pre_suspend method is called.
In principle, the difference between a page allocation failure on a page fault and an allocation failure of internal kernel data structures is that the former is caused by the user process trying to use memory that has been overcommitted (and thus trying to use more memory that is available in the VM), while the latter could happen even if the process is not doing anything. But in practice there is no real difference, since the amount of free memory varies continuously due to allocs/deallocs by both the kernel and the process, and we cannot say that the "culprit" for a user page allocation failure is the process (or likewise that a kernel data structure allocation failure is the kernel's fault). I guess any such non-recoverable alloc failure can be considered analogous to the Linux OOM killer kicking in.

src/kernel/kernel.c
```diff
@@ -41,44 +23,75 @@ void deallocate_stack(heap h, u64 size, void *stack)
...
+    init_context(c, CONTEXT_TYPE_KERNEL);
+    init_refcount(&c->refcount, 1, init_closure(&kc->free, free_kernel_context, kc, false));
+    c->pause = kernel_context_pause;
+    c->resume = kernel_context_resume;
+    c->schedule_return = kernel_context_schedule_return;
+    c->pre_suspend = kernel_context_pre_suspend;
+    init_closure(&kc->kernel_return, kernel_context_return, kc);
+    c->fault_handler = 0;
+    c->transient_heap = h;
+    void *stack = allocate_stack(h, SYSCALL_STACK_SIZE);
```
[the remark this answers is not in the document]
wjhun: I meant to remove that, but now that you mention it I think it does make sense to combine the (kernel and syscall) context and stack allocations into one. I also moved the allocation for them to the linear backed heap. One potentially useful side effect for debugging is that, given a stack pointer in kernel memory, the corresponding context struct may be easily found by masking off low order bits.

I've also adjusted the default stack size for everything in the kernel to 32KB. (As a point of reference, Linux appears to use 16KB stacks for kernel threads for x86-64 and arm64, based on definitions of THREAD_SIZE.)

src/unix/unix.c
```diff
@@ -412,66 +395,23 @@ void process_get_cwd(process p, filesystem *cwd_fs, inode *cwd)
...
 }
 
 static timestamp stime_updated(thread t)
 {
+    thread_lock(t);
     timestamp ts = t->stime;
-    if (t->sysctx)
-        ts += now(CLOCK_ID_MONOTONIC_RAW) - t->start_time;
+    if (t->syscall && t->syscall->start_time)
+        ts += now(CLOCK_ID_MONOTONIC_RAW) - t->syscall->start_time;
```
francescolavra, line 413: `t->syscall` can be modified without the thread lock, e.g. it's set to 0 in syscall_finish(), so there could be a null pointer access here. In fact, running the apache httpd web server with 2 CPUs I can replicate a condition where this happens.

wjhun: Rather than requiring the thread lock to access t->syscall, I removed utime_updated() and stime_updated() (both of which are unsafe, as the start_time fields are also not protected by the thread or other lock) and have the calling routines simply read t->utime and t->stime directly.

The downside is that the times will not be current to now(), but the assumption here is that the granularity given by times being accumulated on each context switch will give sufficient resolution and accuracy for the dependent syscalls (times, getrusage, and clock_gettime).

PR #1636, opened 2021-12-08
Klibs: fix make dependencies to generate symbol list

francescolavra:
Updating the symbol list in klib-syms.lds after a given klib binary is built does not work well when make is run with multiple parallel jobs, because different klib binaries can be built in parallel and thus the symbol list file can be updated concurrently by multiple jobs, which can lead to inconsistencies in the final contents of the file and ultimately cause klib loading to fail due to missing kernel symbols.
This PR changes the make target for generating the symbol list so that this target has the klib binaries as dependencies, and thus is executed only after all binaries have been built. In addition, the following improvements have been made:
- the symbol list is cleared before being re-built: this eliminates symbols that were used in previous versions of klibs but are no longer used
- duplicated entries in the symbol list are removed only once, instead of after building each klib binary
- similarly, the backup file is deleted only once, when the final symbol list is ready

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -114,11 +114,31 @@ SRCS-mbedtls-tls= \
 	$(MBEDTLS_DIR)/library/ssl_ticket.c \
 	$(MBEDTLS_DIR)/library/ssl_tls.c \
 
-all: $(PROGRAMS)
+all: klib-syms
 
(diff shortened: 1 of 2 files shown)


sanderssj (commented):
LGTM

wjhun (approved):

PR #1637, opened 2021-12-09
Sockets: add management of receive buffer size

francescolavra:
This PR changes the socket implementation (for network sockets, Unix domain sockets, and netlink sockets) so that the total length (in bytes) of the input data buffered in a given socket does not exceed the value of the so_rcvbuf parameter (that is configurable via a "so_rcvbuf" attribute in the manifest). The existing limit on the maximum number of packets given by the size of the input queues is kept in place.
In addition, any UDP packets that are discarded due the length of a socket receive buffer being exceeded are now properly deallocated.
getsockopt() has been amended so that it returns the correct value for the SO_SNDBUF and SO_RCVBUF socket options.
The network socket runtime test has been amended to exercise these new features.

The first commit fixes a bug in which a netlink message can be dequeued (and the memory allocated for it can be leaked) without having been read by userspace.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -113,6 +113,10 @@ typedef struct netsock {
     } info;
 } *netsock;
 
+#define DEFAULT_SO_RCVBUF   0x34000 /* same as Linux */
+
+int so_rcvbuf;
+
 static sysreturn netsock_bind(struct sock *sock, struct sockaddr *addr,
         socklen_t addrlen);
 static sysreturn netsock_listen(struct sock *sock, int backlog);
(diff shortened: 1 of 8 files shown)


wjhun (approved):
Looks good. Would it be useful to implement the setting of SO_RCVBUF via setsockopt(2) now?

francescolavra:
> Would it be useful to implement the setting of SO_RCVBUF via setsockopt(2) now?

Yes, I think it would make a useful future addition. There would need to be a new member in struct sock where the per-socket receive buffer size is stored. Another thing that should be done is making getsockopt and setsockopt work with all types of sockets, not just network sockets.

PR #1638, opened 2021-12-14
Global/static variables: prevent unwanted modifications

francescolavra:
This change set aims at reducing the kernel attack surface by minimizing the amount of global and static variables whose memory is mapped as read-write in the MMU. To this end, the `const` modifier is being added to constant variables (this instructs the linker to put those variables in the rodata section, which can be mapped as read-only when mapping the kernel ELF file), and variables that are only modified during boot are placed in two new linker sections that are initially mapped as read-write but re-mapped as read-only before starting the user program.
The first 3 commits remove some global/static variables that are not needed.

### DIFF
--- a/rules.mk
+++ b/rules.mk
@@ -110,7 +110,7 @@ KERNCFLAGS+=    -mno-sse \
 		-mno-sse2
 endif
 KERNCFLAGS+=	-fno-omit-frame-pointer
-KERNLDFLAGS=	--gc-sections -n -L $(OUTDIR)/klib
+KERNLDFLAGS=	--gc-sections -z max-page-size=4096 -L $(OUTDIR)/klib
 
 ifeq ($(MEMDEBUG),mcache)
 CFLAGS+= -DMEMDEBUG_MCACHE
--- a/src/kernel/pvclock.c
+++ b/src/kernel/pvclock.c
@@ -2,8 +2,8 @@
 #include <pvclock.h>
 #include <apic.h>
 
-static heap pvclock_heap;
-static volatile struct pvclock_vcpu_time_info *vclock = 0;
+BSS_RO_AFTER_INIT static heap pvclock_heap;
+BSS_RO_AFTER_INIT static volatile struct pvclock_vcpu_time_info *vclock;
 
 u64 pvclock_now_ns(void)
 {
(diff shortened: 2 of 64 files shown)


rules.mk
```diff
@@ -110,7 +110,7 @@ KERNCFLAGS+=    -mno-sse \
 		-mno-sse2
 endif
 KERNCFLAGS+=	-fno-omit-frame-pointer
-KERNLDFLAGS=	--gc-sections -n -L $(OUTDIR)/klib
+KERNLDFLAGS=	--gc-sections -z max-page-size=4096 -L $(OUTDIR)/klib
```
wjhun, line 113: Just curious what this is for - were you seeing issues with larger page sizes being used for aarch64?

src/kernel/pvclock.c
```diff
@@ -2,8 +2,8 @@
 #include <pvclock.h>
 #include <apic.h>
 
-static heap pvclock_heap;
-static volatile struct pvclock_vcpu_time_info *vclock = 0;
+BSS_RO_AFTER_INIT static heap pvclock_heap;
+BSS_RO_AFTER_INIT static volatile struct pvclock_vcpu_time_info *vclock = 0;
```
wjhun, line 6: can remove the initializer

src/x86_64/flush.c
```diff
@@ -6,16 +6,15 @@
 #define COMP_QUEUE_SIZE (MAX_FLUSH_ENTRIES*2)
 #define ENTRIES_SERVICE_THRESHOLD (MAX_FLUSH_ENTRIES/2)
 
-static boolean initialized = false;
-static int flush_ipi;
-static heap flush_heap;
+BSS_RO_AFTER_INIT static boolean initialized = false;
```
wjhun, line 9: unnecessary initializer

wjhun (approved):
Looks good

rules.mk
francescolavra: The `-n` linker flag turns off page alignment of ELF sections, and this was causing issues because the boot code that maps the kernel ELF binary requires the file offset of all sections to have the same page offset as the base virtual address. But without the `-n` flag, the linker (on x86_64) was aligning ELF sections to 2MB boundaries (which was increasing the kernel binary file size considerably), so I added the `-z max-page-size=4096` flag.

src/kernel/pvclock.c
francescolavra: Right, I now removed the initializer for this variable, as well for the `have_rdseed`, `have_rdrand` and `initialized` variables in other files.

PR #1639, opened 2021-12-14
update ftrace to use latest timer interface

wjhun:
This fixes a regression introduced with the timer changes.


sanderssj (commented):
LGTM

PR #1640, opened 2021-12-14
restore gdbstub operation; add validate_virtual_writable()

wjhun:
This change deletes the incorrect attempt to acquire the lwIP lock in gdb_send
and removes unnecessary setting of the current thread during memory transfers,
instead relying on validate_virtual() (and the newly-added
validate_virtual_writable()) to insure that memory accesses won't fault. Some
extraneous console output during memory access failures has been removed as
well.

### DIFF
--- a/src/kernel/page.c
+++ b/src/kernel/page.c
@@ -194,6 +194,21 @@ boolean validate_virtual(void * base, u64 length)
     return traverse_ptes(u64_from_pointer(base), length, stack_closure(validate_entry));
 }
 
+closure_function(0, 3, boolean, validate_entry_writable,
+                 int, level, u64, vaddr, pteptr, entry)
+{
+    pte pte = pte_from_pteptr(entry);
+    if (!pte_is_present(pte))
+        return false;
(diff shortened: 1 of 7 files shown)


src/kernel/page.c
```diff
@@ -194,6 +194,19 @@ boolean validate_virtual(void * base, u64 length)
     return traverse_ptes(u64_from_pointer(base), length, stack_closure(validate_entry));
 }
 
+closure_function(0, 3, boolean, validate_entry_writable,
+                 int, level, u64, vaddr, pteptr, entry)
+{
+    pageflags pf = pageflags_from_pteptr(entry);
+    return pageflags_is_present(pf) && pageflags_is_writable(pf);
```
sanderssj, line 201: Why not just use pte_is_present here? Also, is it true on every arch that every level must be writable, or just the leaf? I think that's the case, but not totally sure.

sanderssj (commented):
LGTM, just some questions on the new validate_virtual_writable

src/kernel/page.c
wjhun: You're right, this should only be testing the leaves for writability. I rewrote the expression accordingly.

PR #1642, opened 2021-12-18
PC physical ID heap: do not use low memory area

francescolavra:
The low memory address range from 0 to 1MB is used for things like starting secondary CPUs, so it should not be used for the bootstrap heap and should not be available for general memory allocation, otherwise when a secondary CPU is being started the trampoline code that is copied to low memory could overwrite data used for other purposes.


wjhun (approved):
LGTM

PR #1643, opened 2021-12-20
init_fdesc/release_fdesc: change from inline to regular functions

francescolavra:
The definition of the fdesc_io_complete closure in the unix_internal.h header file causes the closure function and the symbols it references to be included in the VDSO shared library, which bloats the VDSO with unneeded code and relocations.
This has a noticeable performance impact and is causing intermittent failures in the Jenkins job that measures web server performance.
This change moves the fdesc_io_complete and fdesc_init function definitions to unix.c, so that the VDSO library does not have unneeded code; for uniformity, the fdesc_release function is also being moved to unix.c.

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -61,6 +61,37 @@ void deallocate_fd(process p, int fd)
     process_unlock(p);
 }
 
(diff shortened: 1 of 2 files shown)


wjhun (approved):
LGTM

PR #1644, opened 2021-12-22
x86_64 cpu: add support for SMEP and UMIP

francescolavra:
Supervisor Memory Execute Protection (SMEP) is a security feature that generates a page fault exception if the kernel tries to execute an instruction from a memory address accessible by the user program. If this exception occurs, the kernel dumps the current execution context to the console and halts.
User-Mode Instruction Prevention (UMIP) is a security feature that generates a general protection exception if the user program tries to execute one of the the SGDT, SIDT, SLDT, SMSW, and STR instructions. If this exception occurs, the kernel sends a SEGV signal to the program. Tested on Google Cloud N2 instance with Intel Ice Lake CPU.

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -223,6 +223,11 @@ define_closure_function(1, 1, context, default_fault_handler,
             goto bug;
         }
 
+        if (is_instruction_fault(frame) && !user) {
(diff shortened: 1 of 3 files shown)


wjhun (approved):
LGTM

PR #1645, opened 2021-12-22
mutex implementation

wjhun:
Building off of the context switching work, this change implements a two-phase
mutex lock which will first spin for some number of iterations (specified on mutex
allocation) before finally suspending the running context if the mutex could
not be acquired. Suspended contexts are queued and processed in a FIFO
fashion, with the unlock function dequeueing and directly scheduling the next
context waiting on the mutex, if any. This is useful for preventing cores from being
excessively tied up with spinning while there is resource contention. The spin
factor argument given on instantiation may be adjusted according to the needs
of the workload; a value of 0 indicates no spinning is desired - in which case
a context will suspend whenever the mutex cannot be immediately acquired - but
may also be made large to prioritize spinning, leaving suspension as a
fallback option that frees up cpu cycles for other contexts to run.

The new context_suspend() function suspends a context by saving the running
processor state to its frame and then adjusting it to return to the caller on
frame return.

As a first client, the lwIP lock has been converted from a spinlock to a
mutex. It is instantiated with a large spin factor in order to maintain
existing performance levels observed with our webg test. As with spinlocks,
the factor of bus contention when considering performance increases with the
number of cores contending for a mutex. This could potentially be ameliorated
in the future by using something like the MCS lock in place of the
compare-and-swap loop for the first phase. Using MCS would also insure that
waiters are processed in a FIFO order for both phases of lock acquisition, not
just the second one.

### DIFF
--- a/src/kernel/mutex.c
+++ b/src/kernel/mutex.c
@@ -1,0 +1,148 @@
+#include <kernel.h>
+
+//#define MUTEX_DEBUG
+#ifdef MUTEX_DEBUG
+#define mutex_debug(x, ...) do {log_printf(" MTX", "%s: " x, __func__, ##__VA_ARGS__);} while(0)
+#else
+#define mutex_debug(x, ...)
+#endif
+
+#ifdef KERNEL
+#define mutex_pause kern_pause
+#else
+#define mutex_pause()
+#endif
+
+/* This implements a two-phase mutex which optionally allows a degree for
+   spinning when aquiring the lock. A limitation of this is that the spinning
+   is only for an uncondended lock, and the waiter does not enter the queue of
+   waiters until the second phase (context suspend). As such, waiters for a
+   mutex are not strictly processed in FIFO order.
+
+   Linux uses an MCS lock for the first phase, but the unqueueing (removal)
+   operation is complex (yet necessary in order to migrate waiters into the
+   queue of suspended tasks). It may be worth implementing something similar
+   here, perhaps moreso for the lessened bus contention than the FIFO
+   property. */
+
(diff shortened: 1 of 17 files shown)


src/kernel/mutex.c
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /* race covered by dequeue loop in unlock */
+    assert(!frame_is_full(ctx->frame));
+
+    mutex_debug("ctx %p about to wait, count %d\n", ctx, m->count);
+    ctx->waiting_on = m;
+    while (!enqueue(m->waiters, ctx)) {
+        mutex_pause();      /* XXX timeout */
+    }
```
francescolavra, line 80: In a condition where the waiter queue is full, this would present the same problem as a spinlock, i.e. the mutex-acquiring context, by spinning indefinitely, could prevent a suspended mutex-holding context from being resumed and releasing the mutex.
I wonder if we should add a struct list to struct context and replace the queue here with a linked list (and maybe replace the `dependents` vector in struct pending_fault with a list as well), or perhaps use a resizable vector instead of a queue. But I realize we would need to protect the list (or the vector) with an additional lock internal to the mutex (similar to the faulting_lock used to handle page faults).

wjhun: Ok, I think I have a solution. In the latest commit, the waiters queue is replaced by a lock-free singly-linked list (based on the one described in the MCS lock papers) composed of waiting contexts.

For another PR, I'll look at adding generic support for such a list - e.g. to use for pending_fault dependents. There are probably a number of places in the kernel that could benefit from such a lock-free list.

PR #1646, opened 2021-12-23
Aarch64: report atomic hwcap in program aux vector

sanderssj:
The dynamic linker can use the cpu's capabilities as reported in the aux vector
pushed onto the program's initial stack to generate additional subdirectories
to search when looking for dynamic libraries. This allows glibc to load hardware
optimized versions of a library when available. On the aarch64 platform, glibc
checks for atomic instructions on the cpu to prefer those versions, located in
/usr/lib/aarch64-linux-gnu/atomics. If the filesystem only contains the atomics
version but atomics are not reported in hwcaps, the dynamic library loading
will fail as the atomics subdirectory will not be checked.

In order to focus on correctly loading dynamic libraries on aarch64, this PR only
implements reporting hwcaps on the aux vector only for aarch64, and
only for the atomic capability, which is the only one required to generate
the desired set of dynamic library search paths. Broader cpu feature checking
is possible now, but without hardware acceleration those features are likely
emulated and there's currently not an obvious benefit to reporting them at this time.


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1647, opened 2021-12-23
syscalls: truncate: fix call to filesystem_put_node()

francescolavra:
filesystem_put_node() should be called only if the previous call to filesystem_get_node() was successful, otherwise the filesystem reference count is wrongly decremented (and the filesystem lock is wrongly released if held by another thread).


wjhun (approved):
LGTM

PR #1648, opened 2021-12-28
Microsoft Azure fixes

francescolavra:
The first commit adds a schedule_timer_service() call to the VMbus interrupt handler, which fixes an issue causing sporadic failures in IP address assignment on Azure instances.
The second commit modifies the net/direct.c code to ensure that the data structures for handling a network connection are not accessed after the connection has been closed; this fixes a crash occurring on Azure instances when the cloud_init klib reports the instance status to the wire server.

### DIFF
--- a/klib/cloud_init.c
+++ b/klib/cloud_init.c
@@ -200,7 +200,7 @@ closure_function(5, 1, void, cloud_download_save,
     apply(sh, s);
 }
 
-closure_function(7, 1, status, cloud_download_recv,
+closure_function(7, 1, boolean, cloud_download_recv,
                  cloud_download_cfg, cfg, buffer_handler, out, value_handler, vh, bytes, content_len, bytes, received, status_handler, sh, merge, m,
                  buffer, data)
 {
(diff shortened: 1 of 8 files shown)


klib/cloud_init.c
```diff
@@ -269,21 +271,21 @@ closure_function(7, 1, status, cloud_download_recv,
...
 
-closure_function(1, 1, buffer_handler, cloud_download_ch,
+closure_function(1, 1, input_buffer_handler, cloud_download_ch,
                  cloud_download_cfg, cfg,
                  buffer_handler, out)
 {
     cloud_download_cfg cfg = bound(cfg);
     status_handler sh = (status_handler)&cfg->complete;
-    buffer_handler in = 0;
+    input_buffer_handler in = 0;
```
sanderssj, line 288: Should this actually be initialized to INVALID_ADDRESS as in cloud_azure.c? If there is a connection established but some other condition (in this particular case I think it would be transmit failure on output buffer handler) causes it to skip initialization of 'in', which would return 0, which is not checked in direct_receive_service, and perhaps could lead to trying to apply a null pointer if there were something in the receive queue. Maybe something about this particular case could prevent that from happening, but seems like INVALID_ADDRESS is the safer initialized value? Also, same question in a one other place.

klib/radar.c
```diff
@@ -164,15 +165,15 @@ closure_function(2, 1, status, telemetry_recv,
...
 }
 
-closure_function(3, 1, buffer_handler, telemetry_ch,
+closure_function(3, 1, input_buffer_handler, telemetry_ch,
                  const char *, url, buffer, data, value_handler, vh,
                  buffer_handler, out)
 {
     buffer data = bound(data);
-    buffer_handler in = 0;
+    input_buffer_handler in = 0;
```
sanderssj, line 176: Same question as above.

sanderssj (commented):
LGTM aside from the one question

klib/cloud_init.c
francescolavra: Yes, I agree, INVALID_ADDRESS is a safer initialization value. I fixed that, both here and in klib/radar.c

wjhun (approved):
LGTM

PR #1649, opened 2022-01-01
Unix pipe: only report EPOLLIN event when data is available to read

francescolavra:
The Unix pipe implementation is incorrectly reporting both EPOLLIN and EPOLLHUP events on a pipe reader file descriptor after closing the writer file descriptor. This is causing `Null check operator used on a null value` unhandled exceptions when running a Dart program without AOT compilation.
This change fixes the above issue by only reporting EPOLLIN if data is available in the pipe buffer. A test case is being added to the pipe runtime test to check that the correct poll events are reported after closing the write side of a pipe.


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1650, opened 2022-01-03
Network socket: modify poll events to match Linux implementation

francescolavra:
In Linux, a TCP socket that has never been connected reports (POLLHUP | POLLOUT) events when polled, and a TCP socket whose connection with the remote peer has been closed reports (POLLIN | POLLOUT).
This commit modifies the events reported in socket_events() to match the Linux implementation. This fixes an issue that makes the redis server (version 5.0.5) use 100% CPU when subject to a workload from YCSB; more specifically, the redis server relies on TCP sockets whose connection with the remote peer has been closed to report the EPOLLIN event.


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1652, opened 2022-01-14
Add getsockopt support for reuseport/addr and tcp_nodelay

sanderssj:
This adds support in getsockopt for reading the values of SO_REUSEADDR,
SO_REUSEPORT, and TCP_NODELAY. SO_REUSEPORT is not currently
supported so it always returns 0 and setting it is a no-op. This also fixes
setting TCP_NODELAY in setsockopt which was setting the opposite value
requested.


wjhun (approved):
LGTM

PR #1653, opened 2022-01-15
restore correct detection of runtime test errors by defaulting to qemu debug exit

wjhun:
With the introduction of acpica and the default installation of the
acpi_powerdown vm_halt method on the PC platform, runtime CI tests have not
been passing the test exit codes back to the shell. This results in "make
runtime-tests" passing even if tests have failed with a non-zero exit code.

This adds a "debug_exit" manifest flag to indicate that vm_exit() should
return to qemu via the QEMU_HALT() function regardless of whether the platform
has installed a vm_halt method. An option has been added to mkfs to allow the
amending of manifest root tuples from the command line. Runtime tests are now
built with "debug_exit:t" added to the root tuples by default.

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -276,7 +276,7 @@ ifeq ($(IMAGE),)
 endif
 	@ echo "MKFS	$@"
 	@ $(MKDIR) $(dir $(IMAGE))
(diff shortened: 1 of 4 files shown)


francescolavra (commented):
Looks good to me

PR #1654, opened 2022-01-17
Allocate vga buffer address from kernel va space

sanderssj:
The vga buffer physical was being mapped to the identity virtual address
which is in the user virtual address area. This change allocates a new
virtual address from kernel virtual addresses to map to the vga buffer.


wjhun (approved):
LGTM

PR #1657, opened 2022-01-21
Add print_tree command to gdb script to print rbtree

sanderssj:
This change adds a new python gdb command, print_tree, to print the pointers of
rbnodes of an rbtree in order. If a second argument, containing type, is passed
the script will print the pointers to the containing struct instead of the
rbnode. Also added are helper macros offsetof() and container_of() which can
be used in gdb expressions.

Also added is a new array called 'processes' containing the pointers to the two
process structs in order to facilitate easier process debugging.

Example usage:
```
(gdb) print_tree processes[1]->threads thread
(struct thread *)0xffffc00000e03000
(struct thread *)0xffffc00000e04000
(struct thread *)0xffffc00000e05000
(struct thread *)0xffffc00000e06000
(gdb) 
```

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -13,6 +13,9 @@
 #define pf_debug(x, ...) thread_log(current, x, ##__VA_ARGS__);
 #endif
 
+#define MAX_PROCESSES 2
+process processes[MAX_PROCESSES];
+
 BSS_RO_AFTER_INIT static unix_heaps u_heap;
 
 unix_heaps get_unix_heaps()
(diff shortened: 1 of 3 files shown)


src/unix/unix.c
```diff
@@ -12,6 +12,9 @@
 #define pf_debug(x, ...) thread_log(current, x, ##__VA_ARGS__);
 #endif
 
+#define MAX_PROCESSES 2
+static process processes[MAX_PROCESSES];
```
wjhun, line 16: I needed to remove the static here to run the example.

wjhun (approved):
One note, otherwise looks great!

src/unix/unix.c
sanderssj: Interesting--I wonder if gdb version or something affects this. I removed the static though so it should work now.

PR #1658, opened 2022-01-24
Fix tfs-fuse resolve_cstring calls and add partition select flag

sanderssj:
Tfs-fuse was incorrectly passing nulls to resolve_cstring and resolve_cstring_follow causing 
crashes. Those have been replaced with the rootfs global pointer.

This change also adds a new flag '-b' which mounts the boot partition instead of the 
default root partition.


wjhun (approved):
LGTM

Issue #1661, opened 2022-01-27
Nightly issue, mutex?


jasonrichardsmith:
I have to use nightly right now because I need this code.
https://github.com/nanovms/nanos/commit/c3a488e4be78e8697fc422f05f6cb934d2192bf9

VMS I provisioned the other day are fine but they are now failing on nightly. I am getting errors below.

```
ops pkg load python_3.8.6 -c config.json -n
 100% |████████████████████████████████████████|  [0s:0s]
warning: overwriting existing file /etc/resolv.conf hostpath old: /tmp/python3_temp328684329/resolv.conf new: etc/resolv.conf
booting /home/jason/.ops/images/python3 ...
en1: assigned 10.0.2.15
/.local/lib/python3.8/site-packages/googlecloudprofiler/client.py:167: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if len(self._profilers) is 0:
{"timestamp": 1643296662.4657683, "level": null, "name": "emailservice-server", "message": "starting the email service in dummy mode.", "severity": "INFO"}
{"timestamp": 1643296662.4674828, "level": null, "name": "emailservice-server", "message": "Profiler disabled.", "severity": "INFO"}
{"timestamp": 1643296662.4689507, "level": null, "name": "emailservice-server", "message": "Tracing enabled.", "severity": "INFO"}
en1: assigned FE80::BC72:AEFF:FEC3:7D6
{"timestamp": 1643296671.4780874, "level": null, "name": "emailservice-server", "message": "Tracing disabled.", "severity": "INFO"}
E0127 15:17:51.493388929       1 cpu_linux.cc:50]            Cannot determine number of CPUs: assuming 1
epoll_ctl error: add: EPOLLEXCLUSIVE not supported
epoll_ctl error: add: EPOLLEXCLUSIVE not supported
D0127 15:17:51.496470846       1 is_epollexclusive_available.cc:100] epoll_ctl with EPOLLEXCLUSIVE failed with error: 22. Not using epollex polling engine.
I0127 15:17:51.498017771       1 ev_epollex_linux.cc:1624]   Skipping epollex because it is not supported.
I0127 15:17:51.499028795       1 ev_epoll1_linux.cc:122]     grpc epoll fd: 3
D0127 15:17:51.499764070       1 ev_posix.cc:173]            Using polling engine: epoll1
D0127 15:17:51.502260830       1 lb_policy_registry.cc:42]   registering LB policy factory for "grpclb"
[...]

PR #1662, opened 2022-01-27
fix uninitialized variable in init_context(), refactor

wjhun:
This fixes missing initialization of the context's next_waiter field and moves
init_context() to kernel.h. Per-architecture context initialization now occurs
in init_context_machine().

#1661


sanderssj (commented):
LGTM

PR #1663, opened 2022-01-28
Add new mem_debug compile options and nohdr flag for mem_debug_backed; smp for e2e tests

sanderssj:
This change adds new convenience compile options to add a mem_debug wrapper to the linear
backed heap with MEMDEBUG=backed or MEMDEBUG=all. However, the linear backed heap uses
2MB pages, which would require 4MB additional memory for padding on each allocation. For
a non-trivial program this consumes too much memory, so mem_debug_backed has been
modified with a nohdr flag that does not allocate padding, forgoing allocation metadata and
overrun checks, leaving only the initialized and freed patterns checks since they require no
additional memory. The CI script has been modified to use MEMDEBUG=all.

The e2e.go program has also been modified to pass the number of host cpus to ops in
order to have smp testing for the e2e tests.

### DIFF
--- a/src/runtime/heap/mem_debug.c
+++ b/src/runtime/heap/mem_debug.c
@@ -172,6 +172,25 @@ heap mem_debug_objcache(heap meta, heap parent, u64 objsize, u64 pagesize)
 }
 
 #ifdef KERNEL
+static void *mem_debug_backed_alloc_map_nohdr(backed_heap h, bytes b, u64 *phys)
+{
+    mem_debug_backed_heap mdh = (mem_debug_backed_heap)h;
+    void *v = alloc_map(mdh->parent, b, phys);
+#ifdef MEMDBG_INIT
+    set_pattern(v, b, &pat_init, sizeof(pat_init));
+#endif
+    return v;
+}
+
+static void mem_debug_backed_dealloc_unmap_nohdr(backed_heap h, void *v, u64 p, bytes b)
+{
+    mem_debug_backed_heap mdh = (mem_debug_backed_heap)h;
+#ifdef MEMDBG_FREE
+    set_pattern(v, b, &pat_freed, sizeof(pat_freed));
+#endif
+    dealloc_unmap(mdh->parent, v, p, b);
+}
+
 static void *mem_debug_backed_alloc_map(backed_heap h, bytes b, u64 *phys)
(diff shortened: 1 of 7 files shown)


src/runtime/heap/mem_debug.c
```diff
@@ -203,23 +222,23 @@ static void mem_debug_backed_dealloc_unmap(backed_heap h, void *v, u64 p, bytes
 
 static u64 mem_debug_backed_alloc(heap h, bytes b)
 {
-    return u64_from_pointer(mem_debug_backed_alloc_map((backed_heap)h, b, 0));
+    return u64_from_pointer(alloc_map((backed_heap)h, b, 0));
```
francescolavra, line 225: We should be able to use mem_debug_backed_alloc_map_nohdr() here, which should work regardless of whether the nohdr option has been specified on heap creation.

src/runtime/heap/mem_debug.c
```diff
@@ -203,23 +222,23 @@ static void mem_debug_backed_dealloc_unmap(backed_heap h, void *v, u64 p, bytes
...
 }
 
 static void mem_debug_backed_dealloc(heap h, u64 a, bytes b)
 {
-    mem_debug_backed_dealloc_unmap((backed_heap)h, pointer_from_u64(a), 0, b);
+    dealloc_unmap((backed_heap)h, pointer_from_u64(a), 0, b);
```
francescolavra, line 230: Same here, we should be able to use mem_debug_backed_dealloc_unmap_nohdr()

src/runtime/heap/mem_debug.c
sanderssj: I'm not following you I think--the alloc_map/dealloc_unmap function used should depend on the nohdr flag at creation, as a normal alloc/dealloc should still use the header checks if that's how it was set up. Always using the nohdr versions would skip the boundary and header checks.

francescolavra: You are right, somehow I was under the impression that invoking alloc_map() here would call the non-debug version (thereby skipping all checks) of the backed heap allocation function. But now I see that's clearly not the case, since we still operate on a memdebug heap.

wjhun (approved):
LGTM

PR #1664, opened 2022-01-28
TFS: run flush completions asynchronously in kernel code

francescolavra:
run_flush_completions() in tfs/tlog.c is called with the TFS lock held. In the kernel code, one of the flush completions can trigger the following sequence of function calls: pagecache_sync_volume -> pagecache_scan -> pagecache_commit_dirty_pages -> filesystem_storage_write; the last function in this sequence tries to acquire the TFS lock, which results in a deadlock.
This change solves the above issue by invoking the TFS log flush completions asynchronously.

### DIFF
--- a/src/tfs/tlog.c
+++ b/src/tfs/tlog.c
@@ -477,7 +477,11 @@ static void run_flush_completions(log tl, status s)
     if (tl->flush_completions) {
         status_handler sh;
         vector_foreach(tl->flush_completions, sh)
+#ifdef KERNEL
+            async_apply_status_handler(sh, s);
+#else
             apply(sh, s);
+#endif
         vector_clear(tl->flush_completions);
     }
 }


src/tfs/tlog.c
```diff
@@ -477,7 +477,11 @@ static void run_flush_completions(log tl, status s)
     if (tl->flush_completions) {
         status_handler sh;
         vector_foreach(tl->flush_completions, sh)
+#ifdef KERNEL
+            async_apply_status_handler(sh, s);
```
sanderssj, line 481: It looks like log_extend_link would sometimes be run async this way, which modifies the old extension's staging buffer outside of any lock in that situation, which maybe is okay? It's implied that this closure would be the only reference so it might be fine.

My other concern is guaranteeing this completion is run in a shutdown situation where this might be left on an async queue and thus missing an fs metadata write if the filesystem sync completes first.

francescolavra: > It looks like log_extend_link would sometimes be run async this way, which modifies the old extension's staging buffer outside of any lock in that situation, which maybe is okay? It's implied that this closure would be the only reference so it might be fine.

Yes, I think it's ok for log_extend_link() to be called without the TFS lock, because it's the last action being made on a log extension before closing it, so no other parts of the code should access that extension.

> My other concern is guaranteeing this completion is run in a shutdown situation where this might be left on an async queue and thus missing an fs metadata write if the filesystem sync completes first.

During shutdown, filesystem_flush() is called, which first flushes the TFS log (thus writing any pending filesystem metadata), and then on flush completion syncs the page cache volume, which takes care of writing any pending file data. Only after this last action is completed, vm_exit() is called. So it seems we cannot miss a metadata write.

wjhun (approved):
LGTM

src/tfs/tlog.c
sanderssj: Yeah, I was thinking of the case where a rebuild happened at the same time as shutdown, which could cause the log_extend_link to run off the async queue, which would generate new i/o and whether or not the shutdown sync could race by that. It looks like the sync would be dependent on the log_extend_link to finish (which leads to the rebuild finishing), and then the sync would run.  Thanks!

PR #1669, opened 2022-01-29
fix deadlock scenario in pipe reads and writes

wjhun:
A deadlock may occur in the following scenario: pipe_read_bh or pipe_write_bh
invoke notify_dispatch, with the pipe lock held, to indicate a cleared
condition (EPOLLIN or EPOLLOUT, respectively). A coincident epoll_wait takes
the efd lock and then invokes pipe_read_events, which tries to acquire the
pipe lock. The notify_dispatch invokes wait_notify, which in turn is blocked
waiting for the efd lock. This patch eliminates the deadlock by moving these
notify_dispatch invocations to follow release of the pipe lock.

This deadlock would materialize when testing webg on qemu with large number (n
= 16) of cpus and 'ab -n 1000000 -c 100', which would hang after a few passes.

### DIFF
--- a/src/unix/pipe.c
+++ b/src/unix/pipe.c
@@ -143,8 +143,8 @@ closure_function(5, 1, sysreturn, pipe_read_bh,
         goto out;
     }
 
-    buffer b = pf->pipe->data;
     pipe_lock(pf->pipe);
+    buffer b = pf->pipe->data;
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

PR #1672, opened 2022-02-01
uname(2): derive nodename from either /etc/hostname or IP address

francescolavra:
With this change, the string placed in the `nodename` field of the struct passed to the uname syscall is derived from the contents of the /etc/hostname file, if present in the root filesystem, or from the IP address of the default network interface.
Using the IP address allows different instances created from the same image to have different host names, similarly to what happens with Linux instances in AWS or GCP.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -1,5 +1,6 @@
 #include <unix_internal.h>
 #include <filesystem.h>
+#include <lwip.h>
 #include <storage.h>
 
 // lifted from linux UAPI
@@ -18,6 +19,8 @@ typedef struct syscall_stat {
     u64 errors;
     u64 usecs;
 } *syscall_stat;
+
+static buffer hostname;
 
 static struct syscall_stat stats[SYS_MAX];
 BSS_RO_AFTER_INIT boolean do_syscall_stats;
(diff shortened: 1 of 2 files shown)


src/unix/syscall.c
```diff
@@ -2605,6 +2637,16 @@ void init_syscalls(tuple root)
     syscall = syscall_handler;
     syscall_io_complete = closure(h, syscall_io_complete_cfn);
     io_completion_ignore = closure(h, io_complete_ignore);
+    vector hostname_v = split(h, alloca_wrap_cstring("etc/hostname"), '/');
+    tuple hostname_t = resolve_path(root, hostname_v);
+    buffer b;
```
wjhun, line 2642: This vector-of-buffers dealloc seems generally useful - maybe it could accompany split(), join(), etc. in vector.h?

src/unix/syscall.c
```diff
@@ -2506,6 +2525,19 @@ closure_function(0, 2, void, syscall_io_complete_cfn,
...
 }
 
+closure_function(0, 1, status, hostname_done,
+                 buffer, b)
+{
+    hostname = b;
+
+    /* Remove trailing newline, if any. */
+    char last_char = *(char *)(buffer_end(hostname) - 1);
+    if (last_char == '\n')
```
wjhun, line 2535: This could also check for a carriage return in case the file was written with CR-LF (e.g. Windows).

wjhun (approved):
LGTM

src/unix/syscall.c
francescolavra: OK, I wrote a `split_dealloc()` function in vector.h that deallocates what is allocated by split().

src/unix/syscall.c
francescolavra: OK, I modified this part so that any trailing CR and LF characters are removed.

PR #1673, opened 2022-02-01
fs_full test: retry errored writes to ensure disk is full

sanderssj:
The fs_full test tries to fill up the disk in several different circumstances,
but sometimes a small amount of additional space may become available
after the first write error indicating the disk is full. This has an effect on
write checks which may then not report the correct amount that could
be written and failing the test. The write function has been modified to
retry failed writes twice after a brief delay to ensure all sectors that can
be written to the disk have been written.


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1674, opened 2022-02-02
Klibs: syslog improvements

francescolavra:
This PR contains miscellaneous improvements to the syslog klib:
- in case of failure to resolve the syslog server name, the backoff delay for retrying DNS resolution is limited to 1 minute, to prevent long delays before logs start to be shipped after a previously unresolvable server becomes resolvable
- the dns_gethostbyname() function is now called before sending syslog packets, even after the server name has been
successfully resolved, so that when the cached DNS entry expires a new DNS request will be sent and any changes to the server IP address will be observed
- if the IMAGE_NAME environment variable is present, it is used to populate the APP_NAME field in syslog messages, while
the program name is used as a fallback

### DIFF
--- a/klib/syslog.c
+++ b/klib/syslog.c
@@ -439,7 +439,10 @@ int init(status_handler complete)
     }
     if (syslog.server) {
         syslog_server_resolve();
-        syslog.program = get(root, sym(program));
+        tuple env = get_environment();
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1677, opened 2022-02-07
use an MCS lock for mutex spinwaiters

wjhun:
This updates the two-phase mutex to use a cancelable MCS (Mellor-Crummey and
Scott) lock, which allows only one spin waiter to contend for the mutex at a
time. As before, a mutex lock operation will spin for only m->spin_iterations
before reverting to suspending the context. Before suspending, the cpuinfo is
removed from the MCS list, and the corresponding context is added to the
mutex's queue of waiters. (The waiters list is now a standard doubly-linked
list, protected by a spinlock within the mutex.) If the MCS lock was acquired
while spinning, the mutex itself is then taken and the MCS lock is released,
passing ownership to the next queued spinner, if there is one.

On an unlock, the mutex is released by zeroing m->turn. This may allow the
current holder of the MCS lock, which is spinning while trying to acquire
m->turn, to proceed with taking the mutex. If there is a (suspended) context
waiting on the mutex, it is treated with equal priority to the MCS owner and
is scheduled to be resumed. The resumed context then spins to acquire the
mutex.

While MCS is relatively straightforward to implement, the ability to exit the
MCS queue - critical to prevent cores from spinning indefinitely - is a
somewhat delicate process that has been adapted from the optimistic spin queue
implementation in Linux by Peter Zijlstra, et al.

### DIFF
--- a/src/kernel/mutex.c
+++ b/src/kernel/mutex.c
@@ -2,7 +2,7 @@
 
 //#define MUTEX_DEBUG
 #ifdef MUTEX_DEBUG
-#define mutex_debug(x, ...) do {log_printf(" MTX", "%s: " x, __func__, ##__VA_ARGS__);} while(0)
+#define mutex_debug(x, ...) do {log_printf(" MTX", "cpu %.2d %s: " x, ci->id, __func__, ##__VA_ARGS__);} while(0)
 #else
 #define mutex_debug(x, ...)
 #endif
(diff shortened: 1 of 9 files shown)


src/kernel/mutex.c
```diff
@@ -14,70 +14,168 @@
...
         return true;
     }
-    prev_tail->next_waiter = ctx;
+    mutex_debug("   inserting into waiters list and suspending\n");
+    list_insert_before(&m->waiters, &ctx->mutex_l);
+    spin_unlock(&m->waiters_lock);
     context_pre_suspend(ctx);
     context_suspend();
+  acquire:
+    while (!compare_and_swap_64((u64*)&m->turn, 0, u64_from_pointer(ctx)))
```
wjhun, line 175: I've been trying to work out a little timeout here - just counting down from m->spin_iterations - at which point this will suspend and re-schedule the context (via a new context_yield() call). I'm running into a hang which I'm trying to resolve - but I'm curious what you all think: is it even worth doing this? As far as I see it, the context needs to keep running (spinning) as it is the head contender (along with the MCS owner) - so it shouldn't just go back into the waitlist but rather be immediately scheduled. So the only advantage of breaking out of the spin here is just to make a pass through runloop before returning to this context anyway. And there's the possibility that the context could be resumed on another cpu - which supposedly could incur cache hits.

Should I even bother? Another point is that, if we enable interrupts by default during kernel execution (which was attempted last year but blocked due to issues that I think have since been resolved), external interrupts could be handled while spinning and some bh processing could occur as well. That maybe seems like better placed effort.

src/kernel/mutex.c
```diff
@@ -14,70 +14,168 @@
...
 
-    assert(!ctx->next_waiter);
-    if (!wait)
-        return mutex_cas_take(m, ctx);
+    assert(!ctx->waiting_on);
+    if (!wait) {
+        boolean acquired = false;
+        if (compare_and_swap_64((u64*)&m->mcs_tail, u64_from_pointer(ci), 0)) {
+            acquired = compare_and_swap_64((u64*)&m->turn, 0, u64_from_pointer(ctx));
+            mcs_unlock(m, ci);
```
francescolavra, line 97: As far as I can see, at this stage we haven't touched the MCS lock at all, so mcs_tail will never equal `ci`, and the mcs_unlock() call here is unnecessary.

src/kernel/mutex.c
```diff
@@ -14,70 +14,168 @@
...
+    }
+    mutex_debug("   spin timeout; removing from MCS list\n");
+
+    /* spin timed out; extract waiter from list and block
+
+       step A: first undo the setting of prev->mcs_next (unless we are handed
+       the lock in the meantime), thus making prev spin wait in
+       mutex_get_next() for a valid next pointer */
+    while (prev->mcs_next != ci || !compare_and_swap_64((u64*)&prev->mcs_next,
+                                                        u64_from_pointer(ci), 0)) {
```
francescolavra, line 134: Also here the `prev->mcs_next != ci` part seems redundant

src/kernel/mutex.c
francescolavra: I think spinning indefinitely should be avoided here, since if the mutex owner has been suspended it may not be able to resume due to the spinner context taking up all the CPU.
But it seems to me that the MCS owner doesn't necessarily have to release the MCS lock when going to the waiter list (as long as the number of MCS spin iterations cannot be infinite): it could suspend itself while holding the MCS lock, this way there will be no contention when the mutex is released, because the head of the waiter list will be the same as the MCS owner.

src/kernel/mutex.c
```diff
@@ -14,70 +14,168 @@
...
-    assert(!ctx->waiting_on);
-    assert(!m->turn);
-    m->turn = ctx;
+    do {
+        /* If we're removing ci from the list and it is the last item queued,
+           try to rewind tail to point to prev.
+
+           If we're unlocking ci and at the end of the list, then this is just
+           a simple removal. */
+        if (m->mcs_tail == ci &&
```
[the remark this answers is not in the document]
wjhun: I picked this up after seeing it used in the linux osq and qrwlocks. In my understanding, the point of checking the value before attempting a CAS is to help reduce bus traffic. The CAS results in a 'lock cmpxchg' which puts the cacheline containing the address into the exclusive locked state, and the line may also be made dirty again even if the exchange didn't happen. (Actually I think it is always written ... AMD64 PM states "If the compared operands were unequal, CMPXCHG writes the same value to the memory operand that was read.") The optimization of spinning on a read and only attempting the CAS once the desired value is read is supposed to help keep the cache lines in a shared state (and accordingly reduce bus traffic / MESI transitions) while spinning as well as eliminate loss of memory bandwidth from the unnecessary writes.

A quick search gives a few discussions about this:

https://en.wikipedia.org/wiki/Spinlock#Significant_optimizations
https://stackoverflow.com/a/63350048

src/kernel/mutex.c
wjhun: This is a bug; the CAS of m->mcs_tail is supposed to exchange 0 -> ci, not ci -> 0. The idea here is that we don't want a try lock to jump ahead of other MCS spinners and grab the mutex out-of-turn, so it first needs to grab the (uncontended) MCS lock before trying to grab the mutex (and it unlocks MCS either way).

src/kernel/mutex.c
wjhun: I considered holding the MCS lock until the mutex unlock, and in fact originally tried to write it that way. However, for MCS to work properly, the spinning occurs on a per-CPU "local" variable - not on the mutex itself - and the previous and next links are also per-cpu variables. This is why these are stored in the cpuinfo. Even if a spin times out and the context is suspended, these per-cpu variables are still associated with _this_ mutex acquisition and would not be available for use by another.

Also, an explicit mutex handoff at the time of waking up a context means that no spinner could possibly take the lock and get work done between the point at which the context is scheduled to be resumed and the time that it wakes up and assumes ownership of the mutex. Say, for instance, that there is a considerable amount of work already queued on the runqueue, and after a mutex unlock the context return is scheduled behind it (with the mutex now owned by the scheduled context). Between the interval that this context is scheduled and resumes running, other contexts that are spinning cannot take the mutex because of this strict ordering. Removal from MCS basically trades this fair, FIFO property for (hopefully) better utilization of the resource protected by the lock.

There's yet another potential issue - not one that would affect us now, but that could become an issue as our scheduling scheme evolves (e.g. associate priorities and quanta with contexts). If we start to associate priorities with contexts and schedule returns according to priority, we could have an issue where a low-priority mutex waiter times out of its spin and suspends. An unlock wakes up the waiter, but at the same time a higher priority context becomes runnable and attempts to take the same mutex. While it is spinning, there is a priority inversion that will cause the mutex to be held up until the higher-priority context times out and joins the waiter list. This is more time that the mutex is held up without any useful work being done. It seems then that some special provision for priority inheritance would need to be taken into account.

By allowing removal from MCS, we can say that all members of the MCS list are actively-running spinners, and forward progress of ownership through them should be guaranteed provided each mutex holder behaves. I expect that, in the future, the spinning will not only check m->spin_iterations but also check if a context requires rescheduling (like need_resched() in linux - e.g. some task running as a hypothetical kernel thread exhausts its time quantum). If that is ever the case, removal from the MCS list will be essential.

PR #1678, opened 2022-02-16
fault in user pages prior to accessing them under the filesystem lock

wjhun:
A number of system calls pass user buffers - like pathnames and buffers for
returning results (e.g. getdents, stat) - to functions which attempt to access
those buffers under the filesystem lock. This can lead to a deadlock
situation, as a fault taken for a file-backed page could attempt to acquire
the already-held filesystem lock. To remedy this problem, the functions
fault_in_user_memory() and fault_in_user_string() are added here to induce
faults for user buffers prior to taking the lock. These functions may be
applied prior to any section of code in which a page fault cannot be safely
handled.

The function validate_user_memory_permissions() has been added to validate
that vmaps covering a specified range adhere to the required (or disallowed)
permission flags, and also that the range is completely covered by vmaps. A
true return value from this function indicates that page faults over this
range should not result in a permissions violation or missing vmap fault.

Tests for the buffer permissions check, use of a pathname from a file-backed
mapping, and storing of filesystem information to a file-backed mapping have
been added to the mmap runtime test. All of these tests would fail prior to
implementation of the above remedies.


sanderssj (commented):
LGTM, although I think symlink_internal and getcwd access user buffers after fs lock but without the new fault-in checks.

src/unix/syscall.c
```diff
@@ -1000,7 +993,7 @@ sysreturn dup3(int oldfd, int newfd, int flags)
 
 sysreturn mkdir(const char *pathname, int mode)
 {
-    if (!validate_user_string(pathname))
+    if (!fault_in_user_string(pathname))
```
francescolavra, line 996: mkdirat() should have this change as well

sanderssj:
Could we also add a version of validate_user_memory_permissions that takes a process argument instead of assuming current is valid?

wjhun:
> Could we also add a version of validate_user_memory_permissions that takes a process argument instead of assuming current is valid?

Sure, that makes sense. Just updated and pushed.

PR #1684, opened 2022-02-28
Add tracelist manifest symbol to whitelist syscalls for tracing

sanderssj:
This PR adds a new manifest symbol 'tracelist' that takes a list of syscalls or sets of syscalls to print trace
information for. This is the inverse of the 'notrace' operation. Syscalls are now registered with a flag to place
them in one or more sets which can be specified by special names starting with a percent '%'. The set names
and assignment of syscalls into sets follows the strace application. The list of currently supported groups
are: %file, %desc, %network, %process, %signal, %memory.


src/unix/syscall.c
```diff
@@ -2661,36 +2660,77 @@ void init_syscalls(tuple root)
...
+        for (int i = 0; i < sizeof(_linux_syscalls) / sizeof(_linux_syscalls[0]); i++) {
+            struct syscall *s = bound(p)->syscalls + i;
+            if (!s->name)
+                continue;
 
-        buffer name = alloca_wrap_buffer(s->name, runtime_strlen(s->name));
-        if (!buffer_compare(name, v))
-            continue;
+            buffer name = alloca_wrap_buffer(s->name, runtime_strlen(s->name));
+            if (!buffer_compare(name, v))
```
wjhun, line 2725: and here

wjhun (approved):
LGTM

francescolavra (commented):
LGTM

PR #1686, opened 2022-03-07
tracelog

wjhun:
This change introduces a trace logging feature into the Nanos kernel.
Debugging prints that were previously directed to the console may now be
buffered for later retrieval using the tprintf() and vtprintf() functions.
The tracelog was designed to minimize the cost of adding trace entries, with
the ultimate aim of minimizing impact on execution time of the code being
analyzed. Each cpu has a dedicated trace buffer, avoiding the need to take a
lock when adding trace data. These per-cpu buffers are merged together (and
replaced with fresh buffers) by a background task that is scheduled when
buffer storage runs low or a request to download trace data is made (via the
tracelog HTTP interface on port 9091). Trace entries may be filtered based on
a "tag" symbol as well as optional attributes specified in the trace (e.g. a
thread id for unix thread traces).

The tracelog is only built when "TRACE=tracelog" is specified on the make
commandline. Tracing output from various parts of the kernel may be selected
individually using a comma-separated DEBUG variable, with DEBUG=all indicating
that all traces should be enabled. If DEBUG is specified without building
support for the tracelog, trace output is directed to the console.

For more details on tracelog usage, see doc/tracelog.txt


src/kernel/tracelog.c
```diff
@@ -0,0 +1,568 @@
...
+    list_init(&tracelog.collated_entries);
+    list_init(&tracelog.free_tracelog_buffers);
+    tracelog.n_free = 0;
+    tracelog.alloc_size = TRACELOG_DEFAULT_BUFFER_SIZE;
+    tracelog.trace_tags = 0;
+    init_closure(&tracelog.collator, tracelog_collator);
+    tracelog.collator_scheduled = false;
+    tracelog.disabled = false;
+    
+    for (int i = 0; i < present_processors; i++) {
```
francescolavra, line 559: total_processors

src/kernel/tracelog.c
```diff
@@ -0,0 +1,568 @@
...
+        tracelog_buffer b = allocate_tracelog_buffer_locked(); /* don't need lock here */
+        assert(b != INVALID_ADDRESS);
+        cpuinfo ci = cpuinfo_from_id(i);
+        ci->tracelog_buffer = b;
+        vector cv = allocate_vector(h, 8);
```
francescolavra, line 564: What's the purpose of this vector allocation?

platform/pc/Makefile
```diff
@@ -386,7 +393,7 @@ PCI_BUS=	pci.0
 endif
 QEMU_TAP=	-netdev tap,id=n0,ifname=tap0,script=no,downscript=no
 QEMU_NET=	-device $(NETWORK)$(NETWORK_BUS),mac=7e:b8:7e:87:4a:ea,netdev=n0 $(QEMU_TAP)
-QEMU_USERNET=	-device $(NETWORK)$(NETWORK_BUS),netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostfwd=tcp::9090-:9090,hostfwd=udp::5309-:5309
+QEMU_USERNET=	-device $(NETWORK)$(NETWORK_BUS),netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostfwd=tcp::9090-:9090,hostfwd=tcp::9091-:9091,hostfwd=udp::5309-:5309
```
sanderssj, line 396: Since tracelog is exclusive with ftrace, should we just reuse the same port number?

src/kernel/kernel.c
```diff
@@ -184,3 +184,19 @@ void __attribute__((noreturn)) context_switch_finish(context prev, context next,
     ((void (*)(u64, u64))a)(arg0, arg1);
     runloop();
 }
+
+#ifndef CONFIG_TRACELOG
+void tprintf(symbol tag, tuple attrs, const char *format, ...)
+{
+    vlist a;
+    buffer b = little_stack_buffer(256);
+    vstart(a, format);
```
sanderssj, line 193: I know usage is a little inconsistent in the kernel, but it's probably not a bad idea to include the matching vend() call

src/kernel/tracelog.c
```diff
@@ -0,0 +1,568 @@
...
+#include <net.h>
+
+//#define TRACELOG_DEBUG
+#ifdef TRACELOG_DEBUG
+#define tracelog_debug rprintf
+#else
+#define tracelog_debug(x, ...)
+#endif
+
+#define TRACELOG_HTTP_PORT                 9091
```
sanderssj, line 12: As mentioned earlier, should we just use 9090 since this is exclusive with ftrace

platform/pc/Makefile
wjhun: Ok, I changed all the instances of 9091 with 9090. There isn't strictly a reason that tracelog and ftrace couldn't be used together, but it kind of seems counterintuitive to do function profiling while having debug tracing output enabled.

src/kernel/kernel.c
wjhun: Ok, this is fixed in both definitions of tprintf().

src/kernel/tracelog.c
wjhun: This was left-over cruft from an earlier implementation. I removed it.

PR #1689, opened 2022-03-09
getsockopt/setsockopt: fix handling of TCP_NODELAY socket option

francescolavra:
If the socket on which getsockopt/setsockopt is called is in listening state, it is not possible to get or set TCP options on the lwIP PCB structure, because the structure associated to a listening socket is a struct tcp_pcb_listen, not a struct tcp_pcb. A struct tcp_pcb_listen does not have a TCP flags field, thus trying to access this field leads to a potential corruption of other fields in struct tcp_pcb_listen.
This commit fixes the above issue by adding a tcpflags_t field to the netsock structure and accessing this field in getsockopt/setsockopt if there is no struct tcp_pcb to be accessed. To match the Linux implementation, any socket created from a listening socket inherits its TCP options.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -104,6 +104,7 @@ typedef struct netsock {
     union {
 	struct {
 	    struct tcp_pcb *lw;
+	    tcpflags_t flags;
 	    enum tcp_socket_state state; // half open?
 	} tcp;
 	struct {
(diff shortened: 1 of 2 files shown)


francescolavra:
> The Linux behavior sounds pretty reasonable. Could we not simulate it by stashing such options in the listening netsock struct and then applying those options for each newly-accepted socket in accept_tcp_from_lwip()?

Yes, that makes sense, I'm going to modify the PR accordingly.

sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1690, opened 2022-03-11
Use readv/writev in tools for broader compatibility

sanderssj:
Some versions of OS X and BSD do not have support for preadv and pwritev, so
use lseek and readv/writev instead for the tools.


tools/tfs-fuse.c
```diff
@@ -220,15 +220,16 @@ closure_function(2, 1, void, req_handle,
                 if ((++iov_count == IOV_MAX) || (xfer == total))
                     break;
             }
+            lseek(bound(d), offset, SEEK_SET);
```
francescolavra, line 223: lseek() could be called just once (like you did for mkfs and dump), instead of at each iteration of the while loop.

sanderssj: Fixed!

wjhun (approved):
LGTM

PR #1691, opened 2022-03-15
Volumes: add support for attachment points

francescolavra:
This change adds detection of the "attachment point" of a storage device, expressed as an integer number; the attachment point, when known, identifies uniquely a given volume among all volumes attached to an instance. An unknown attachment point is indicated by the value -1, while 0 indicates the root volume.
The attachment point can be used in a volume mount directive alternatively to the volume ID: if the <volume_id> part of the <volume_id>:<mount_path> syntax has the format %<attach_id> (where <attach_id> is an integer number), then an attached volume whose attachment point equals <attach_id> matches the mount directive.
Detection of a volume attachment point is currently implemented in the xenblk and nvme disk drivers, which are used in AWS instances.

This functionality can be used in the `volume attach` Ops command by specifying a volume identifier argument with format %<attach_id>:<volume_name> instead of <volume_name>.

### DIFF
--- a/src/drivers/ata-pci.c
+++ b/src/drivers/ata-pci.c
@@ -331,7 +331,7 @@ closure_function(3, 1, boolean, ata_pci_probe,
     register_interrupt(irq, (thunk)&dev->irq_handler, "ata pci");
     apply(bound(a),
           storage_init_req_handler(&dev->req_handler, (block_io)&dev->read, (block_io)&dev->write),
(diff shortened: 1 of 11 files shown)


francescolavra:
Merged as https://github.com/nanovms/nanos/commit/e84bf58d92f30677482ae7a79b25a96df4d176d1.

PR #1694, opened 2022-03-17
id heap: implement next-fit search based on allocation size

francescolavra:
Using a first-fit allocation policy in the id heap does not scale well with heap size, because after a substantial amount of allocations, for each subsequent allocation a large number of heap locations needs to be scanned (especially for small-sized allocations) before a free location is found.
Yet, a simple next-fit allocation policy leads to excessive memory fragmentation when allocations of different sizes are interleaved, because a large gap is formed when a large allocation follows a small allocation (to satisfy the id heap alignment property).
This commit implements a more advanced next-fit allocation policy where different next-fit locations are stored for different allocation sizes: this prevents a large allocation from affecting the next-fit pointer for smaller allocations and allows them to fill the gaps left by large allocations.

### DIFF
--- a/src/runtime/heap/id.c
+++ b/src/runtime/heap/id.c
@@ -12,10 +12,14 @@
 #define id_debug(x, ...)
 #endif
 
+/* The location for next-fit search is stored for allocation size values from pagesize to
+ * (pagesize << (NEXT_BIT_COUNT - 1)). */
+#define NEXT_BIT_COUNT  10
+
 typedef struct id_range {
     struct rmnode n;            /* range in pages */
     bitmap b;
-    u64 next_bit;               /* for next-fit search */
+    u64 next_bit[NEXT_BIT_COUNT];   /* for next-fit search */
 } *id_range;
 
 #define ID_HEAP_FLAG_RANDOMIZE  1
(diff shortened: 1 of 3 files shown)


sanderssj (commented):
LGTM

src/runtime/heap/id.c
```diff
@@ -110,10 +122,13 @@ static u64 id_alloc_from_range(id_heap i, id_range r, u64 pages, range subrange)
...
     u64 start_bit = ri.start;
-    if ((i->flags & ID_HEAP_FLAG_RANDOMIZE) && max_start > 0)
+    if ((i->flags & ID_HEAP_FLAG_RANDOMIZE) && max_start > 0) {
         start_bit += random_u64() % max_start;
-    else if (point_in_range(ri, r->next_bit))
-        start_bit = r->next_bit;
+    } else {
+        u64 next_bit = get_next_bit(r, page_order);
+        if (point_in_range(ri, next_bit))
+            start_bit = MIN(next_bit, start_bit + max_start);
```
wjhun, line 130: Trying to understand why max_start is added to start_bit here... is the reasoning that, if next_bit extends past the acceptable range from which we can allocate, then we want to start at the highest possible bit to avoid first-fit-like overhead?

wjhun (approved):
I think the (existing) max_start assignment could use some scrutiny, otherwise LGTM.

src/runtime/heap/id.c
wjhun: I understand that max_start (margin) is relative. But I thought it was only necessary for the randomize computation, so I was wondering why it is being added here (i.e. instead of the assignment being ```MIN(next_bit, start_bit)```).

francescolavra: MIN(next_bit, start_bit) would always equal start_bit, so we wouldn't be using next_bit at all.
On the other end, we cannot just assign start_bit = next_bit because if next_bit is not aligned to the alloc size (which can happen if next_bit is being borrowed from a smaller alloc size) we may not be able to alloc even if there is enough room (because we do first bitmap_alloc_within_range(start_bit, ri.end) and then bitmap_alloc_within_range(ri.start, start_bit)).

PR #1697, opened 2022-03-23
Virtio SCSI / HyperV: add detection of volume attachment point

francescolavra:
This change adds detection of a volume attachment point in the virtio SCSI and HyperV storage drivers.
With these changes, it is possible for GCP and Azure images to mount volumes by using mount directives that reference an attachment point alternatively to a volume name (see also https://github.com/nanovms/ops/pull/1301).


wjhun (approved):
LGTM

PR #1699, opened 2022-03-24
support SO_KEEPALIVE and SO_BROADCAST sockopts, socket_write_udp(): fail on sending to an unconnected socket

wjhun:
This adds support for SO_KEEPALIVE and SO_BROADCAST socket options under the
SOL_SOCKET protocol level. The netsock test has been modified to check that
these options, if set on a listening socket, are inherited by accepted
sockets.

Also, the netsock test was previously failing under Linux, with the write() in
netsock_test_basic() failing with the errno EDESTADDRREQ (89). This was a
result of a missing connect() on the datagram socket, a condition that was
silently being ignored by Nanos. socket_write_udp() has been updated to report
this error when attempting to send to an unconnected udp socket, and a test
for this condition, as well as the required call to connect(), have been added
to the netsock test.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -682,8 +682,12 @@ static sysreturn socket_write_udp(netsock s, void *source, u64 length,
     err_t err = ERR_OK;
 
     /* XXX check how much we can queue, maybe make udp bh */
-    /* XXX check if remote endpoint set? let LWIP check? */
     lwip_lock();
+    if (!dest_addr && !udp_is_flag_set(s->info.udp.lw, UDP_FLAGS_CONNECTED)) {
+        lwip_unlock();
+        return -EDESTADDRREQ;
+    }
+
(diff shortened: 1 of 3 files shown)


src/net/netsyscall.c
```diff
@@ -682,7 +682,9 @@ static sysreturn socket_write_udp(netsock s, void *source, u64 length,
     err_t err = ERR_OK;
 
     /* XXX check how much we can queue, maybe make udp bh */
-    /* XXX check if remote endpoint set? let LWIP check? */
+    if (!dest_addr && !udp_is_flag_set(s->info.udp.lw, UDP_FLAGS_CONNECTED))
```
sanderssj, line 685: If it's possible for the write to race with a socket close then checking the udp.lw should be inside the lwip lock

sanderssj (commented):
LGTM

src/net/netsyscall.c
wjhun: Agreed. I moved this check under the lwip lock.

PR #1701, opened 2022-03-24
Unix domain sockets: implement getsockname and setsockopt

francescolavra:
These features are needed by TiDB (https://github.com/pingcap/tidb).


sanderssj (commented):
LGTM

src/unix/socket.h
```diff
@@ -67,6 +67,8 @@ struct sock {
     sysreturn (*accept4)(struct sock *sock, struct sockaddr *addr,
             socklen_t *addrlen, int flags);
     sysreturn (*getsockname)(struct sock *sock, struct sockaddr *addr, socklen_t *addrlen);
+    sysreturn (*setsockopt)(struct sock *sock, int level,
```
wjhun, line 70: Why not add a getsockopt method too?

wjhun (approved):
LGTM

src/unix/socket.h
francescolavra: OK, I'm going to add a getsockopt method too. As of now it is implemented by network sockets only, though.

PR #1700, opened 2022-03-24
implement SIOCGIFNAME socket ioctl, handle netdevice ioctls for any socket type

wjhun:
This change implements the SIOCGIFNAME ioctl for getting an interface name by
index. It also moves the various SIOC?* netdevice-related ioctls from
netsock_ioctl() to socket_ioctl(), as they are valid for any type of socket,
regardless of family or type (as stated in netdevice(7)). These changes are
required for if_indextoname(3) to work properly.

PR #1703, opened 2022-03-25
add netlink support for RTM_GETROUTE requests 

wjhun:
This work allows IP routing table information to be extracted from the kernel
using RTM_GETROUTE netlink messages. As with the remainder of netlink code,
only IPv4 is supported at the moment. This support is sufficient for
retrieving the gateway address for each interface. The netlink runtime test
has been expanded to test this functionality as well as optionally dump
details from rtmsg responses received from the kernel.

Based on PR #1700

### DIFF
--- a/src/unix/netlink.c
+++ b/src/unix/netlink.c
@@ -189,6 +189,98 @@ enum {
     IFA_FLAGS,
 };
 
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
LGTM

PR #1704, opened 2022-03-29
Unix fault handler: do not change CPU state

francescolavra:
When the fault handler is invoked, the CPU state is the same as the state when the interrupt was triggered. Therefore, it is not necessary for the fault handler to set the state when returning the current context; in fact, this causes a wrong state to be set when a user thread generates a page fault that is resolved without blocking, which may cause any subsequent exception generated by the thread to be handled as if it was generated by the kernel.


wjhun (approved):
LGTM

PR #1706, opened 2022-04-01
unix_fault_handler: deliver SIGSEGV on user page faults from syscall context

### LINKED ISSUE #1705: kernel write to invalid user stack page causes panic
This appears when running the sigoverflow runtime test on my local machine. The unhandled page fault is triggered by a call to rt_sigprocmask(), probably resulting from a pthread_create(), when the user stack has overflown - a condition the test is trying to induce. That the touching of the page occurs from within a syscall as opposed to userspace is mere chance; in other cases, the test passes fine.

To solve this, I propose handling kernel page faults on user addresses as if they occurred from userspace; that is, send a SIGSEGV to a process when a fault on an invalid page occurs, regardless of whether the fault occurs in user or kernel mode.

```
(arguments:(0:test)
 children:(sigoverflow:())
 debug_exit:t
 environment:(PWD:/
              USER:bobby)
 expected_exit_code:11
 program:/sigoverflow)
qemu-system-x86_64 -machine q35 -m 2G  -display none -device pcie-root-port,port=0x10,chassis=1,id=pci.1,bus=pcie.0,multifunction=on,addr=0x3 -device pcie-root-port,port=0x11,chassis=2,id=pci.2,bus=pcie.0,addr=0x3.0x1 -device pcie-root-port,port=0x12,chassis=3,id=pci.3,bus=pcie.0,addr=0x3.0x2 -serial stdio -drive if=none,id=hd0,format=raw,file=/home/wjhun/src/nanos/master1/output/image/disk.raw -device virtio-scsi-pci,bus=pci.2,addr=0x0,id=scsi0 -device scsi-hd,bus=scsi0.0,drive=hd0 -device isa-debug-exit -no-reboot   -device virtio-net,bus=pci.3,addr=0x0,netdev=n0 -netdev user,id=n0,hostfwd=tcp::8080-:8080,hostfwd=tcp::9090-:9090,hostfwd=udp::5309-:5309 -enable-kvm -cpu host || exit 1
en1: assigned 10.0.2.15
expecting signal stack overflow to segfault...

No vmap found for page fault in kernel mode
cpu: 0, context type: 2
 interrupt: 000000000000000e (Page fault)
     frame: ffffc00000ca8000
      type: syscall
active_cpu: 0000000000000000
 stack top: ffffc00000cafff0
error code: 0000000000000002
   address: 000000007fbfcfd0

   rax: 0000000000000200
   rbx: ffffc00000ca8000
   rcx: 0000000000000002
   rdx: 000000007fbfcfd0
   rsi: 000000007fbfd050
   rdi: 0000000000000000
   rbp: ffffc00000caffe8
   rsp: ffffc00000caffb0
    r8: 0000000000000006
    r9: 000000007fbfcfd0
   r10: 0000000000000008
   r11: 0000000000000246
   r12: ffffc00001607000
   r13: 000000000000000e
   r14: 0000000000000000
   r15: ffffc00000ca80f8
   rip: ffffffff800a60ae	(rt_sigprocmask + 000000000000004e/00000000000001a0)
rflags: 0000000000010046
    ss: 0000000000000008
    cs: 0000000000000010
    ds: 0000000000000000
    es: 0000000000000000
fsbase: 0000000000000000
gsbase: 0000000000000000

frame trace:

stack trace:
ffffc00000caffb0:   ffffffff800bfe33	(syscall_handler + 00000000000003c3/00000000000005dd)
ffffc00000caffb8:   0000000000000000
ffffc00000caffc0:   ffffc00000ca8000
ffffc00000caffc8:   000000000042d600
ffffc00000caffd0:   0000000000000000
ffffc00000caffd8:   00000000004f2018
ffffc00000caffe0:   00000000004004e8
```

wjhun:
This addresses a common problem whereby the kernel can panic on a user page
fault that occurs within a syscall context. Such faults are usually caused by
the user program supplying an invalid buffer address, and as such should be
treated as segmentation violations. This handles such cases by marking the
relevant thread's current syscall as having been "abandoned," thereafter
delivering SIGSEGV to the violating process - just as if the fault had
occurred from within user mode. The abandoned syscall is "cleaned up" in
thread_return() before returning control to the program (albeit not
completely; see the comment in that function).

This also addresses a few cosmetic issues with output that occurs after
terminal signal delivery. To aid the debugging of such violations, frames of
the the offending thread and syscall_context (if applicable) are dumped.

The second commit fixes the incorrect setting of the siginfo si_addr
(sifields.sigfault.addr) field for SIGFPE, SIGILL and SIGTRAP signals.
The signal test has been amended with a test that checks validity of
the reported address after a SIGILL, and also includes a test for a terminal
page fault from within a syscall.

Resolves #1705

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -141,9 +141,9 @@ const char *string_from_mmap_type(int type)
 }
 
 #define format_protection_violation(vaddr, ctx, vm, str)        \
-    "page_protection_violation%s\naddr 0x%lx, rip 0x%lx, "     \
+    "page_protection_violation%s\naddr 0x%lx, pc 0x%lx, "       \
     "error %s%s%s vm->flags (%s%s %s%s%s)\n",                   \
-        str, vaddr, frame_return_address(ctx->frame),           \
+        str, vaddr, frame_fault_pc(ctx->frame),                 \
         is_write_fault(ctx->frame) ? "W" : "R",                 \
         is_usermode_fault(ctx->frame) ? "U" : "S",              \
         is_instruction_fault(ctx->frame) ? "I" : "D",           \
@@ -191,13 +191,12 @@ define_closure_function(1, 1, context, unix_fault_handler,
 {
     thread t = bound(t);
     const char *errmsg = 0;
-
-    u64 vaddr = fault_address(ctx->frame);
+    u64 fault_pc = frame_fault_pc(ctx->frame);
     boolean user = is_usermode_fault(ctx->frame);
 
     if (is_div_by_zero(ctx->frame)) {
         if (current_cpu()->state == cpu_user) {
-            deliver_fault_signal(SIGFPE, t, vaddr, FPE_INTDIV);
+            deliver_fault_signal(SIGFPE, t, fault_pc, FPE_INTDIV);
             schedule_thread(t);
             return 0;
         } else {
@@ -206,8 +205,8 @@ define_closure_function(1, 1, context, unix_fault_handler,
         }
     } else if (is_illegal_instruction(ctx->frame)) {
         if (current_cpu()->state == cpu_user) {
-            pf_debug("invalid opcode fault in user mode, rip 0x%lx", ctx->frame[SYSCALL_FRAME_PC]);
-            deliver_fault_signal(SIGILL, t, vaddr, ILL_ILLOPC);
+            pf_debug("invalid opcode fault in user mode, rip 0x%lx", fault_pc);
+            deliver_fault_signal(SIGILL, t, fault_pc, ILL_ILLOPC);
             schedule_thread(t);
             return 0;
         } else {
@@ -216,8 +215,8 @@ define_closure_function(1, 1, context, unix_fault_handler,
         }
     } else if (is_breakpoint(ctx->frame)) {
         if (current_cpu()->state == cpu_user) {
-            pf_debug("breakpoint in user mode, rip 0x%lx", ctx->frame[SYSCALL_FRAME_PC]);
-            deliver_fault_signal(SIGTRAP, t, vaddr, TRAP_BRKPT);
+            pf_debug("breakpoint in user mode, rip 0x%lx", fault_pc);
+            deliver_fault_signal(SIGTRAP, t, fault_pc, TRAP_BRKPT);
             schedule_thread(t);
             return 0;
         } else {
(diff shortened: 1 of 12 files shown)


src/unix/unix.c
```diff
@@ -232,18 +232,26 @@ define_closure_function(1, 1, context, unix_fault_handler,
     } else if (is_page_fault(ctx->frame)) {
         pf_debug("page fault, vaddr 0x%lx, ctx %p, type %d, pc 0x%lx",
                  vaddr, ctx, ctx->type, ctx->frame[SYSCALL_FRAME_PC]);
+        if (vaddr < PAGESIZE /* no zero page access */ ||
+            vaddr > (USER_LIMIT - PAGESIZE)) {
```
sanderssj, line 236: Is there something special about the last user page that it can't be used?

test/runtime/signal.c
```diff
@@ -584,20 +585,29 @@ sigsegv_handler(int signo)
...
         asm volatile("hlt");
 #endif
 #ifdef __aarch64__
         asm volatile("wfi");
 #endif
+        break;
+    case 2:
+        /* should cause sigsegv by passing a bad buffer address to a syscall */
+        open((const char *)BAD_LOAD, O_RDONLY);
+        assert(0); /* shouldn't arrive here... */
```
sanderssj, line 609: On my mac, the open returns EFAULT instead of generating a SIGSEGV and the test fails at this assert.

sanderssj (commented):
LGTM, just had a couple questions.

src/unix/unix.c
wjhun: Thanks, this shouldn't be calling bug. I'll change this to segv for zero page or kernel memory access. I also mistook vaddr as being the page address, so the latter expression should be ```vaddr >= USER_LIMIT```.

src/unix/unix.c
```diff
@@ -232,18 +232,25 @@ define_closure_function(1, 1, context, unix_fault_handler,
     } else if (is_page_fault(ctx->frame)) {
         pf_debug("page fault, vaddr 0x%lx, ctx %p, type %d, pc 0x%lx",
                  vaddr, ctx, ctx->type, ctx->frame[SYSCALL_FRAME_PC]);
-        vmap vm = vmap_from_vaddr(t->p, vaddr);
+        vmap vm;
+        if (vaddr >= PAGESIZE && vaddr < USER_LIMIT)
+            vm = vmap_from_vaddr(t->p, vaddr);
+        else
+            vm = INVALID_ADDRESS;
```
francescolavra, line 239: It seems to me we could do without the (vaddr >= PAGESIZE && vaddr < USER_LIMIT) check and just leave `vmap vm = vmap_from_vaddr(t->p, vaddr)` as it was before: if vaddr is not between PAGESIZE and USER_LIMIT, vmap_from_vaddr() will return INVALID_ADDRESS.
Besides, there is a `if (vaddr >= USER_LIMIT) goto bug` at the beginning of this function (which in my opinion should not be there), so vaddr cannot be < USER_LIMIT here.

wjhun: I realize the vmap lookup should fail for out of bounds addresses, but I think the negligible cost of this range check is worth the protection against the risk that t->p->vmaps could ever contain bad entries, however minor, as such a condition might possibly create a path to an exploit. I removed the vaddr check at the top of the function, though.

wjhun:
I added a commit to fix the incorrect setting of the siginfo si_addr (sifields.sigfault.addr) field for SIGFPE, SIGILL and SIGTRAP signals.

sanderssj (commented):
LGTM

francescolavra (commented):
LGTM

PR #1707, opened 2022-04-08
aarch64: add SMP support

### LINKED ISSUE #1328: interface for per-cpu init callbacks and resource allocation
We have various areas around the kernel that want some state stored in the per-cpu info. We could use some interface for registering per-cpu initialization callbacks to be invoked when CPUs come online.

One thing to take into consideration when coming up with such an interface is that we may eventually want the ability to bring cores on and offline. There isn't a pressing need to implement this now, but it would be nice if interfaces started supporting it. (So, for instance, per-cpu init functions should not assume that they are only being called during kernel init...)

francescolavra:
This change adds support for multiple CPUs in the ARM port. The available CPUs are detected by walking the ACPI MADT and then brought online via the PSCI firmware.
As part of these changes, an interface for generic per-CPU init callbacks is being added, replacing the platform_timer_percpu_init thunk.
Closes #1328, #1397.


sanderssj (commented):
LGTM

src/aarch64/gic.c
```diff
@@ -157,26 +158,70 @@ boolean gic_int_is_pending(int irq)
     return pending;
 }
 
+static void gicr_get_base(void)
+{
+    cpuinfo ci = current_cpu();
+    u64 mpid = mpid_from_cpuid(ci->id);
```
wjhun, line 164: Could this not read the mpid directly from the psr?

wjhun (approved):
Looks good!

src/aarch64/gic.c
francescolavra: Yes, I changed it to use read_mpid() instead of mpid_from_cpuid().

PR #1708, opened 2022-04-08
select(2): allow blocking indefinitely with zero nfds

wjhun:
Our select implementation incorrectly assumed that a call with zero nfds
specified and an indefinite timeout should return immediately. Such a call may
legitimately be used to wait for signal reception, though, so
select_internal() has been changed here to enter an indefinite, interruptible
block whenever a NULL timeout is specified, regardless of the value of
nfds. The signal test now checks that such a call to select behaves in the
same manner as pause(2).

### DIFF
--- a/test/runtime/signal.c
+++ b/test/runtime/signal.c
@@ -45,15 +45,20 @@ static void yield_for(volatile int * v)
         fail_error("timed out");
 }
 
-static void * tgkill_test_pause(void * arg)
+static void * tgkill_test_pause(void *use_select)
 {
+    int rv;
     child_tid = syscall(SYS_gettid);
     sigtest_debug("child enter, tid %d\n", child_tid);
+    if (use_select) {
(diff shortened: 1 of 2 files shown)


test/runtime/signal.c
```diff
@@ -45,15 +45,20 @@ static void yield_for(volatile int * v)
...
 }
 
-static void * tgkill_test_pause(void * arg)
+static void * tgkill_test_pause(void *use_select)
 {
+    int rv;
     child_tid = syscall(SYS_gettid);
     sigtest_debug("child enter, tid %d\n", child_tid);
+    if (use_select) {
+        rv = syscall(SYS_pselect6, 0, NULL, NULL, NULL, NULL);
```
francescolavra, line 54: pselect() has 6 parameters (nfds, readfds, writefds, exceptfds, timeout, sigmask), here we have 5 arguments.

wjhun: Thanks, fixed.

PR #1711, opened 2022-04-15
Unix special files: add /proc/meminfo

francescolavra:
In the current implementation, the following info is returned: MemTotal, MemFree, MemAvailable (whose value is approximated as the sum of free memory and page pache occupancy) and Cached.


wjhun (approved):
LGTM

PR #1712, opened 2022-04-15
Add support for UBSan (undefined behavior sanitizer)

francescolavra:
This change allows specifying an `UBSAN` Makefile variable that enables instrumenting the compiler-generated code with instructions to catch bugs that can result in undefined behavior. When UBSan detects an error, it generates a software trap (on x86-64, an "invalid opcode (UD2)" exception; on aarch64, a "brk" exception that results in an ESR value of 0xf20003e8; on risc-v, a breakpoint exception), which triggers a crash dump in the kernel.
UBSan checks for unaligned accesses and null pointers are being disabled because the kernel contains legitimate uses of unaligned accesses (mainly in third-party code such as acpica and lwIP) and null pointers (in the x86-64 code for secondary CPU bringup).
`UBSAN=1` is being added to the make command line in the build-pc-memdebug CircleCI job (now renamed to build-pc-debug) so that UBSan is enabled as part of the CI checks.

The `-fsanitize-undefined-trap-on-error` compiler flag is being used so that UBSan generates a software trap instead of calling an error handler function, because instrumenting the entire kernel code with calls to error handlers would result in a kernel binary file significantly larger (mainly because source file and line number information needs to be kept for each error path), which cannot be loaded by the stage2 bootloader. If more information is needed to find the cause of an UBSan error, the kernel can be rebuilt by omitting the `-fsanitize-undefined-trap-on-error` compiler flag for the source file where the error happens (which can be identified from the kernel crash dump via the instruction that caused the exception), and the error handler functions need to be included in the kernel binary. The following code provides an implementation of the error handler functions:
```
#include "kernel.h"

struct source_location {
    const char *file_name;
    u32 line;
    u32 column;
};

enum {
   type_kind_integer = 0x000,
   type_kind_float = 0x0001,
   type_kind_unkown = 0xffff,
};

struct type_descriptor {
    u16 kind;
    u16 info;
    char name[1];
};

struct out_of_bounds_data {
    struct source_location loc;
    struct type_descriptor *array_type;
    struct type_descriptor *index_type;
};

struct shift_out_of_bounds_data {
    struct source_location loc;
    struct type_descriptor *lhs_type;
    struct type_descriptor *rhs_type;
};

struct type_mismatch_data {
    struct source_location loc;
    struct type_descriptor *type;
    u8 log_alignment;
    u8 type_check_kind;
};

struct pointer_overflow_data {
    struct source_location loc;
};

struct overflow_data {
    struct source_location loc;
    struct type_descriptor *type;
};

static void ubsan_prologue(struct source_location *loc)
{
    rprintf("UBSan detected error at %s:%d:%d\n",loc->file_name, loc->line, loc->column);
}

static void ubsan_epilogue(void)
{
    dump_context(get_current_context(current_cpu()));
    vm_exit(VM_EXIT_FAULT);
}

void __ubsan_handle_out_of_bounds(struct out_of_bounds_data *data, u64 index)
{
    ubsan_prologue(&data->loc);
    rprintf("out of bounds: index %ld, array type %s, index type %s\n", index,
            data->array_type->name, data->index_type->name);
    ubsan_epilogue();
}

void __ubsan_handle_shift_out_of_bounds(struct shift_out_of_bounds_data *data, long lhs, long rhs)
{
    ubsan_prologue(&data->loc);
    rprintf("shift out of bounds: LHS %ld (%s), RHS %ld (%s)\n", lhs, data->lhs_type->name,
            rhs, data->rhs_type->name);
    ubsan_epilogue();
}

void __ubsan_handle_type_mismatch_v1(struct type_mismatch_data *data, void *pointer)
{
    ubsan_prologue(&data->loc);
    rprintf("type mismatch: type %s, alignment %d, pointer %p\n", data->type->name,
            1 << data->log_alignment, pointer);
    ubsan_epilogue();
}

void __ubsan_handle_pointer_overflow(struct pointer_overflow_data *data, void *base, void *result)
{
    ubsan_prologue(&data->loc);
    rprintf("pointer overflow: base %p, result %p\n", base, result);
    ubsan_epilogue();
}

void __ubsan_handle_negate_overflow(struct overflow_data *data, long old_val)
{
    ubsan_prologue(&data->loc);
    rprintf("negate overflow: type %s, old value 0x%lx\n", data->type->name, old_val);
    ubsan_epilogue();
}

#define UBSAN_OVERFLOW_HANDLER(op)                                                              \
    void __ubsan_handle_##op##_overflow(struct overflow_data *data, void *lhs, void *rhs)       \
    {                                                                                           \
        ubsan_prologue(&data->loc);                                                             \
        rprintf(#op " overflow: type %s, LHS 0x%lx, RHS 0x%lx\n", data->type->name, lhs, rhs);  \
        ubsan_epilogue();                                                                       \
    }

UBSAN_OVERFLOW_HANDLER(add)
UBSAN_OVERFLOW_HANDLER(sub)
UBSAN_OVERFLOW_HANDLER(mul)
UBSAN_OVERFLOW_HANDLER(divrem)
```

### DIFF
--- a/src/x86_64/apic.c
+++ b/src/x86_64/apic.c
@@ -96,7 +96,7 @@ void msi_format(u32 *address, u32 *data, int vector)
     u32 dm = 0;             // destination mode: ignored if rh == 0
     u32 rh = 0;             // redirection hint: 0 - disabled
     u32 destination = 0;    // destination APIC
-    *address = (0xfee << 20) | (destination << 12) | (rh << 3) | (dm << 2);
+    *address = (0xfeeu << 20) | (destination << 12) | (rh << 3) | (dm << 2);
 
     u32 mode = 0;           // delivery mode: 000 fixed, 001 lowest, 010 smi, 100 nmi, 101 init, 111 extint
     u32 level = 0;          // trigger level: 0 - deassert, 1 - assert
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
       - run: gsutil setmeta -h "Custom-Time:`date -Iseconds`" gs://nanos/release/${COMMIT_ID}/*
       - run: gsutil acl ch -u AllUsers:R gs://nanos/release/${COMMIT_ID}/*
 
-  build-pc-memdebug:
+  build-pc-debug:
     docker:
       # specify the version
       - image: cimg/go:1.16
@@ -45,13 +45,13 @@ jobs:
       - run: sudo apt-get update
       - run: sudo apt-get install nasm qemu-system-x86 ent ruby rustc
       - run: cd .. && git clone git@github.com:nanovms/ops.git
-      - run: make MEMDEBUG=all
+      - run: make MEMDEBUG=all UBSAN=1
       - run: curl https://ops.city/get.sh -sSfL | sh
       - run:
           command: |
             OPS_DIR=$HOME/.ops
             PATH=$HOME/.ops/bin:$PATH
-            make MEMDEBUG=all test-noaccel
+            make MEMDEBUG=all UBSAN=1 test-noaccel
 
   nightly-build:
     docker:
@@ -209,7 +209,7 @@ workflows:
     jobs:
       - build-pc
       - build-virt
-      - build-pc-memdebug
+      - build-pc-debug
       - build-riscv
   nightly:
     triggers:
--- a/rules.mk
+++ b/rules.mk
@@ -128,6 +128,10 @@ endif
 KERNCFLAGS+=	-fno-omit-frame-pointer
 KERNLDFLAGS=	--gc-sections -z max-page-size=4096 -L $(OUTDIR)/klib
 
+ifneq ($(UBSAN),)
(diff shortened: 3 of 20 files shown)


src/x86_64/apic.c
```diff
@@ -96,7 +96,7 @@ void msi_format(u32 *address, u32 *data, int vector)
     u32 dm = 0;             // destination mode: ignored if rh == 0
     u32 rh = 0;             // redirection hint: 0 - disabled
     u32 destination = 0;    // destination APIC
-    *address = (0xfee << 20) | (destination << 12) | (rh << 3) | (dm << 2);
```
sanderssj, line 99: This is probably too nitpicky, but if we wanted to keep the shift value separate from the operand, would the ULL suffix on the 0xfee constant eliminate the UBSan error?

sanderssj (commented):
LGTM

src/x86_64/apic.c
francescolavra: Good point, indeed specifying the 0xfee constant as a signed number (as in `0xfeeu`) eliminates the UBSan error, which makes total sense. I updated the code accordingly.

wjhun (approved):
LGTM

PR #1713, opened 2022-04-19
syscall_handler: do not bypass syscall context resume

wjhun:
The syscall handler must resume the syscall context installed on low-level
entry, completing the context switch. syscall_handler() was previously exiting
on shutdown or an invalid syscall without performing this restore, causing an
assertion failure when the pause method for the context is called on the next
context switch. This materialized when the clone() wrapper in glibc 2.34
attempted to invoke the unimplemented clone3(2) syscall. With this fix, glibc
safely handles the -ENOSYS from the invalid syscall and falls back to using
clone(2).

https://github.com/nanovms/ops/issues/1321

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -2444,14 +2444,6 @@ void syscall_handler(thread t)
     syscall_restart_arch_setup(f);
     set_syscall_return(t, -ENOSYS);
 
-    if (shutting_down)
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

francescolavra:
LGTM

PR #1714, opened 2022-04-20
ACPI: add support for entering S3 sleeping state (VM suspend)

francescolavra:
This change implements a handler for the ACPI sleep button event, which is triggered in a GCP instance during a VM suspend operation. This handler communicates with the ACPI firmware to put the system in the ACPI S3 state, at which point the VM is suspended; then, when a VM resume operation is requested, guest execution resumes from where it was suspended (at the call to AcpiEnterSleepState()), after which the sleep button handler indicates to the ACPI firmware that wakeup has completed and the VM resumes normal operation.

### DIFF
--- a/src/drivers/acpi.c
+++ b/src/drivers/acpi.c
@@ -187,6 +187,32 @@ void init_acpi_tables(kernel_heaps kh)
     AcpiGetDevices(NULL, acpi_device_handler, NULL, NULL);
 }
 
+static UINT32 acpi_sleep(void *context)
+{
+    acpi_debug("sleep");
+    const u64 sleep_state = 3;  /* S3 state */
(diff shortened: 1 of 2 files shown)


wjhun (approved):
LGTM

urbien:
I wonder if this can be used to reduce cold boot times for serverless applications?

francescolavra:
> I wonder if this can be used to reduce cold boot times for serverless applications?

Yes, using the VM suspend feature can reduce boot times in two ways:
- moving a VM from a 'suspended' state to a 'running' state is typically faster than moving it from 'stopped' to 'running' (at least that's what I observed in Google Cloud)
- the process running in the VM does not have to go through its initialization steps (the gain in boot time reduction here depends on the complexity of the application)

PR #1715, opened 2022-04-23
Copy flags from listening socket at accept syscall

sanderssj:
A socket generated by an accept syscall from a listening socket should inherit
the tcp flags at the time of the accept syscall, not when the connection is
initiated at the network stack. This way, any changes to the listener's tcp flags
between the initiation of the connection and the accept syscall are appropriately
reflected in the child.


wjhun (approved):
LGTM

PR #1717, opened 2022-04-26
Free status tuple when file write fails

sanderssj:
Thousands of failed write syscalls can eventually exhaust memory if the status tuple is not freed.


wjhun (approved):
LGTM

francescolavra (commented):
LGTM

PR #1718, opened 2022-04-26
Page cache: implement write-back policy for sync to storage device

### LINKED ISSUE #1129: pagecache: implement true write-back support
To further optimize filesystem write performance, pages on the dirty cache list can be written back with some predetermined lag (governed by an upper bound) in order to coalesce multiple writes to each page and aggregate block device writes as much as possible. In theory, this can provide a performance boost - at some risk of data loss (should the system crash between a write and sync to disk).

See #1125, #1128

francescolavra:
The existing write-through policy to synchronize the page cache to the storage device is being replaced with a write-back policy, which allows coalescing successive writes to the same file locations and aggregating writes to adjacent locations, and therefore minimizes the number of disk write operations.
pagecache_write_sg_finish(), instead of triggering a filesystem write operation, marks relevant pages as dirty before letting the write syscall return to userspace. In order to correctly update the file length value in the filesystem metadata before returning from a write syscall, the TFS code has been changed so that the file length is updated when calling either filesystem_check_or_reserve_extent (which happens during a write syscall), or filesystem_storage_write (which happens during a cache writeback).
The existing scan timer, previously used only to scan file-backed mappings to look for dirty pages, is now in charge of committing all dirty pages resulting from either memory writes to file-backed mappings, or direct file writes.
In order to avoid unchecked accumulation of pending disk write operations when the application is doing heavy file writes, the
page cache code keeps track of whether a writeback is in progress, and the scan timer only initiates a new writeback after completion of any previous writeback.
Closes #1129.

The first commit is a fix to an existing issue that was causing the kernel to hang when executing a file write stress test.

### DIFF
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -496,8 +496,8 @@ static boolean pagecache_set_dirty(pagecache_node pn, range r)
     return true;
 }
 
-closure_function(6, 1, void, pagecache_write_sg_finish,
-                 pagecache_node, pn, range, q, sg_list, sg, status_handler, completion, boolean, complete, context, saved_ctx,
+closure_function(5, 1, void, pagecache_write_sg_finish,
+                 pagecache_node, pn, range, q, sg_list, sg, status_handler, completion, context, saved_ctx,
                  status, s)
 {
     pagecache_node pn = bound(pn);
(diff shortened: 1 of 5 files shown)


sanderssj (commented):
LGTM

PR #1719, opened 2022-04-26
build_exec_stack: point AT_RANDOM at random values

wjhun:
build_exec_stack() was passing the user stack pointer in the AT_RANDOM field
without actually placing random data at that location. This allocates 16 bytes
at the top of the stack, fills it with random data, and passes that to the
user program via the AT_RANDOM field. This random data is needed by glibc to
produce entropy when initializing the pointer and stack guards.


sanderssj (commented):
LGTM

PR #1721, opened 2022-05-01
In-kernel tests

### LINKED ISSUE #937: in-kernel tests
We could use a way to invoke tests within the kernel for facilities that can't be unit tested, e.g. page mapping calls in x86/page.c and anything else not necessarily exposed through the system call interface. It could be triggered by a debug flag in the manifest, placing the in-kernel test on the runqueue instead of a user program.

francescolavra:
In-kernel tests can be used to check the correct behavior of kernel code that cannot be easily exercised via syscalls. In order to
avoid bloating the main kernel binary, in-kernel tests can be implemented as klibs.
This change set adds to the klib source folder a test subfolder that will contain the source of klibs for in-kernel tests. In image
files, klib binaries are meant to be installed in a test subfolder of the klib root directory; if the root tuple contains a `klib_test` attribute, the kernel loads all klibs found in the test subfolder, which will then execute the in-kernel tests.
The existing "test" klib (which checks the correct behavior of the klib loading code) is being moved to the test subfolder, and the "klibs" runtime test application is being renamed as "ktest" (because it will run different in-kernel tests).
Klibs containing in-kernel tests are not included in tarballs created by CircleCI automated builds.
The last commit adds to the ktest runtime test image an additional klib that tests the kernel page table code.
The second-to-last commit is a fix to a bug that has been exposed by the page table in-kernel test.

Closes #937.


sanderssj (commented):
LGTM

PR #1722, opened 2022-05-01
x86: add support for FSGSBASE instructions

francescolavra:
The FSGSBASE instructions allow reading and writing the FS and GS registers with dedicated CPU instructions, avoiding the need for the arch_prctl syscall and therefore providing performance benefits.
This change adds detection of the FSGSBASE capability from the CPU; if this capability is detected, the FSGSBASE instructions are enabled and their support is reported to userspace in the auxiliary vector.


wjhun (approved):
LGTM

francescolavra:
In the thread runtime test, I added a new test case that retrieves the FS register value (which is used for thread-local storage) and verifies that the value is different for different threads

PR #1725, opened 2022-05-09
Fix load program limits and exit syscall

sanderssj:
An elf binary can ask for loading of a program segment above the current limit (1GB) causing a panic. The current
limit appears arbitrary, so it has been adjusted to 3GB, and the user stack allocation has been moved to just under
4GB so that it will not conflict with program segments.

Additionally, the exit syscall was not triggering a shutdown if called by the last thread of a program (in lieu of exit_group).
This has also been fixed.

### DIFF
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -36,8 +36,8 @@ static void build_exec_stack(process p, thread t, Elf64_Ehdr * e, void *start,
 {
     exec_debug("build_exec_stack start %p, tid %d, va 0x%lx\n", start, t->tid, va);
 
-    /* allocate process stack at top of first 2gb of address space */
-    u64 stack_start = 0x80000000 - PROCESS_STACK_SIZE;
+    /* allocate process stack at top of 32-bit address space */
+    u64 stack_start = 0x100000000 - PROCESS_STACK_SIZE;
     if (aslr)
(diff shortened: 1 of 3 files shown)


src/unix/exec.c
```diff
@@ -36,8 +36,8 @@ static void build_exec_stack(process p, thread t, Elf64_Ehdr * e, void *start,
 {
     exec_debug("build_exec_stack start %p, tid %d, va 0x%lx\n", start, t->tid, va);
 
-    /* allocate process stack at top of first 2gb of address space */
-    u64 stack_start = 0x80000000 - PROCESS_STACK_SIZE;
+    /* allocate process stack at top of 32-bit address space */
+    u64 stack_start = 0x100000000 - PROCESS_STACK_SIZE;
```
francescolavra, line 40: The [0x80000000, 0x100000000) address range is used for mmap() with the MAP_32BIT flag, so if part of this range is used for program segments or the stack, this needs to be reflected in the p->virtual32 heap (on x86).

sanderssj: Thanks for pointing that out! I added the elf load segments to the p->virtual32 heap for x86.

francescolavra (approved):
LGTM

PR #1726, opened 2022-05-11
AWS: implement sending memory metrics to CloudWatch

francescolavra:
This change set adds a new klib that implements sending memory utilization metrics to AWS CloudWatch.
Analogously to the implementation in the Linux CloudWatch agent, the metrics being sent are under the "CWAgent" namespace, and have an associated dimension whose name is "host" and whose value is an instance identifier formatted as in the following example: "ip-111-222-111-222.us-west-1.compute.internal".
The list of supported metrics is: mem_used, mem_used_percent, mem_available, mem_available_percent, mem_total, mem_free and mem_cached. A description for each of these metrics can be found at https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html, in the section for the Linux CloudWatch agent.
In order for the cloudwatch klib to retrieve the appropriate credentials needed to communicate with the CloudWatch server, the AWS instance on which it runs must be associated to an IAM role with the CloudWatchAgentServerPolicy, as described in https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/create-iam-roles-for-cloudwatch-agent.html.
Memory metrics are defined as standard-resolution metrics, i.e. they are stored with a resolution of 60 seconds.

The cloudwatch klib is configured via a `cloudwatch` tuple in the image manifest: sending of memory metrics is enabled by specifying a sending interval (expressed in seconds) via a `mem_metrics_interval` attribute inside the `cloudwatch` tuple.
The cloudwatch klib depends on the tls klib for cryptographic operations (which are needed in order to sign the requests being
sent to the CloudWatch server).
Example Ops configuration to send memory metrics with a 60-second interval:
```
{
  "RunConfig": {
    "Klibs": ["cloudwatch", "tls"]
  },
  "ManifestPassthrough": {
    "cloudwatch": {
      "mem_metrics_interval": "60"
    }
  }
}
```


sanderssj (approved):
LGTM

PR #1730, opened 2022-05-25
virtio block driver: fix handling of VIRTIO_BLK_F_SEG_MAX feature

francescolavra:
The set of features negotiated between a virtio driver and a device is the logical AND of the features supported by the driver and
those supported by the device. The driver was failing to set VIRTIO_BLK_F_SEG_MAX among its supported features, which was preventing the driver from detecting that this feature is supported by the virtio block device implementation in Qemu. Setting this feature bit allows the driver to retrieve from the device the maximum number of segments (which is 254 in Qemu v5.2.0) allowed in a virtio request.
The previously held assumption that a device that does not support this feature can handle up to (virtqueue_size - 2) segments is wrong and was causing a failure to boot on AWS Firecracker (which does not support this feature). This change fixes the issue by setting the maximum number of segments to 1 when the feature is not supported.

### DIFF
--- a/src/virtio/virtio_storage.c
+++ b/src/virtio/virtio_storage.c
@@ -92,6 +92,9 @@ struct virtio_blk_config {
 #define VIRTIO_BLK_S_IOERR      1
 #define VIRTIO_BLK_S_UNSUPP     2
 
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1731, opened 2022-05-26
Improve ntp klib accuracy and correction rate

sanderssj:
Instead of using a single rate to correct both frequency and offset errors,
these changes use a least squares linear regression to estimate the values
separately using a relatively large number of data points, where the slope
represents the frequency error and the intercept equals the offset. With this
method of calculating frequency error, it the ntp klib can now track
the reference clock with sub-millisecond per second accuracy. The frequency
error gets applied to all clocks except for MONOTONIC_RAW (by definition).
The slewing adjustments are applied only to the realtime clock. With the
separation of error components, the offset error can be much more
aggressively slewed away, correcting offsets of over a second within a single
query interval. Older samples are adjusted for offset changes so that the
linear regression is not influenced by the error introduced by the slewing
frequency. The slew rate is inversely proportional to the size of the offset
so that errors in slewing are reduced as the offset size approaches zero.
All computations are performed with a 32-bit fractional signed
fixed point type which has enough precision to keep errors in the
nanoseconds range. To prevent a non-sensical linear regression
from setting the clock to a bad value, the estimated frequency
is clamped to a reasonable limit, and enough bad regression
computations will result in resetting regression state to discard
any bad values contributing to the problem.

Additional sanity checks have been added for the ntp response packet
based on recommendations by the NTP RFC. These mostly involve
discarding packets as the result of server or network misbehavior.

Time related functions which read vdso state now check a generation
number to ensure that all vdso_dat reads are consistent with each
other.

Two new manifest options have been added: ntp_max_slew_ppm which
determines the offset slewing rate and ntp_max_freq_ppm which controls
the maximum frequency error rate.

The initial polling interval is set to the minimum at startup in order to calculate
initial frequency and offset errors so they will be corrected within the
first few minutes of runtime.

In order to help with debugging, a new print format %f has been added for
printing s64 fixed point values. This is similar to %T for timestamps
but with more decimal values printed.

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -1,6 +1,13 @@
 #include <kernel.h>
 #include <lwip.h>
 
+//#define NTP_DEBUG
+#ifdef NTP_DEBUG
+#define ntp_debug(x, ...) do {tprintf(sym(ntp), 0, x, ##__VA_ARGS__);} while(0)
+#else
+#define ntp_debug(x, ...)
+#endif
+
 #define NTP_SERVER_DEFAULT  "pool.ntp.org"
 #define NTP_PORT_DEFAULT    123
 
@@ -13,9 +20,24 @@
 
 #define NTP_JIGGLE_THRESHOLD    30
 
-#define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
+#define NTP_RESET_THRESHOLD_MIN 1 /* seconds*/
+#define NTP_MAX_SLEW_LIMIT 500000ll /* ppm */
+#define NTP_MAX_FREQ_LIMIT 100000ll /* ppm */
 
-#define NTP_RESET_THRESHOLD_MIN 60  /* 60 seconds */
+#define PPM_SCALE(x) (((s64)(x)<<CLOCK_FP_BITS) / 1000000ll)
+#define fpmax (~(1ull<<63))
+
+#define DEFAULT_MAX_FREQ_PPM 25000
+#define DEFAULT_MAX_SLEW_PPM 83333
+#define MAX_BAD_REGRESSIONS 5
+#define MAX_SAMPLES 30
+#define MIN_SAMPLES 4
+#define get_sample(i) (ntp.samples + ((MAX_SAMPLES + (ntp.sample_index - ntp.nsamples) + i) % MAX_SAMPLES))
+
+#define MAX_DISPERSION seconds(16)
+#define MIN_SLEW_DURATION seconds(1)
+#define MAX_SLEW_DURATION seconds(10000)
+#define MIN_OFFSET_CORRECTION nanoseconds(1)
 
 struct ntp_ts {
     u32 seconds;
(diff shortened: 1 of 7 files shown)


src/runtime/extra_prints.c
```diff
@@ -240,7 +240,23 @@ static void print_timestamp(string b, timestamp t)
 static void format_timestamp(buffer dest, struct formatter_state *s, vlist *a)
 {
     timestamp t = varg(*a, timestamp);
-    print_timestamp(dest, t);
+    print_decimal(dest, t, 3);
```
francescolavra, line 243: This changes the resolution of printed timestamp values from microseconds to milliseconds. Is that intentional?

sanderssj: No, that should still be six decimal digits. I've changed it back, thanks!

klib/ntp.c
```diff
@@ -145,6 +313,218 @@ static void ntp_query(const ip_addr_t *server_addr, boolean lwip_locked)
     ntp_schedule_query();
 }
 
+/* Convert 32-bit ntp fixed point to nanos 64-bit fixed point */
+static s64 ntp_fixed_conv(u32 t)
+{
+    return ((u64)PP_NTOHL(t))<<16;
```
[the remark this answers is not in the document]
sanderssj: Thanks, that is more resilient way to write that.

klib/ntp.c
```diff
@@ -13,9 +20,24 @@
 
 #define NTP_JIGGLE_THRESHOLD    30
 
-#define NTP_MAX_SLEW_RATE   ((1ll << CLOCK_CALIBR_BITS) / 2000) /* 500 PPM */
+#define NTP_RESET_THRESHOLD_MIN 1 /* seconds*/
+#define NTP_MAX_SLEW_LIMIT 500000ll /* ppm */
+#define NTP_MAX_FREQ_LIMIT 100000ll /* ppm */
+
+#define PPM_SCALE(x) (((s64)(x)<<CLOCK_FP_BITS) / 1000000ll)
+#define fpmax (~(1ll<<63))
```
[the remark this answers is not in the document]
sanderssj: Thanks, I've updated it.

src/kernel/clock.c
```diff
@@ -51,6 +72,16 @@ closure_function(1, 1, boolean, timer_adjust_handler,
     return true;
 }
 
+void clock_rtc_step(s64 step)
+{
+    vdso_update_gen();
+    __vdso_dat->rtc_offset += step;
+    vdso_update_gen();
+    pqueue_walk(kernel_timers->pq, stack_closure(timer_adjust_handler, step));
```
wjhun, line 80: It seems to me that this pqueue_walk() - and the one in clock_reset_rtc() - should be called with the timerqueue lock held. Maybe a timer_walk() function could be added to timer.c that does this?

wjhun (approved):
A few comments, otherwise LGTM

klib/ntp.c
```diff
@@ -55,9 +96,23 @@ static struct {
...
+    struct ntp_ts last_originate_time;
+    int bad_regressions;
+    struct ntp_sample samples[MAX_SAMPLES];
+    int nsamples;
+    int sample_index;
+    timestamp slew_start;
+    s64 offset;                 /* negative means ahead of reference clock */
+    s64 base_freq;              /* negative means faster than reference clock */
+    s64 slew_freq;
+    s64 est_freq;
```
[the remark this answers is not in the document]
sanderssj: It was useful for debugging in-between response updates while I was working on it, but not really needed anymore. I'll take them out.

src/kernel/clock.c
sanderssj: Yes, that should be protected. I was thinking that when Francesco merges his PR for the timer changes I update this to use that solution since he's already solved it.

sanderssj: I updated clock_step_rtc() to use the new timer_adjust functions to protect the timer updates.

PR #1732, opened 2022-05-27
virtio-rng support

### LINKED ISSUE #1440: implement virtio-rng
Google Cloud supports virtio-rng, which looks quite straightforward to implement. This can be used to periodically call src/runtime:chacha20_randomstir() (or a variant that sources entropy from a buffer rather than random_seed()). (Using it as a direct source of random bytes may be tricky as the interface would need to be asynchronous - however /dev/urandom reads could be made blocking...TBD)

See section "5.4 Entropy Device" in the virtio-1.1 spec.

wjhun:
This introduces support for the virtio entropy device. When available, this
device will serve as the main source of entropy for the random number
generator. A hardware entropy source (e.g. rdseed and rdrand on x86-64) serves
as a fallback option when a virtio entropy device is not found.

Entropy data is stored in a double-buffered fashion with the intent to keep
entropy available for randomize functions without the need to block. Should an
underrun in entropy data occur, the aforementioned fallback option is used to
fulfill the request.

Other changes include adding a mutex to guard chacha20 usage and splitting
large getrandom(2) requests into smaller chunks to avoid starving other work.

Resolves #1440

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -80,6 +80,7 @@ SRCS-kernel.elf= \
 	$(SRCDIR)/virtio/virtio_mmio.c \
 	$(SRCDIR)/virtio/virtio_net.c \
 	$(SRCDIR)/virtio/virtio_pci.c \
+	$(SRCDIR)/virtio/virtio_rng.c \
 	$(SRCDIR)/virtio/virtio_storage.c \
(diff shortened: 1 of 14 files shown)


francescolavra (commented):
LGTM

PR #1733, opened 2022-06-02
clone(2): fix check for attempts to create new process

francescolavra:
The existing clone(2) syscall implementation was detecting whether the user program is trying to create a new process by checking the child_stack argument against a null value. This logic is based on the fact that the glibc fork() wrapper that is provided as part of the NPTL threading implementation invokes clone(2) with child_stack set to 0. However, it is possible to create a new process even if child_stack is non-zero: notably, the posix_spawn() implementation in glibc invokes clone() with a valid stack pointer, which is later unmapped by the parent process after the child starts executing the newly created process. The correct way for clone(2) to detect an attempt to create a new process is by checking for the CLONE_THREAD flag.


wjhun (approved):
LGTM

PR #1734, opened 2022-06-02
Reflect a volume's readonly status in /proc/mounts

sanderssj:
The readonly status added in https://github.com/nanovms/nanos/pull/1728 should be shown in /proc/mounts:
```
root / tfs ro 0 0
testvol /test tfs rw 0 0
testvolro /testro tfs ro 0 0
```


francescolavra (commented):
OK for me

wjhun (approved):
LGTM

PR #1736, opened 2022-06-06
fixing release build target

eyberg:


francescolavra:
I think the test klibs shouldn't be included in release tarballs.
I changed the Makefile to not include the test subfolder in the list of files copied into the klibs folder of a release tarball (the `bash -O extglob` part in the Makefile is necessary in order for the copy command to be run under bash, otherwise it errors out with "Syntax error: "(" unexpected").

eyberg:
makes sense - thanks

wjhun (approved):
LGTM

PR #1737, opened 2022-06-07
Azure: Handle multi-packet goalstate response from wireserver

sanderssj:
When the Azure cloud init issues a GET request for the goal state in order to
report status, the XML reponse from the wireserver may be large enough to
require multiple packets. This change fixes a bug where the response
was assumed to fit in a single packet, so the request parser was never
called unless the response was small enough. The http parser is moved
to the receive handler from the packet handler so that parser state is
preserved between packets. The response parser now closes the connection
when any complete http response is received.

### DIFF
--- a/klib/cloud_azure.c
+++ b/klib/cloud_azure.c
@@ -35,8 +35,8 @@ static void azure_report_retry(azure az)
         az->report_backoff <<= 1;
 }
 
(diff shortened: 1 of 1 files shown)


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1738, opened 2022-06-10
Syscall stat fixes

francescolavra:
This changeset fixes two issues found in the code that computes syscall execution statistics:
- the syscall_time field of the thread struct was not being initialized at syscall entry
- the count_syscall_save() function was not handling properly multiple consecutive calls without count_syscall_resume() being called in-between


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1740, opened 2022-06-11
Page cache: add global lock

francescolavra:
This lock protects from concurrent access to the list of page cache volumes.

### DIFF
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -85,6 +85,16 @@ static inline void pagelist_touch(pagelist pl, pagecache_page pp)
 }
 
 #ifdef KERNEL
+static inline void pagecache_lock(pagecache pc)
+{
+    spin_lock(&pc->global_lock);
+}
+
+static inline void pagecache_unlock(pagecache pc)
+{
+    spin_unlock(&pc->global_lock);
+}
+
 static inline void pagecache_lock_state(pagecache pc)
 {
     spin_lock(&pc->state_lock);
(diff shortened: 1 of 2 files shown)


sanderssj (commented):
LGTM

src/kernel/pagecache.c
```diff
@@ -1333,11 +1349,18 @@ u64 pagecache_get_node_length(pagecache_node pn)
...
+                 pagecache, pc,
                  rbnode, n)
 {
     pagecache_page pp = struct_from_list(n, pagecache_page, rbnode);
-    refcount_release(&pp->refcount);
+    if (!pp->evicted)
+        refcount_release(&pp->refcount);
+    pagecache pc = bound(pc);
+    pagecache_lock_state(pc);
+    pagelist_remove(&pc->free, pp);
```
wjhun, line 1361: Can we assert that pp is actually in the free state here? Could something else be possibly holding a reference to it at this point? (e.g. a dirty page has not completed writing)

Perhaps it would be safer to flag the page to be terminated (instead of being added to the free list) and move the pagecache_page deallocate to pagecache_page_free().

wjhun: > Moving the pagecache_page deallocate to pagecache_page_free() cannot be done with the current code, because if pagecache_page_free() is called as a result of a page eviction, the page is still referenced by its node

Right, that's what I meant by flagging the page - to set a boolean within the page indicating that it should not be added to the free list but rather deallocated altogether. Though I guess that doesn't catch pages for the node that have already been evicted and are sitting on the free list.

If, by definition, a pagecache node being released means no outstanding page references are possible, I think it's fine as-is (with an assert). Maybe a comment could elucidate this.

francescolavra: Adding an assert that pp is in the free state here revealed a missing refcount release after a write, and upon further inspection the code that handles refcounting was shown to be racy with regards to the pagecache drain functionality. I tried to fix these issues in a separate PR (#1754), and the changes to pagecache_page_release() are now in a commit in that PR.

PR #1742, opened 2022-06-14
create_thread (x86): properly initialize SS field in thread frame

francescolavra:
The SS field contains the segment selector (index of GDT entry and requested privilege level) for the stack segment. Its value for a thread frame is usually set by the interrupt handler when an interrupt is triggered during execution of a user thread. However, if a signal is dispatched to a thread before any such interrupts occur, the ucontext struct passed to the signal handler will contain a zero value (copied from the SS field in the thread frame) in the ss field, which is incorrect and causes the thread frame to be restored (when the signal handler returns) with an invalid SS value; this invalid value then triggers a general protection fault during frame_return.
The first commit fixes the above issue by properly initializing the SS value in a thread frame during thread creation.

The second commit fixes a potential security flaw that could cause the kernel to access a rogue cpuinfo struct.


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1743, opened 2022-06-16
Syscalls: add timer_settime and settimeofday

francescolavra:
These syscalls allow changing the system (wall clock) time.
In addition, a few fixes are being applied to the existing timer code:
- use the monotonic clock in all itimer types
- implement proper locking in order to safely change the real-time clock from multiple sources, such as syscalls and NTP, and safely access the timerqueue during a real-time clock change in SMP systems
- move resetting of RTC offset value in the VDSO structure so that timer reordering uses the updated offset
- properly adjust expiry time for absolute timers based on real-time clock

The TFD_TIMER_CANCEL_ON_SET option is now implemented.
A set of test cases is being added to the runtime tests to exercise the new syscalls and test the behavior of active timers when the real-time clock undergoes a discontinuous change.

### DIFF
--- a/src/kernel/clock.c
+++ b/src/kernel/clock.c
@@ -57,7 +57,7 @@ void clock_reset_rtc(timestamp wallclock_now)
                 __func__, now(CLOCK_ID_REALTIME), wallclock_now);
     timestamp n = now(CLOCK_ID_REALTIME);
     rtc_settimeofday(sec_from_timestamp(wallclock_now));
-    pqueue_walk(kernel_timers->pq, stack_closure(timer_adjust_handler, wallclock_now - n));
-    timer_reorder(kernel_timers);
+    timer_adjust_begin(kernel_timers);
     reset_clock_vdso_dat();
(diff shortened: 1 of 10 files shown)


src/kernel/clock.c
```diff
@@ -39,6 +39,8 @@ closure_function(1, 1, boolean, timer_adjust_handler,
                 void *, v)
 {
     timer t = v;
+    if (t->absolute)
```
wjhun, line 44: Shouldn't this be ```if (!t->absolute)```?

wjhun (approved):
One comment, otherwise looks good

src/kernel/clock.c
francescolavra: For a real-time-based timer, the expiry timestamp as calculated by timer_expiry() is (t->expiry - __vdso_dat->rtc_offset). When the system time is changed, __vdso_dat->rtc_offset is changed, so if t->expiry is adjusted by the same amount (and this is done for relative timers) the resulting expiry timestamp remains the same, whereas if t->expiry is not touched (for absolute timers) the expiry timestamp is affected by the change in the system time.

PR #1744, opened 2022-06-16
move scheduling of thunks to async_apply() and async_apply_bh()

wjhun:
This change creates the async_apply() and async_apply_bh() functions for
scheduling thunks to the runqueue and bhqueues, respectively. Previously,
thunks were directly enqueued to the scheduling queue, which proved to be
error-prone as return values were often unchecked, leading to queueing
failures being undetected. The selection between enqueue() vs
enqueue_irqsafe() variants is also risky, as a (non-irqsafe) enqueue()
competing with an enqueue from an interrupt handler could potentially lead to
a hang. The semantics for the two queues have been clarified such that
scheduling to the runqueue is not allowed from interrupt handling; scheduling
from interrupt level must occur only to the bhqueue. This interface also hides
the implementation details of the scheduling queues, thus making it easier to
revise these details without affecting much of the calling code.

All async_apply*() functions are now void returns, indicating that such
scheduling should never fail. Internally, the functions simply assert that
enqueues succeed, but in the future they could possibly lead to alternate
paths to mitigate the limit (such as suspending the execution context, if
possible, while queues are drained and resources are made available). It may
also be desirable to adopt a lock-free queue that is extendable, avoiding the
need for fixed queue sizes altogether.

Within the scheduler, the queue sizes set on initialization are now defined in
config.h, and the async_queue_1 size has been expanded to make room for large
bursts of enqueues, such as those that occur while servicing virtqueues.

### DIFF
--- a/src/kernel/tracelog.c
+++ b/src/kernel/tracelog.c
@@ -75,7 +75,7 @@ build_assert((TRACELOG_ENTRY_SIZE & (TRACELOG_ENTRY_SIZE - 1)) == 0);
 static inline void schedule_collator(void)
 {
     if (!atomic_swap_boolean(&tracelog.collator_scheduled, true))
-        enqueue_irqsafe(runqueue, &tracelog.collator);
+        async_apply_bh((thunk)&tracelog.collator);
 }
 
 static inline tracelog_buffer get_tracelog_buffer(cpuinfo ci)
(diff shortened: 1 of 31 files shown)


src/kernel/tracelog.c
```diff
@@ -75,7 +75,7 @@ build_assert((TRACELOG_ENTRY_SIZE & (TRACELOG_ENTRY_SIZE - 1)) == 0);
 static inline void schedule_collator(void)
 {
     if (!atomic_swap_boolean(&tracelog.collator_scheduled, true))
-        enqueue_irqsafe(runqueue, &tracelog.collator);
+        async_apply((thunk)&tracelog.collator);
```
francescolavra, line 78: This is called (indirectly) by tprintf(), which is called by the various debug prints when any of the `*_DEBUG` defines is enabled. Some of these debug prints are done in interrupt context (e.g. those in src/*/interrupt.c). So we should use async_apply_bh here.

PR #1746, opened 2022-06-22
Syscalls: add recvmmsg implementation

francescolavra:
This Linux-specific syscall allows receiving multiple messages from a socket. The timeout feature (which even in Linux does not work as intended) is not implemented.
In addition, the functionality of the existing sendmmsg syscall has been expanded to cover not only network sockets, but any type of socket that implements the sendmsg callback.
For network sockets, the code that handles data transmission and reception has been refactored to use iovec structures instead of a single contiguous buffer; this allows avoiding the additional buffer allocation and copying that was done for sendmsg, sendmmsg, and recvmsg.
For datagram sockets, the MSG_TRUNC option is now implemented (both as an input flag specified by the user process in a syscall, and as an output flags set by the kernel).

### DIFF
--- a/src/kernel/kernel.h
+++ b/src/kernel/kernel.h
@@ -583,6 +583,14 @@ static inline void clear_fault_handler(void)
         }                                                               \
     } while (0);
 
+
+#define contextual_closure_init(__name, __var, ...)                                 \
+    do {                                                                            \
+        context __ctx = get_current_context(current_cpu());                         \
+        __closure(u64_from_pointer(__ctx) | CLOSURE_COMMON_CTX_IS_CONTEXT, (__var), \
+                  sizeof(struct _closure_##__name), __name, ##__VA_ARGS__);         \
+    } while (0);
+
 static inline boolean is_contextual_closure(void *p)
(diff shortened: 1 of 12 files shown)


sanderssj (commented):
LGTM

src/kernel/kernel.h
```diff
@@ -567,6 +567,14 @@ static inline void clear_fault_handler(void)
         }                                                               \
     } while (0);
 
+
+#define contextual_closure_init(__name, __var, ...)                                 \
```
wjhun, line 571: This could be achieved via an init_closure() and a call to apply_context_to_closure(), as used in iov_op(). Though I think this looks nicer...maybe change iov_op() to use this so there aren't two ways to init a closure with context?

wjhun (approved):
LGTM

src/kernel/kernel.h
francescolavra: OK, I replaced the existing call to apply_context_to_closure() with the use of the new contextual_closure_init() macro.

PR #1747, opened 2022-06-23
vm_exit(): add idle_on_exit option

francescolavra:
This change introduces a new option, called "idle_on_exit", which when specified in the manifest keeps a VM running (with all VCPUs halted) after the user program exits successfully.
The implementation of the machine_halt() function for aarch64 and risc-v is being fixed so that it actually halts the CPU and does not return.
Generation of the broadcast shutdown IPI (which is only needed when the idle_on_exit option takes effect) is being removed from where it is not needed.
Handing of the reboot_on_exit option is being modified so that this option only takes effect if the exit code is not the expected one (i.e. if either the application terminates with error, or the kernel crashes), because this was the original intent behind this feature, and rebooting unconditionally makes it impossible to cleanly stop a running instance (e.g. with `ops instance stop`).

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -27,8 +27,6 @@
 #define BOOT_PARAM_OFFSET_CMDLINE_SIZE  0x0238
 #define BOOT_PARAM_OFFSET_E820_TABLE    0x02D0
 
-//#define SMP_DUMP_FRAME_RETURN_COUNT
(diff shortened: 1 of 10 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1748, opened 2022-06-23
Improve robustness of fs_full test

sanderssj:
This test was failing frequently with false test failures. The tfs log flush
may be delayed a second after the last write that fills the disk, and that
flush may free disk space if an old log gets removed. Write_blocks has
been refactored so that it will retry after at least twice the tfs log flush 
delay to ensure enough time for the log flush and completions to run
so that any eligible log space will be freed before giving up on the write.

The write test when the filesystem is full now allows writing up to bfree 
blocks instead of assuming that any write will fail. The file remove and 
free space check now syncs and waits after the remove is called to
allow time for the free space to be updated before calling stat.

### DIFF
--- a/test/runtime/fs_full.c
+++ b/test/runtime/fs_full.c
@@ -7,6 +7,7 @@
 #include <sys/vfs.h>
 #include <assert.h>
 #include <errno.h>
+#include <config.h>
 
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1749, opened 2022-06-24
Unix stdout: describe file descriptor as a pseudo-tty

francescolavra:
Describing stdout as a pseudo-tty makes glibc apply line buffering (instead of full buffering) when the process writes to stdout. This removes the need for calling `setbuf(stdout, NULL)` in the runtime test programs.
On x86, tested with glibc version 2.23, 2.24, 2.27, 2.28, 2.31, and 2.34; musl libc does not apply full buffering (tested with version 1.2.2), with or without this change.
On ARM, tested with glibc version 2.28.
On risc-v, tested with glibc version 2.32.


wjhun (approved):
LGTM

PR #1750, opened 2022-06-27
clone3: Add syscall

richiejp:
Introduce clone_internal similar to Linux's kernel_clone and
reimplement clone and implement clone3 in terms of it.

Most of Linux's clone3 functionality is irrelevant to Nanos. The main
thing we are concerned about is the stack. The stack size is now
passed to the kernel and the stack pointer points to the bottom of the
stack instead of the top.

It seems that at some point glibc started using clone3. This get's
programs linked against glibc-2.35 to run.

### DIFF
--- a/src/unix/thread.c
+++ b/src/unix/thread.c
@@ -51,6 +51,60 @@ sysreturn arch_prctl(int code, unsigned long addr)
 }
 #endif
 
+struct clone_args_internal {
+     u64 flags;
+     int *child_tid;
+     int *parent_tid;
+     void *stack;
+     bytes stack_size;
+     u64 tls;
+};
+
+sysreturn clone_internal(struct clone_args_internal *args, bytes size)
+{
+     u64 flags = args->flags;
+
+     if (!args->stack_size)
+          return set_syscall_error(current, EINVAL);
+
+     if (!(flags & CLONE_THREAD)) {
+          thread_log(current, "attempted to create new process, aborting.");
+          return set_syscall_error(current, ENOSYS);
+     }
+
+     if (!validate_user_memory(args->stack, args->stack_size, true))
+          return set_syscall_error(current, EFAULT);
+
+     if (((flags & CLONE_PARENT_SETTID) &&
+          !validate_user_memory(args->parent_tid, sizeof(u64), true)) ||
+         ((flags & CLONE_CHILD_CLEARTID) &&
+          !validate_user_memory(args->child_tid, sizeof(u64), true)))
+          return set_syscall_error(current, EFAULT);
+
+     thread t = create_thread(current->p, INVALID_PHYSICAL);
+     context_frame f = thread_frame(t);
+
+     clone_frame_pstate(f, thread_frame(current));
+     thread_clone_sigmask(t, current);
+
+     set_syscall_return(t, 0);
+     f[SYSCALL_FRAME_SP] = (u64)args->stack;
+     if (flags & CLONE_SETTLS)
+	  set_tls(f, args->tls);
+     if (flags & CLONE_PARENT_SETTID)
+	  *(args->parent_tid) = t->tid;
+     if (flags & CLONE_CHILD_SETTID)
+	  *(args->child_tid) = t->tid;
+     if (flags & CLONE_CHILD_CLEARTID)
+	  t->clear_tid = args->child_tid;
+     t->blocked_on = 0;
+     t->syscall = 0;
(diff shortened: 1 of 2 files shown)


src/unix/thread.c
```diff
@@ -51,6 +51,60 @@ sysreturn arch_prctl(int code, unsigned long addr)
...
+sysreturn clone_internal(struct clone_args_internal *args, bytes size)
+{
+     u64 flags = args->flags;
+
+     if (!args->stack_size)
+          return set_syscall_error(current, EINVAL);
+
+     if (!(flags & CLONE_THREAD)) {
+          thread_log(current, "attempted to create new process, aborting.");
+          return set_syscall_error(current, ENOSYS);
```
francescolavra, line 72: I know these calls to set_syscall_error() are in the current code, but they are actually unnecessary, because these are direct returns (i.e. the syscall handler returns immediately without blocking). I would replace `set_syscall_error(current, <errno>)` with `-<errno>`; e.g. the above line becomes `return -ENOSYS;`

src/unix/thread.c
```diff
@@ -51,6 +51,60 @@ sysreturn arch_prctl(int code, unsigned long addr)
...
+     bytes stack_size;
+     u64 tls;
+};
+
+sysreturn clone_internal(struct clone_args_internal *args, bytes size)
+{
+     u64 flags = args->flags;
+
+     if (!args->stack_size)
+          return set_syscall_error(current, EINVAL);
```
francescolavra, line 68: return -EINVAL;

src/unix/thread.c
```diff
@@ -60,51 +114,59 @@ sysreturn clone(unsigned long flags, void *child_stack, int *ptid, unsigned long
...
+
+sysreturn clone3(struct clone_args *args, bytes size)
+{
+     thread_log(current,
+         "clone3: args_size: %ld, pidfd: %p, child_tid: %p, parent_tid: %p, exit_signal: %ld, stack: %p, stack_size: 0x%lx, tls: %p",
+         size, args->pidfd, args->child_tid, args->parent_tid, args->exit_signal,
+         args->stack, args->stack_size, args->tls);
+
+     if (size < sizeof(*args))
+          return set_syscall_error(current, EINVAL);
```
francescolavra, line 149: return -EINVAL;

src/unix/thread.c
```diff
@@ -51,6 +51,60 @@ sysreturn arch_prctl(int code, unsigned long addr)
...
+struct clone_args_internal {
+     u64 flags;
+     int *child_tid;
+     int *parent_tid;
+     void *stack;
+     bytes stack_size;
+     u64 tls;
+};
+
+sysreturn clone_internal(struct clone_args_internal *args, bytes size)
```
francescolavra, line 63: Should be static

src/unix/thread.c
```diff
@@ -51,6 +51,60 @@ sysreturn arch_prctl(int code, unsigned long addr)
...
+
+     if (!args->stack_size)
+          return set_syscall_error(current, EINVAL);
+
+     if (!(flags & CLONE_THREAD)) {
+          thread_log(current, "attempted to create new process, aborting.");
+          return set_syscall_error(current, ENOSYS);
+     }
+
+     if (!validate_user_memory(args->stack, args->stack_size, true))
```
francescolavra, line 75: Since args->stack points to the top of the stack, this validation should be done from (args->stack - args->stack_size) to args->stack

francescolavra (changes requested):

src/unix/thread.c
richiejp: I switched this around again so that clone_internal expects the bottom of the stack. Thanks!

francescolavra (approved):

richiejp:
I pushed again to add validation to the args pointer passed by the user.

src/unix/thread.c
```diff
@@ -60,51 +116,62 @@ sysreturn clone(unsigned long flags, void *child_stack, int *ptid, unsigned long
...
+         /* no stack size given, just validate the top word */
+         .stack = child_stack - sizeof(u64),
+         .stack_size = sizeof(u64),
+         .tls = newtls,
+    };
+
+    return clone_internal(&args);
 }
 
+struct clone_args {
```
wjhun, line 132: Structs defined by the syscall interface typically go in ```src/unix/system_structs.h```, if arch-independent, or ```src/<arch>/unix_machine.h``` otherwise.

PR #1752, opened 2022-06-28
Improve pagecache behavior in low memory situations

sanderssj:
This PR makes a number of changes related to the pagecache and writing files in a low memory
situation. Currently the user program can totally exhaust memory preventing kernel allocations from
occurring and thus preventing a safe and clean shutdown of nanos. One easy way to do that is for
the user program to perform large file writes in order to fill up the pagecache. The pagecache also
currently has a number of bugs causing it to leak memory or fail to free eligible pages with its
drain function.

A new reserve heap wrapper has been added that presents a view of the parent heap where the memory
is reduced by some amount specified at wrapper creation. This reserves that amount of memory at the parent
so the consumer of the wrapper heap cannot exhaust the parent heap. Allocations that go below the reserved
amount will fail and the total reported memory is also appropriately adjusted. This wrapper is applied to the
user program's anonymous page heap as well as the pagecache so neither can exhaust critical kernel memory.

Currently, when committing a dirty node, the pagecache will create a write transaction for an entire contiguous
dirty range of a node. If this range is large enough, the transaction will run the OS out of memory by either
creating an extremely large sg_list or by requiring the storage driver to use more memory than is available.
This PR modifies the committing process to be performed in chunks so that memory usage in both instances
is limited. Relatedly, sg_lists, when freed, are put a free list and not released to the heap. This has been
modified so that if an sg_list contains an extremely large buffer, that buffer is reallocated to a specified
threshold and the extra memory released back to the heap.

Other changes include releasing sg_lists when handling errors, adding a missing refcount_release related
to pagecache_write_sg and making sure that all pagecache pages required for a write are either in a
READING or WRITING state so that they won't be drained.


src/runtime/heap/reserve.c
```diff
@@ -0,0 +1,111 @@
...
+    rl->mgmt = (tuple)n;
+    return n;
+}
+
+heap reserve_heap_wrapper(heap meta, heap parent, bytes reserved)
+{
+    reservelock rl = allocate(meta, sizeof(*rl));
+    if (rl == INVALID_ADDRESS)
+        return INVALID_ADDRESS;
+    assert(parent->total && parent->allocated && parent->allocated(parent) > reserved);
```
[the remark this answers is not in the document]
sanderssj: Thanks, this should be heap_free()!

src/kernel/pagecache.c
```diff
@@ -542,8 +545,18 @@ closure_function(6, 1, void, pagecache_write_sg_finish,
...
     pagecache_debug("%s: pn %p, q %R, sg %p, status %v\n", __func__, pn, q, sg, s);
-    if (!is_ok(s))
+    if (!is_ok(s)) {
+        if (sg)
+            sg_list_release(sg);
+        pagecache_lock_state(pc);
+        for (int i = bound(pi); i < end; i++)  {
+            pagecache_page pp = page_lookup_nodelocked(pn, i);
+            if (pp != INVALID_ADDRESS)
+                pagecache_page_release_locked(pc, pp);
```
francescolavra, line 555: Here the node is actually not locked, so the page lookup is not safe. Since all relevant pages are guaranteed to be in the node's rbtree (pagecache_write_sg inserts them if not found), we can avoid a lookup for each page, and retrieve the next page via rbnode_get_next(), as done in the loop below. Looks like the check for (!is_ok(s)) and the cleanup/release stuff can be inserted just below the existing call to page_lookup_nodelocked().

sanderssj: Thanks, I fixed that.

wjhun (approved):
LGTM

PR #1754, opened 2022-07-01
Pagecache: fix page refcounting and races with page eviction

francescolavra:
The first commit fixes handling of concurrent page writes and reads.
The second commit makes page refcounting race-free with regard to the pagecache drain functionality, fixes a missing refcount release after a write completion, and fixes a 'refcount_release: invalid count 0' error occurring when a node with evicted pages is deallocated.
The third commit fixes a memory leak occurring when pages belonging to a deallocated node are released, and properly removes these pages from the global free page list.

### DIFF
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -232,13 +232,12 @@ static void enqueue_page_completion_statelocked(pagecache pc, pagecache_page pp,
 
 static boolean realloc_pagelocked(pagecache pc, pagecache_page pp)
 {
-    pagecache_debug("%s: pc %p pp %p refcount %d state %d\n", __func__, pc, pp, pp->refcount.c, page_state(pp));
+    pagecache_debug("%s: pc %p pp %p refcount %d state %d\n", __func__, pc, pp, pp->refcount, page_state(pp));
     pp->kvirt = allocate(pc->contiguous, U64_FROM_BIT(pc->page_order));
     if (pp->kvirt == INVALID_ADDRESS) {
         return false;
     }
-    assert(pp->refcount.c == 0);
-    refcount_reserve(&pp->refcount);
+    assert(pp->refcount++ == 0);
     pp->write_count = 0;
     #ifdef KERNEL
     pp->phys = physical_from_virtual(pp->kvirt);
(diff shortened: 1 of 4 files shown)


francescolavra:
In pagecache_drain(), I simplified the logic by removing the loop that evicts at most DRAIN_ITER_MAX pages at a time (since we don't have a temporary evict list anymore, we don't need to evict pages in multiple iterations).

src/kernel/pagecache.c
```diff
@@ -533,41 +546,56 @@ closure_function(5, 1, void, pagecache_write_sg_finish,
...
             zero(pp->kvirt + offset, copy_len);
         }
-        pagecache_lock_state(pc);
-        assert(page_state(pp) != PAGECACHE_PAGESTATE_READING);
         if (page_state(pp) != PAGECACHE_PAGESTATE_DIRTY)
+            /* Don't release the page, otherwise if it has been marked for eviction it will be
+             * evicted while still dirty. */
             change_page_state_locked(pc, pp, PAGECACHE_PAGESTATE_DIRTY);
+        else
+            pagecache_page_release_locked(pc, pp);
```
sanderssj, line 584: There is still a refcount problem relating to write_sg I think. If I merge this branch into the low memory branch, I can cause an assertion failure running the fs_full test with lower than normal memory (e.g. 384M) where the page state of a page when going to refcount 0 is PAGECACHE_PAGESTATE_NEW instead of FREE. I can make it pass by unconditionally calling pagecache_page_release_locked and not changing the state to DIRTY, but that's not correct for the page state.

sanderssj (commented):
I think there is still something off with write_sg/write_sg_finish refcount handling. I am still trying to dig into it with this branch added to the low memory branch.

src/kernel/pagecache.c
francescolavra: I could reproduce the issue after merging the low memory branch into this branch. Assuming you resolved the merge conflicts in the same way as I did, the issue is caused by the code in pagecache_write_sg() that changes the page state to WRITING (at line 696 in your branch); removing that code fixes the issue.

sanderssj: You're right -- I forgot to remove those lines after the merge, and it does also make it work for me. Looks good, thanks!

src/runtime/sg.c
```diff
@@ -143,6 +143,26 @@ u64 sg_copy_to_buf_and_release(void *target, sg_list sg, u64 n)
...
+    sg_buf sgb;
+    u64 sgb_index = 0;
+    while (n > 0 && (sgb = sg_list_peek_at(sg, sgb_index)) != INVALID_ADDRESS) {
+        u64 len = MIN(n, sg_buf_len(sgb));
+        n -= len;
+        volatile u8 *bp = sgb->buf + sgb->offset;
+        while (len > 0) {
+            (void)*bp;
+            bp += PAGESIZE;
+            if (len <= PAGESIZE)
```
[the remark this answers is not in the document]
francescolavra: Right, now I made len an s64 and removed the conditional break

PR #1755, opened 2022-07-01
NTP: add ability to configure multiple servers

francescolavra:
With this change, it is possible to configure the NTP klib with multiple NTP servers, by inserting an `ntp_servers` array in the root tuple of the manifest, where each array element follows the syntax `<addr>[:<port]`; `<addr>` can be an IP address or a fully qualified domain name; if the optional `<port>` is not present, the default NTP port 123 is used.
Example configuration with 2 servers: `ntp_servers:[us.pool.ntp.org 12.34.56.78:1234]`
When multiple servers are configured, the NTP klib starts by using the first server, and switches to the next server (in a round robin fashion) if no valid response is received from the current server.

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -71,6 +71,11 @@ struct ntp_sample {
     u64 peer_dispersion;
 };
 
+typedef struct ntp_server {
+    char *addr;
+    u16 port;
+} *ntp_server;
+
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

PR #1756, opened 2022-07-13
Klibs: ntp: add IPv6 support and improve handling of multiple NTP servers

francescolavra:
The first commit allows the NTP klib to work on VMs with IPv6 internet connectivity.
The second commit changes the behavior of the klib so that if multiple NTP servers are configured, all servers are probed in parallel: this prevents non-reachable or invalid servers from delaying synchronization of the VM system time if there is at least one valid and reachable server.

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -717,21 +717,41 @@ static void ntp_server_add(heap h, buffer addr, u16 port)
 
 static boolean ntp_server_parse(heap h, buffer server)
 {
-    int separator = buffer_strchr(server, ':');
-    if ((separator == 0) || (separator == buffer_length(server) - 1))
+    if (buffer_length(server) == 0)
         return false;
-    buffer addr;
+    buffer addr = 0;
     u16 port;
+    int separator;
+    if (byte(server, 0) == '[') {   /* numeric IPv6 address */
+        int addr_end = buffer_strchr(server, ']');
+        if (addr_end < 0)
+            return false;
+        if (addr_end == buffer_length(server) - 1) {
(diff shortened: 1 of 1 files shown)


klib/ntp.c
```diff
@@ -677,17 +724,19 @@ define_closure_function(0, 2, void, ntp_query_func,
...
-    lwip_lock();
-    ntp_server server = vector_get(ntp.servers, ntp.current_server);
-    err_t err = dns_gethostbyname(server->addr, &server_addr, ntp_dns_cb, 0);
-    lwip_unlock();
-    if (err == ERR_OK)
-        ntp_query(&server_addr, false);
-    else if (err != ERR_INPROGRESS) {
-        rprintf("%s: failed to resolve hostname: %d\n", __func__, err);
-        ntp_query_complete(false);
+    int current_server = ntp.current_server;
```
sanderssj, line 727: I think you need an ntp_lock() around this section for the timer handler.

sanderssj (commented):
LGTM other than a missing lock.

klib/ntp.c
francescolavra: It seems to me a lock would only be needed if we weren't using a local variable (`current_server`) and were using `ntp.current_server` directly. But since the `ntp.current_server` value here is only read once when initializing the local variable (and the `ntp.servers` vector is never modified after init), it seems we can do safely without the lock.

sanderssj: That makes sense. Works for me.

wjhun (approved):
LGTM

PR #1758, opened 2022-07-20
TFS log: do not use the page cache

francescolavra:
Using the kernel page cache as an intermediary between the TFS log code and the storage device has a number of disadvantages:
- log extensions read during system startup are copied twice in memory (first in the page cache and then in the TFS log buffers), and one of the two copies (the one in the page cache) is never used afterwards
- log extensions written (by the log_flush function) after a filesystem modification are not immediately flushed to disk, which is inconsistent with the semantics of the fsync() syscall
- each write of a log extension chunk that begins at a page boundary and does not span an entire page triggers an unnecessary read from disk in the page cache code
- when the TFS log is rebuilt, the storage space occupied by the extensions of the log being retired is freed while there might still be dirty pages associated to these extensions, which can cause filesytem corruption if this storage space is reused before dirty pages are flushed to disk

This PR changes the TFS log code to bypass the page cache and use the filesystem storage interface directly.

### DIFF
--- a/src/tfs/tlog.c
+++ b/src/tfs/tlog.c
@@ -54,9 +54,9 @@ typedef struct log_ext *log_ext;
 
 #endif
 
-declare_closure_struct(1, 1, void, log_ext_sync_complete,
-                       log_ext, ext,
-                       status, s);
+declare_closure_struct(3, 3, void, log_storage_op,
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1759, opened 2022-07-27
Scheduler: implement fair scheduling between user threads

francescolavra:
This change introduces a new scheduling algorithm that assigns CPU resources to a task based on how much CPU time the task used in the past. It uses a priority queue where tasks are ordered based on their CPU time: when a task is enqueued, it is positioned according to the time it has been running since it was last scheduled; when a task is dequeued, its CPU time is subtracted from the CPU time of any other tasks in the queue, so that their priority in the queue is increased; this is implemented by storing a reference CPU time in the queue (min_runtime struct member), which prevents long-running tasks from being starved by newly created tasks.
This algorithm is applied for scheduling user threads, so that CPU time is distributed fairly between all runnable threads.

The first commit is a fix to a pre-existing bug that has been exposed by modifying the execution order of runnable user threads.

### DIFF
--- a/src/unix/thread.c
+++ b/src/unix/thread.c
@@ -209,16 +209,23 @@ static inline void check_stop_conditions(thread t)
     halt("Terminating.\n");
 }
 
+static void thread_cputime_update(thread t)
+{
+    if (t->start_time != 0) {
+        timestamp diff = now(CLOCK_ID_MONOTONIC_RAW) - t->start_time;
+        t->utime += diff;
+        t->task.runtime = diff;
+        t->start_time = 0;
+    }
+}
+
 static void thread_pause(context ctx)
 {
     if (shutting_down)
         return;
     thread t = (thread)ctx;
     context_frame f = thread_frame(t);
-    assert(t->start_time != 0); // XXX tmp debug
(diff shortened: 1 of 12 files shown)


src/unix/thread.c
```diff
@@ -240,7 +247,8 @@ static void thread_resume(context ctx)
 static void thread_schedule_return(context ctx)
 {
     thread t = (thread)ctx;
-    assert(enqueue_irqsafe(t->scheduling_queue, &t->thread_return));
+    thread_cputime_update(t);   /* so that it is scheduled based on how much CPU time it used */
```
wjhun, line 250: Wouldn't the thread be paused at this point, with t->start_time == 0, thus making the call to thread_cputime_update() a no-op here?

wjhun (approved):
LGTM

src/unix/thread.c
francescolavra: When a user thread is interrupted by a hardware interrupt, the interrupt handler calls context_schedule_return() without the thread being paused. Also, when the Unix fault handler raises a signal or resolves a page fault without blocking, it calls schedule_thread(), again without the thread being paused.

PR #1760, opened 2022-07-28
pagecache: fix committing of nodes with empty `dirty` rangemap

francescolavra:
pagecache_commit_dirty_node() copies any dirty ranges in a node to a buffer and removes them from the `dirty` rangemap; then, it triggers the actual committing of the dirty ranges. If pagecache_commit_dirty_node() is called when the rangemap is empty, the completion closure is invoked immediately even if the committing is ongoing, which is incorrect and can cause issues such as a premature deallocation of fsfiles that have not been synchronized to disk.
In addition, the buffer is not properly deallocated when the rangemap is empty.
This change fixes the above issues.


wjhun (approved):
LGTM

PR #1761, opened 2022-08-01
virtio-net performance improvements

francescolavra:
This change set improves performance of network transfers over the virtio-net network interface. As an example, the Warp (https://github.com/minio/warp) mixed benchamrk (run with `./warp mixed --autoterm --objects 40 --duration 1m`) against a minIO server showed a change in results from `Cluster Total: 59.45 MiB/s, 9.88 obj/s over 53s` to `Cluster Total: 83.11 MiB/s, 13.88 obj/s over 50s`.
The first commit increases the TCP send buffer size and TCP window size (since lwIP implements TCP window scaling, the window size is not limited to a maximum of 65535).
The second commit adds support for the VIRTIO_F_RING_EVENT_IDX feature, which allows a virtIO driver to indicate to the device when it wants to be notified of an update to used buffers, and similarly allows a virtIO device to indicate to the driver when it wants to be notified of an update to available buffers. By avoiding unneeded device interrupts and driver notifications, the number of VM exits and hypervisor syscalls (e.g. to execute KVM_SIGNAL_MSI ioctls) is minimized, which decreases CPU load on the host.
The third commit implements polling mode in the virtIO code and enables it in the TX queue of the virtio_net driver. When a virtIO queue is in polling mode, the driver does not want to be notified (via an interrupt) when a virtIO buffer in the queue has been used by the device. This has been shown to improve the speed of TCP transfers: for example, the average performance of a Go web server as measured by the Jenkins CI job increases by ~12%.

### DIFF
--- a/src/virtio/virtqueue.c
+++ b/src/virtio/virtqueue.c
@@ -92,11 +92,11 @@ typedef struct virtqueue {
     u16 queue_index;
     bytes notify_offset;
     void *ring_mem;
-    bytes avail_offset;
-    bytes used_offset;
     volatile struct vring_desc *desc;
     volatile struct vring_avail *avail;
     volatile struct vring_used *used;    
+    u16 *avail_event;
+    u16 *used_event;
     u64 free_cnt;               /* atomic */
     u16 desc_idx;               /* head of descriptor free list */
     u16 last_used_idx;          /* irq only */
@@ -163,16 +163,11 @@ void vqmsg_commit(virtqueue vq, vqmsg m, vqfinish completion)
     spin_unlock_irq(&vq->lock, irqflags);
 }
 
-closure_function(1, 0, void, vq_interrupt,
-                 virtqueue, vq)
+static void vq_poll(virtqueue vq)
 {
     // ensure we see up-to-date used->idx (updated by host)
     memory_barrier();
-    virtqueue vq = bound(vq);
-    virtqueue_debug_verbose("%s: ENTRY: vq %s: entries %d, last_used_idx %d, used->idx %d, desc_idx %d\n",
-        __func__, vq->name, vq->entries, vq->last_used_idx, vq->used->idx, vq->desc_idx);
     
-    spin_lock(&vq->lock);
     while (vq->last_used_idx != vq->used->idx) {
         volatile struct vring_used_elem *uep = vq->used->ring + (vq->last_used_idx & (vq->entries - 1));
         virtqueue_debug_verbose("%s: vq %s: last_used_idx %d, id %d, len %d\n",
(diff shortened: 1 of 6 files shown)


src/virtio/virtqueue.c
```diff
@@ -219,9 +233,11 @@ status virtqueue_alloc(vtdev dev,
 {
     u64 vq_alloc_size = sizeof(struct virtqueue) + size * sizeof(vqmsg);
     virtqueue vq = allocate_zero(dev->general, vq_alloc_size);
-    vq->avail_offset = size * sizeof(struct vring_desc);
-    vq->used_offset = pad(vq->avail_offset + sizeof(*vq->avail) + sizeof(vq->avail->ring[0]) * size, align);
-    bytes alloc = vq->used_offset + pad(sizeof(*vq->used) + sizeof(vq->used->ring[0]) * size, align);
+    bytes avail_offset = size * sizeof(struct vring_desc);
+    bytes used_offset = pad(avail_offset + sizeof(*vq->avail) + sizeof(vq->avail->ring[0]) * size +
+                            sizeof(u16) /* used_event */, align);
```
sanderssj, line 240: Is the extra 2 bytes for the used_event still appropriate here (and again below) if VIRTIO_F_RING_EVENT_IDX has not been negotiated with the device?

sanderssj (commented):
LGTM, just had a question about the offset calculation change.

src/virtio/virtqueue.c
francescolavra: While those extra bytes are unused if VIRTIO_F_RING_EVENT_IDX has not been negotiated, they only affect the offset and allocation size of the virtqueue available and used rings. For non-legacy devices, the offsets don't really matter, since the location of the rings is passed to device during configuration (e..g. via VTPCI_R_QUEUE_AVAIL and VTPCI_R_QUEUE_USED for PCI devices). For legacy devices, that require a fixed virtqueue layout, the offset do matter, but as far as I can see the calculation of the offsets was wrong before (and is correct with these changes): the virtio spec (version 1.1), at section 2.6.2 "Legacy Interfaces: A Note on Virtqueue Layout", says that the total virtqueue size is calculated as `ALIGN(sizeof(struct virtq_desc)*qsz + sizeof(u16)*(3 + qsz)) + ALIGN(sizeof(u16)*3 + sizeof(struct virtq_used_elem)*qsz)`. The `3 * sizeof(u16)` part does not match the size of struct vring_avail and struct vring_used in the kernel. By adding another 2 bytes we match the formula. In practice, in the majority of cases this doesn't make a difference, due to the padding (necessary for alignment) inserted between the available and used rings. I guess we never came across a legacy device for which those 2 bytes would change the actual offset between available and used rings, so we never noticed the issue before.

sanderssj: Thanks, great explanation!

wjhun (approved):
LGTM

PR #1762, opened 2022-08-03
Syscalls: improve support for file data and metadata synchronization to the storage device

francescolavra:
The first commit changes the implementation of fsync and fdatasync to avoid flushing the TFS log when not needed. The filesystem_sync() function has been removed in favor of using filesystem_flush(), because there was duplicated code between these functions, and each filesystem_sync() call was triggering two calls to pagecache_sync_volume(). Similarly, filesystem_sync_node() (which was triggering an unnecessary call to pagecache_sync_volume) is being replaced by fsfile_flush(). In order to ensure that the completion of an fync/fdatasync syscall runs in the appropriate syscall context, calls to apply() have been replaced by calls to async_apply_status_handler() in the TFS code and the storage drivers that implement the flush operation.
The second commit adds support for the O_SYNC and O_DSYNC open(2) flags. When a regular file is opened with O_SYNC, by the time write(2) (or similar) returns, the output data and associated file metadata have been transferred to the underlying hardware. The O_DSYNC flag is similar, but it only flushes file metadata if they are needed to correctly handle a subsequent data retrieval.

### DIFF
--- a/src/kernel/storage.c
+++ b/src/kernel/storage.c
@@ -2,7 +2,6 @@
 #include <pagecache.h>
 #include <storage.h>
 #include <tfs.h>
-#include <unix.h>
 
 //#define STORAGE_DEBUG
 #ifdef STORAGE_DEBUG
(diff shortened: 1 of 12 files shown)


francescolavra:
I removed the context_apply_status_handler() macro and instead replaced calls to apply() with calls to async_apply_status_handler() in the TFS code and the storage drivers that implement the flush operation, in order to ensure that the completion of an fync/fdatasync/write syscall runs in the appropriate syscall context.

wjhun (approved):
[Edit: I just realized the async applies are only for flush completions. LGTM]

PR #1763, opened 2022-08-06
virtio_scsi: honor maximum transfer length limit

francescolavra:
A virtio SCSI disk may have a limit on the number of blocks that can be requested in a single I/O transfer; the value of this limit can be found in the block limits Vital Product Data (VPD) page.
This change adds retrieval of the maximum transfer length for a SCSI disk when the disk is probed, and ensures that this limit is not exceeded when executing STORAGE_OP_READSG and STORAGE_OP_WRITESG requests.
This fixes booting in GCP, where the boot disk has a maximum transfer length of 512 blocks (256 KB), which is less than the size of a TFS log extension.

### DIFF
--- a/src/virtio/scsi.h
+++ b/src/virtio/scsi.h
@@ -243,6 +243,7 @@ struct scsi_res_inquiry
 
 /* Vital Product Data page codes */
 #define SCSI_VPD_DEVID  0x83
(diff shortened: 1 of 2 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1765, opened 2022-08-09
Disk I/O optimization

francescolavra:
This change set aims at minimizing the number of I/O requests and scatter-gather buffers used to transfer a given amount of data to/from disk, which helps parallelizing I/O (especially when the I/O ring queue depth is limited) and minimizes the memory allocation needs for an I/O transfer.
The first commit implements merging of contiguous pagecache pages that have their backing memory areas adjacent (which happens frequently if the cache for a node is filled sequentially starting from offset 0) into a single scatter-gather buffer before doing a filesystem storage operation (both when reading and writing a page cache node).
The second commit implements extension of existing file extents to cover additional storage space beyond the initially allocated space, and removes the 1MB maximum extent size; this helps keeping the amount of file metadata down and minimizing fragmentation of file data into non-contiguous storage areas. Since allocations from the storage space are aligned to the allocation size, requesting large extents can create large unallocated ranges in the storage space; in order to be able to fill these ranges when requesting a new extent that does not fit into a single contiguous storage area, the create_extent() function upon allocation failure retries an allocation with a smaller size (down to a 1MB limit); the code that calls this function has been amended to properly handle the cases where the size of a created extent is smaller than requested.
https://github.com/nanovms/nanos/issues/1165

### DIFF
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -250,21 +250,15 @@ static boolean realloc_pagelocked(pagecache pc, pagecache_page pp)
     return true;
 }
 
-/* If the refcount argument is true, this function must be called with the page state lock held. */
-static void pagecache_add_sgb(pagecache pc, pagecache_page pp, sg_list sg, boolean refcount)
+static sg_buf pagecache_add_sgb(pagecache pc, pagecache_page pp, sg_list sg)
 {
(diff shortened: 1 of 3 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1766, opened 2022-08-11
Add devicetree blob parsing support for RISC-V and ARM

sanderssj:
This change adds basic support for parsing and reading from the devicetree
blob (DTB) found in RISC-V and ARM machines. The DTB contains information
about a machine and its devices and how to access or configure them. Parsing
of the DTB does not require any special kernel features, so the devicetree
code is located in its own directory so it can be used by the kernel or by
user tools.

The ranges of memory in a machine are specified only in the DTB. Nanos has
previously been assuming a fixed size memory for RISC-V and ARM VMs (on QEMU), so
this change now allows Nanos to discover the actual amount of memory on a machine.

The devicetree_dump() function will dump the contents of the DTB, given its
address.

In order to accommodate the limited boot environment in the kernel when the
DTB will normally be accessed, the code uses only stack memory, no heap.
Any extensive processing of the DTB should cache the tree structure to
make navigation more efficient. However, that is not included in this change.

On RISC-V, the devicetree blob is relocated from the end of physical memory to the area
used by init and mapped into kernel virtual address space. On ARM, the blob is left alone
since it is placed at the beginning of physical memory.


src/devicetree/devicetree.c
```diff
@@ -0,0 +1,519 @@
...
+ *
+ * All the calls that begin with dtb will directly parse the devicetree blob
+ * and do not use any heap memory. This is obviously very inefficient but
+ * necessary since some devicetree parsing is required before kernel memory
+ * is set up and mapped.
+ */
+
+#define DTB_DEBUG
+#ifdef DTB_DEBUG
+#define dtb_debug(x)    console(x)
```
wjhun, line 26: These appear to be unused.

platform/riscv-virt/kernel_platform.h
```diff
@@ -1,8 +1,11 @@
 #define PHYSMEM_BASE          0x80000000
-#define BIOS_SIZE             (128*1024)
-#define INIT_PAGEMEM          (PHYSMEM_BASE+BIOS_SIZE)
+#define BIOS_SIZE             (128*KB)
```
wjhun, line 2: I have seen some newer OpenSBI images exceed 128KB. Not sure what a safe size is - maybe double this for now?

wjhun (approved):
LGTM

platform/riscv-virt/kernel_platform.h
sanderssj: Good idea, seems safer. I've updated it to 256K.

PR #1769, opened 2022-08-23
Tracing: add support for selectively enabling trace message types

francescolavra:
This change set adds support for specifying trace flags (i.e. a comma-delimited set of trace message types) in the "trace" symbol of the root tuple. A given trace message is output only if its message type is enabled in the trace flags. For backward
compatibility, any unknown trace flag results in the TRACE_OTHER bit being set, i.e. it enables all messages output via calls to thread_log(). Page-fault-related messages and messages output when returning to user threads are now classified in their own categories ("pf" and "threadrun", respectively), i.e. they need to be explicitly enabled in the trace flags.
If "all" is specified in the trace flags, all tracing messages are enabled.

Example: a "trace:pf,other" value in the manifest enables page-fault-related messages and messages classified as "other".

### DIFF
--- a/klib/radar.c
+++ b/klib/radar.c
@@ -1,6 +1,5 @@
 #include <kernel.h>
 #include <http.h>
-#include <log.h>
 #include <lwip.h>
 #include <pagecache.h>
 #include <storage.h>
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -3,7 +3,6 @@
 #include <pagecache.h>
 #include <storage.h>
 #include <tfs.h>
-#include <log.h>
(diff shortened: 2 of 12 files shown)


wjhun:
Would it be much work to specify comma-delimited flags by name in the manifest value, perhaps similarly to the way trace options are selected for qemu '-d' and the nanos DEBUG build variable? Fixed flag values may be difficult to recall.

francescolavra:
I implemented parsing of a comma-delimited set of flags. The special flag "all" enables all tracing messages

wjhun (approved):
LGTM

PR #1770, opened 2022-09-01
Hyper-V: storvsc_io(): honor maximum segment count for I/O requests

francescolavra:
A storvsc disk has a limit on how many segments (i.e. pages) can be transferred in a given I/O request. This PR changes storvsc_io() so that requests coming from the upper layers exceeding that limit are split into multiple I/O requests. This fixes booting on Azure instances.


wjhun (approved):
LGTM

PR #1771, opened 2022-09-08
tracelog file writer

wjhun:
This adds an option to the tracelog facility for recording trace entries to a file on the root partition. If a 'file' attribute is found in the tracelog manifest tuple, a file with the name of the corresponding value is created in the root directory and file tracing is enabled. For ease of debugging, this attribute may be added to the manifest at build time by using the TRACELOG_FILE build option, e.g.:

$ make TARGET=hw TRACE=tracelog DEBUG=sched TRACELOG_FILE=foo run

The trace data from execution will be left in the file named 'foo' in the root disk image. The tracelog file may be copied out of the image using the 'copylog' make target, which will leave a copy of the file in the output directory.

Note that enabling the file writer will start a periodic timer to collate and flush trace data to disk. This is to ensure that pending trace data is written in a timely fashion in case the system is unable to cleanly shut down. A shutdown completion is registered to flush pending data prior to termination.

The documentation in 'doc/tracelog.txt' has been updated accordingly, and mkfs has been updated to allow multiple instances of the '-t' commandline option.

### DIFF
--- a/src/kernel/tracelog.c
+++ b/src/kernel/tracelog.c
@@ -1,6 +1,8 @@
 #include <kernel.h>
 #include <http.h>
 #include <net.h>
+#include <pagecache.h>
+#include <tfs.h>
 
 //#define TRACELOG_DEBUG
 #ifdef TRACELOG_DEBUG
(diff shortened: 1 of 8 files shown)


src/kernel/tracelog.c
```diff
@@ -307,6 +336,78 @@ static void release_tracelog_buffer_locked(tracelog_buffer tb)
     refcount_release(&tb->refcount);
 }
 
+closure_function(5, 1, void, tracelog_file_write_complete,
+                 sg_list, sg, sg_buf, sbg, buffer, b, status_handler, complete, boolean, flushing,
```
francescolavra, line 340: Nit: the sg_buf member is not needed

src/kernel/tracelog.c
```diff
@@ -307,6 +336,78 @@ static void release_tracelog_buffer_locked(tracelog_buffer tb)
...
+    sg_buf sgb = sg_list_tail_add(sg, size);
+    if (sgb == INVALID_ADDRESS)
+        goto fail_tail_add;
+    sgb->buf = buffer_ref(b, 0);
+    sgb->size = size;
+    sgb->offset = 0;
+    sgb->refcount = 0;
+    status_handler sh = closure(tracelog.h, tracelog_file_write_complete, sg, sgb, b, complete, false);
+    if (sh == INVALID_ADDRESS)
+        goto fail_mem_sgb;
```
francescolavra, line 394: Should go to fail_tail_add instead

wjhun: Thanks. I fixed the error handling here and removed the extraneous argument to tracelog_file_write_complete.

PR #1773, opened 2022-09-29
Unix: create /proc/self/exe even when /proc/self directory exists

francescolavra:
A given image may be created with a /proc/self directory. The existing code is skipping creation of the /proc/self/exe symlink when the /proc/self directory exists; this PR changes the code so that the symlink is always created when not present, regardless of the presence of the containing folder.

### DIFF
--- a/src/unix/special.c
+++ b/src/unix/special.c
@@ -340,21 +340,29 @@ boolean create_special_file(const char *path, spec_file_open open, u64 size)
 
 void register_special_files(process p)
 {
+    filesystem fs = p->root_fs;
+    tuple self_exe;
     heap h = heap_locked((kernel_heaps)p->uh);
 
-    fs_status fss = filesystem_mkdirpath(p->root_fs, 0, "/proc/self", true);
+    fs_status fss = filesystem_get_node(&fs, p->cwd, "/proc/self/exe", false, false, false,
+        &self_exe, 0);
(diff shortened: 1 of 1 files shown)


src/unix/special.c
```diff
@@ -340,10 +340,16 @@ boolean create_special_file(const char *path, spec_file_open open, u64 size)
...
+    tuple self_exe;
     heap h = heap_locked((kernel_heaps)p->uh);
 
-    fs_status fss = filesystem_mkdirpath(p->root_fs, 0, "/proc/self", true);
+    fs_status fss = filesystem_get_node(&fs, p->cwd, "/proc/self/exe", false, false, false,
+        &self_exe, 0);
     if (fss == FS_STATUS_OK) {
+        filesystem_put_node(fs, self_exe);
+    } else {
+        filesystem_mkdirpath(p->root_fs, 0, "/proc/self", true);
```
wjhun, line 352: Nit, but might want to check that the result here is FS_STATUS_OK or FS_STATUS_EXIST and not some other error.

wjhun (approved):
LGTM

src/unix/special.c
francescolavra: OK, now the symlink is created only if filesystem_mkdirpath() returns FS_STATUS_OK or FS_STATUS_EXIST

PR #1775, opened 2022-10-03
cloud_init klib fixes

### LINKED ISSUE #1774: AWS, qemu `cloud_init` klib Issues
I am trying to use the `cloud_init` klib on AWS (my use-case is to avoid having secrets baked in to the image(s) which seems to be the case for `"Env"` variables, and to have flexibility to change variables for each instance and/or reconfigure an existing instance at runtime). So for this, I am thinking to use the AWS EC2 "user data" option, the contents of which are available within the VM via fetching `http://169.254.169.254/latest/user-data`

I am seeing a few issues (which I could split into separate GitHub Issues if preferred):
1. AWS: When using the `cloud_init` klib on AWS, the first two bytes of the `dest` appear to be random binary data (i.e. `0xffff`, or `0x1234`); the rest of the file is correct. I can confirm that fetching the above URL e.g. at run-time using a test `Program` has the fully correct user data. Using a different URL like `http://httpbin.org/uuid` also has the same problem: the first two bytes are corrupt.
2. AWS: The `cloud_init` mechanism only seems to update the `dest` file on the first boot of the instance where the URL returns a 200; i.e. changing the AWS "user data" and rebooting doesn't overwrite the previous file, it remains with the previous data. (perhaps this is by design, though for my purposes I wish it performed the URL fetch on every boot)
3. QEMU: When using the `cloud_init` klib locally / via QEMU, I can't get it to work at all. It seems to hang at boot when using any URL, e.g. `"src": "https://ops.city"`.

Please let me know if there is any other info I can provide or tests that I can run.

francescolavra:
The first commit fixes a crash that may occur if the application layer of a TLS connection requests that the connection is shut down from within the input buffer handler.
The second commit fixes saving of the contents of a downloaded file and handling of files split in multiple HTTP packets.

Closes #1774.

### DIFF
--- a/klib/cloud_init.c
+++ b/klib/cloud_init.c
@@ -133,26 +133,23 @@ define_closure_function(2, 1, void, cloud_download_done,
     apply(bound(complete), s);
 }
 
-closure_function(3, 2, void, cloud_download_save_complete,
-                 value, v, buffer, content, status_handler, sh,
+closure_function(2, 2, void, cloud_download_save_complete,
+                 buffer, content, status_handler, sh,
                  status, s, bytes, len)
 {
-    value v = bound(v);
-    if (!v)
-        deallocate_buffer(bound(content));
+    deallocate_buffer(bound(content));
     apply(bound(sh), s);
     closure_finish();
 }
 
-closure_function(5, 1, void, cloud_download_save,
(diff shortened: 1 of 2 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1777, opened 2022-10-05
pagecache_node_fetch_internal: reset sgb to 0 when submitting early

sanderssj:
When the pagecache fetches a range and builds up an sg list, if it encounters an already-filled page then it will submit the existing sg_list and use a new sg_list for the rest of the range that is not filled. If sgb is not also reset to 0 then a pointer to the earlier submitted sg_list could be used instead.


wjhun (approved):
LGTM

PR #1778, opened 2022-10-07
init_network_iface(): set default DNS server

francescolavra:
Several klibs (cloud_init, cloudwatch, ntp, radar, syslog) may need to resolve host names. If DHCP is not being used (such as when network interfaces are statically configured), no DNS server is configured in lwIP, which prevents dns_gethostbyname() from working.
This change sets a default DNS server when initializing network interfaces, so that host names can be resolved even if DNS servers are not configured via DHCP.


wjhun (approved):
LGTM

PR #1779, opened 2022-10-11
DNS resolver: fix handling of TTL values in DNS table

francescolavra:
A DNS table entry containing a correspondence between a hostname and an IP address should be invalidated when its TTL expires; to do that, lwIP relies on a function (`dns_tmr`) that should be called periodically by the kernel at a defined time interval. This change adds the DNS timer (which calls the above function) to the list of lwIP timers enabled during kernel initialization: this fixes invalidation of obsolete DNS entries, and allows IP address changes to be detected by the klibs that rely on DNS hostname resolution.


wjhun (approved):
LGTM

PR #1782, opened 2022-10-20
Improve handling of empty timerqueues

sanderssj:
The timerqueue is used mainly for tracking timers in the kernel, which checks the next_expiry field to determine when to set the machine timer. There is, however, ambiguity in checking that field when the expiry timestamp is in the past and the timerqueue is not updated. This could mean either the timerqueue was serviced and there are no more timers in the queue and thus the next_expiry field is stale since it is not updated when there are no timers, or it can mean that the timer has fired and queued the timer service routine but it has not been executed or is executing on a different cpu. This ambiguity creates a bug when the timer interrupt hasn't run yet but the next_expiry is in the past and the scheduler assumes it is empty and resets the timer interrupt to fire much later, preventing the expired timer from being serviced at the appropriate time.
This change fixes that by adding a new 'empty' boolean to the timer queue that reflects the current state of the timerqueue. It gets modified only under the timerqueue lock, and can be checked safely outside of the lock in smp, unlike checking the underlying priorityqueue directly. The scheduler can check this variable to determine an empty timerqueue instead of using an expiry_timer value in the past.
This change also revealed a bug in the timerqueue code where it was incorrectly checking the end/empty of the underlying priorityqueue with a null value instead of with INVALID_ADDRESS.
Another small change is to read the current tsc in the runloop when checking if the timer is updated and then reusing that value later when calculatin the next timeout just before running the user thread.

### DIFF
--- a/src/kernel/schedule.c
+++ b/src/kernel/schedule.c
@@ -99,12 +99,12 @@ static void migrate_from_self(cpuinfo ci, u64 first_cpu, u64 ncpus)
     }
 }
 
-static inline boolean update_timer(void)
+static inline boolean update_timer(timestamp here)
 {
     timestamp next = kernel_timers->next_expiry;
     if (!compare_and_swap_32(&kernel_timers->update, true, false))
         return false;
-    s64 delta = next - now(CLOCK_ID_MONOTONIC_RAW);
+    s64 delta = next - here;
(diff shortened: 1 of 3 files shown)


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1784, opened 2022-10-26
select(2): properly set the epoll type for the current thread

francescolavra:
When poll(2) is called on a given thread, the type of the epoll structure associated to the thread is set to EPOLL_TYPE_POLL; this does not always happen when select(2) is called; when a program calls select() after poll(), the epoll type remains set to EPOLL_TYPE_POLL, which may cause the check_fdesc() function to call poll_notify() instead of select_notify() on the epoll_blocked structure, thereby accessing invalid fields in the internal union of this structure.
This change fixes the above issue by adding an epoll type parameter to the select_get_epoll() function (which is being renamed to thread_get_epoll because it is not used just for select), and properly setting the epoll type in this function according to the argument value.

### DIFF
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -741,15 +741,17 @@ closure_function(3, 1, sysreturn, select_bh,
     return syscall_return(t, rv);
 }
 
-static inline epoll select_get_epoll(void)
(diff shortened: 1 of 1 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1785, opened 2022-10-27
Print out list of loaded klibs on panic or assert fail

sanderssj:
When analyzing backtraces it can be useful to know if and where any klibs are loaded. This change adds a new call to print out a list of the loaded klib information including name, load address, and load size. This print will occur when frame_trace is called since it can be useful to know when looking at frame trace addresses and ensures it is printed out for common failures like panics or assertion failures.


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1786, opened 2022-10-27
Update frequency adjust check in now()

sanderssj:
The now() function asserted that the frequency-adjusted timestamp is newer than the last_raw value as a guarantee that time does not go backwards. This change modifies the check to only assert the unmodified timestamp is greater or equal to the last_raw value and instead modifies the frequency-adjusted timestamp to last_raw if the calculation results in a older timestamp.


wjhun (approved):
LGTM

PR #1792, opened 2022-11-11
KVM: fix detection logic when Hyper-V Enlightenments are enabled

### LINKED ISSUE #1791: vmbus error w/vultr
vultr-nanos-boot

francescolavra:
When running under QEMU with KVM acceleration, the hypervisor signature reported at CPUID function 0x40000000 may not be the KVM signature; specifically, if Hyper-V Enlightenments are enabled in the QEMU command line (for example, with the command line option `--cpu host,hv_relaxed`), the hypervisor reported at CPUID function 0x40000000 is Hyper-V, whereas KVM is reported at CPUID function 0x40000100 (see https://github.com/qemu/qemu/blob/d73c46e4a84e47ffc61b8bf7c378b1383e7316b5/docs/hyperv.txt).
This change fixes the hypervisor detection logic so that KVM is properly detected when Hyper-V Enlightenments are enabled. In addition, the check on the eax value (which indicates the maximum cpuid function present in a given leaf) is being fixed so that the detection works properly on VMs (such as AWS c5 instances) where additional cpuid functions (beyond KVM_CPUID_FEATURES) are present.

Closes #1791.

### DIFF
--- a/src/kernel/kvm_platform.c
+++ b/src/kernel/kvm_platform.c
@@ -10,20 +10,23 @@
 #define kvm_debug(x)
 #endif
 
-#define KVM_CPUID_SIGNATURE 0x40000000
+#define KVM_CPUID_BASE      0x40000000
+#define KVM_CPUID_END       0x40000200
+
+#define KVM_CPUID_SIGNATURE 0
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1793, opened 2022-11-11
Add ltrace functionality

francescolavra:
This feature allows tracing calls made by the application binary to dynamic library functions. It is enabled by adding an 'ltrace' attribute to the root tuple, and prints on the serial console (similarly to the Linux ltrace tool) a line for each dynamic library function call (with the function name) and for each signal delivered to the process (with the signal number).
This feature is implemented for the x86 and aarch64 architectures (in riscv, CPU debug registers cannot be accessed from supervisor mode, and the SBI does not currently have the functionality to access debug registers via SBI calls).
The new functions elf_dyn_parse() and elf_plt_get() have been added to the generic ELF code and are used by ltrace to retrieve the information needed to be able to place software breakpoints where dynamic function calls are made and map each breakpoint to the corresponding function.
The Unix fault handler has been changed so that trap exceptions generated after a single-stepped instruction are correctly handled (either by ltrace, or by generating a SIGTRAP signal).

Example Ops configuration to enable ltrace:
```
  "ManifestPassthrough": {
    "ltrace": {}
  }
```

### DIFF
--- a/platform/pc/Makefile
+++ b/platform/pc/Makefile
@@ -34,6 +34,7 @@ SRCS-kernel.elf= \
 	$(SRCDIR)/kernel/linear_backed_heap.c \
 	$(SRCDIR)/kernel/locking_heap.c \
 	$(SRCDIR)/kernel/log.c \
+	$(SRCDIR)/kernel/ltrace.c \
 	$(SRCDIR)/kernel/mutex.c \
 	$(SRCDIR)/kernel/page.c \
 	$(SRCDIR)/kernel/page_backed_heap.c \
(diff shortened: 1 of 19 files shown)


francescolavra:
I modified the code so that ltrace works also on x86-64 PIE binaries generated by recent gcc toolchains, where the linker does not use lazy symbol binding. In general, now it works for all dynamic library function calls, regardless of whether lazy binding is used.

sanderssj (commented):
LGTM, although maybe a prefix before the library call name would be helpful. The ltrace tool on Linux outputs to stderr so you can easily filter it, but that's not an option here.

francescolavra:
> LGTM, although maybe a prefix before the library call name would be helpful. The ltrace tool on Linux outputs to stderr so you can easily filter it, but that's not an option here.

OK, I added the [LTRACE] prefix to each line output by ltrace.

PR #1794, opened 2022-12-02
Klibs: AWS CloudWatch: add support for sending logs

francescolavra:
This change implements a new console driver that sends log messages to AWS CloudWatch when Nanos runs on an AWS instance.
This feature is enabled by loading the cloudwatch and tls klibs and adding a "logging" tuple to the "cloudwatch" tuple in the root tuple. The "logging" tuple may contain the following attributes:
- "log_group": specifies the CloudWatch log group to which log messages should be sent; if not present, the log group is derived from the image name (taken from the environment variables), or from the name of the user program if no IMAGE_NAME environment variable is present
- "log_stream": specifies the CloudWatch log stream to which log messages should be sent; if not present, the log stream is derived from an instance identifier (e.g. 'ip-172-31-23-224.us-west-1.compute.internal')
The log group and the log stream are automatically created if not existing.

Example contents of Ops configuration file:
```
"ManifestPassthrough": {
  "cloudwatch": {
    "logging": {"log_group":"my_log_group","log_stream":"my_log_stream"}
  }
}
```


klib/cloudwatch.c
```diff
@@ -98,6 +135,305 @@ define_closure_function(1, 1, void, cw_aws_cred_handler,
...
+closure_function(3, 1, boolean, cw_log_meta_ibh,
+                 buffer_handler, parser, boolean *, parsed, buffer_handler, out,
+                 buffer, data)
+{
+    status s = apply(bound(parser), data);
+    if (!is_ok(s)) {
+        msg_err("failed to parse response: %v\n", s);
+        timm_dealloc(s);
+        if (data)
+            apply(bound(out), 0);
```
wjhun, line 230: Should this call closure_finish() if data is null here?

wjhun (approved):
LGTM, minor comments

klib/cloudwatch.c
```diff
@@ -256,39 +583,243 @@ define_closure_function(0, 1, input_buffer_handler, cw_metrics_conn_handler,
...
+}
+
+static void cw_log_post(void)
+{
+    tuple req = allocate_tuple();
+    if (req == INVALID_ADDRESS)
+        goto error;
+    set(req, sym(url), alloca_wrap_cstring("/"));
+    set(req, sym(host), alloca_wrap_cstring(cw.log_servername));
+    set(req, sym(x-amz-target), alloca_wrap_cstring("Logs_20140328.PutLogEvents"));
```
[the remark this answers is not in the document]
francescolavra: The "x-amz-target" string itself is not specific to the logs API, it's used across many AWS APIs, not limited to CloudWatch. I don't think it will ever need to be overridden. I changed the cw_aws_req_send() function to take an additional `const char *target` argument and do `set(req, sym(x-amz-target), alloca_wrap_cstring(target))` at the beginning, so this code is not duplicated for each function call.

klib/cloudwatch.c
francescolavra: Indeed we should call closure_finish() if data is null, I changed the code accordingly.

PR #1796, opened 2023-01-04
Unix socket and getsockopt improvements

sanderssj:
This PR fixes several small issues related to sockets. First, it implements the getsockopt syscall for unix sockets,
which was previously not supported, however it only implements SO_TYPE. Second, it adds support for the
SO_PROTOCOL option for inet and inet6 sockets as well as adds some input checking for type and protocol when
creating sockets in these domains. Finally, it fixes an issue with udp sockets where calling shutdown on a udp
socket would not wake any blocked waiters on receive.


wjhun (approved):
LGTM

PR #1799, opened 2023-01-06
lwIP: implement multithreaded operation

francescolavra:
This change set removes the global lwIP lock and adapts the networking code to the lwIP changes that allow concurrent invocation of lwIP raw API functions. The lwIP lightweight protection implementation (which does not provide protection from concurrent access in multi-core systems) is being disabled, and locking primitives (as well as atomic operations needed for reference counting) are being defined.
A spinlock is now used to protect the lwIP heap.
lwIP functions that inolve a tcp_pcb structure are now called with the PCB lock held (by calling `tcp_lock` and `tcp_unlock`).
tcp_ref() and tcp_unref() are used to properly reference-count tcp_pcb structures used in net/netsyscall.c and net/direct.c.
In net/netsyscall.c, the file descriptor lock is used (instead of the global lwIP lock) to protect the netsock structure from concurrent access. In order to avoid deadlock, the lock of a TCP PCB associated to a socket must not be acquired while holding the file descriptor lock (unless the PCB is in a state where lwIP callbacks are never invoked, e.g. before establishing a TCP connection). In addition, in order to avoid a potential deadlock with the epoll code, notify_dispatch is called without holding the file descriptor lock.

Depends on https://github.com/nanovms/lwip/pull/4

### DIFF
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -91,7 +91,7 @@ void *lwip_allocate(u64 size)
     /* To maintain the malloc/free interface with mcache, allocations must stay
        within the range of objcaches and not fall back to parent allocs. */
     assert(size <= U64_FROM_BIT(MAX_LWIP_ALLOC_ORDER));
-    void *p = allocate_zero(lwip_heap, size);
+    void *p = allocate(lwip_heap, size);
     return ((p != INVALID_ADDRESS) ? p : 0);
 }
 
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -11,7 +11,6 @@
 #define IFF_MULTICAST   (1 << 12)
 
 BSS_RO_AFTER_INIT static heap lwip_heap;
-BSS_RO_AFTER_INIT mutex lwip_mutex;
 
 declare_closure_struct(0, 2, void, net_timeout_handler, u64, expiry, u64, overruns);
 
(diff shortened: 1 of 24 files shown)


src/net/net.c
```diff
@@ -83,14 +80,18 @@ void *lwip_allocate(u64 size)
     /* To maintain the malloc/free interface with mcache, allocations must stay
        within the range of objcaches and not fall back to parent allocs. */
     assert(size <= U64_FROM_BIT(MAX_LWIP_ALLOC_ORDER));
-    void *p = allocate_zero(lwip_heap, size);
+    spin_lock(&lwip_heap_lock);
```
wjhun, line 83: Maybe this wouldn't cause a problem as it is called from lwip_init() and before lwip is in use, but the closure allocation from lwip_heap in sys_timeouts_init() (as well as the closure finish, if it's ever actually invoked) should be protected by lwip_heap_lock as well. Using a locking_heap_wrapper in lieu of manually locking would eliminate such cases of unlocked use.

src/net/netsyscall.c
```diff
@@ -945,9 +985,7 @@ sysreturn socket_ioctl(struct sock *s, unsigned long request, vlist ap)
             return -EFAULT;
         if ((ifreq->ifr.ifr_mtu <= 0) || (ifreq->ifr.ifr_mtu > MTU_MAX))
             return -EINVAL;
-        lwip_lock();
         struct netif *netif = netif_find(ifreq->ifr_name);
-        lwip_unlock();
         if (!netif)
             return -ENODEV;
         netif->mtu = ifreq->ifr.ifr_mtu;
```
wjhun, line 991: Need netif_unref(netif) here

wjhun (approved):
LGTM

francescolavra:
In lwipopts.h, I added a `#define LWIP_PBUF_REF_T u32_t` to change the type used for reference-counting pbufs from the default u8_t to u32_t, because the RISC-V architecture doesn't support atomic operations on data smaller than 32 bits

PR #1800, opened 2023-01-06
Network performance improvements for VMs with high vCPU count

francescolavra:
This change set amends the Unix epoll code and the virtio-net driver in order to improve network performance (especially on GCP instances with a high vCPU count).

Analyzing lock stats while stress-testing a Go web server running on a multi-core instance revealed a high contention of the lock that protects the file descriptor list associated to an epoll instance. Since this list is only modified when adding or removing a file descriptor to/from an epoll instance (and not when invoking the epoll_wait syscall, which happens more frequently), replacing the lock with a read-write lock decreases the lock contention and improves the performance of the web server when servicing multiple connections simultaneously.
The same analysis of lock stats also revealed a high contention of the lock that protects the kernel locked heap when processing network packets. In the virtio-net driver, using a dedicated objcache for the TX completion closures and embedding the RX closures in the xpbuf structure decreases the lock contention and improves the performance of the web server.

### DIFF
--- a/src/unix/poll.c
+++ b/src/unix/poll.c
@@ -75,7 +75,7 @@ struct epoll {
     enum epoll_type epoll_type;
     closure_struct(epoll_free, free);
     heap h;
-    struct spinlock fds_lock;
+    struct rw_spinlock fds_lock;
     vector events;              /* epollfds indexed by fd */
     int nfds;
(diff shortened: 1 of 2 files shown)


wjhun (approved):
Looks good!

Issue #1797, opened 2023-01-05
Question(haproxy): level: SOL_TCP(6), optname(s): TCP_MAXSEG(2) / TCP_FASTOPEN(23)


rinor:
While packaging/booting `haproxy v2.7.1` (latest stable release) I'm getting some:
```
netsock_getsockopt error: getsockopt unimplemented optname:
```
---
- **Boot**
```
➜ ops pkg load -f --local -p 8081 -p 8082 haproxy-2.7.1-bullseye 
booting /home/nanos/.ops/images/haproxy ...
en1: assigned 10.0.2.15
netsock_getsockopt error: getsockopt unimplemented optname: fd 3, level 6, optname 2
netsock_getsockopt error: getsockopt unimplemented optname: fd 3, level 6, optname 2
Note: setting global.maxconn to 32760.
Available polling systems :
      epoll : pref=300,  test result OK
       poll : pref=200,  test result OK
     select : pref=150,  test result FAILED
Total: 3 (2 usable), will use epoll.

Available filters :
	[BWLIM] bwlim-in
	[BWLIM] bwlim-out
	[CACHE] cache
	[COMP] compression
	[FCGI] fcgi-app
	[SPOE] spoe
	[TRACE] trace
Using epoll() as the polling mechanism.
netsock_getsockopt error: getsockopt unimplemented optname: fd 4, level 6, optname 2
netsock_getsockopt error: getsockopt unimplemented optname: fd 4, level 6, optname 23
netsock_getsockopt error: getsockopt unimplemented optname: fd 5, level 6, optname 2
netsock_getsockopt error: getsockopt unimplemented optname: fd 5, level 6, optname 23
en1: assigned FE80::F898:32FF:FE8E:EC5C
```
---
- **Identified** as:

```
getsockopt unimplemented optname: fd 3, level SOL_TCP, optname TCP_MAXSEG  
getsockopt unimplemented optname: fd 3, level SOL_TCP, optname TCP_MAXSEG  
getsockopt unimplemented optname: fd 4, level SOL_TCP, optname TCP_MAXSEG  
getsockopt unimplemented optname: fd 4, level SOL_TCP, optname TCP_FASTOPEN
getsockopt unimplemented optname: fd 5, level SOL_TCP, optname TCP_MAXSEG  
getsockopt unimplemented optname: fd 5, level SOL_TCP, optname TCP_FASTOPEN
```
---
- **Testing Patch** just to confirm I got the correct opts
```patch
[...]

PR #1801, opened 2023-01-06
netsock_getsockopt(): add support for TCP_MAXSEG and TCP_FASTOPEN

francescolavra:
These options are accepted when the `level` argument is SOL_TCP.
TCP_MAXSEG indicates the maximum segment size of a TCP connection.
TCP_FASTOPEN indicates the maximum number of pending TCP Fast Open (TFO) requests a listening socket allows before disabling TFO (the kernel currently does not support TFO, thus the returned value is always zero).

Closes https://github.com/nanovms/nanos/issues/1797.


wjhun (approved):
LGTM

PR #1802, opened 2023-01-08
sendfile(2): change file offset on output file

francescolavra:
When sendfile() is called on an output file descriptor corresponding to a regular file, it should write at the current file offset and update the offset according to the number of bytes written. The existing code is writing at offset zero and is not updating the offset.
This change fixes the above issue and amends the sendfile runtime test so that it would have failed without this fix.


wjhun (approved):
LGTM

PR #1804, opened 2023-01-09
File syscalls: initialize fsfile pointer and check for null value

francescolavra:
When a file is opened, the fsf member of the file structure should be properly set to NULL for non-regular files; with this fix, fsync(2) works properly when invoked on a file descriptor corresponding to a directory, and read/write (and similar) syscalls properly return an error code when invoked on file descriptors corresponding to non-regular files.


wjhun (approved):
LGTM

PR #1807, opened 2023-01-12
netsock_shutdown(): fix handling of multiple calls on a socket

francescolavra:
The shutdown syscall can be invoked on a given socket multiple times; if it is invoked once with `how` set to SHUT_RD and once with `how` set to SHUT_RW, at the second call both directions of the underlying TCP connection are shut down, which means that the lwIP tcp_pcb structure can be deallocated at any time and is therefore unsafe to reference. The existing code is failing to detect this condition, which may cause lwIP functions to be called on a tcp_pcb structure that has already been deallocated.
This change fixes the above issue. In addition, the initial check on the state member of the socket structure is now performed with the lwIP lock held, for multi-thread safety.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1074,11 +1074,20 @@ static sysreturn netsock_shutdown(struct sock *sock, int how)
     }
     switch (s->sock.type) {
     case SOCK_STREAM:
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1808, opened 2023-01-14
Klibs: add new library implementing GCP logging

francescolavra:
This change set adds a klib implementing a console driver that sends console output to GCP logs.
Instance-specific information that needs to be known in order to interface with the GCP logging API is retrieved from the instance metadata server, which is reachable at address 169.254.169.254.
GCP logging is enabled by loading the gcp and tls klibs and adding to the root tuple a "gcp" tuple that contains a "logging" tuple.
The "logging" tuple may optionally contain a "log_id" attribute that specifies the [LOG_ID] string that is sent in the "logName" parameter (which is formatted as "projects/[PROJECT_ID]/logs/[LOG_ID]") associated to GCP log entries; if not present, the log ID is derived from the instance hostname as retrieved from the metadata server.

In order for the GCP klib to retrieve the appropriate credentials needed to communicate with the GCP logging server, the instance on which it runs must be associated to a service account (see https://cloud.google.com/compute/docs/access/service-accounts). Instances created by Ops are associated to a service account via the CloudConfig.InstanceProfile configuration parameter (https://github.com/nanovms/ops/pull/1407).

Example contents of Ops configuration file:
```
  "ManifestPassthrough": {
    "gcp": {
      "logging": {
        "log_id": "my_log"
      }
    }
  }
```

The first commit is a fix to an issue that has been detected while stress-testing the GCP klib.

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -4,6 +4,7 @@ MBEDTLS_DIR=	$(VENDORDIR)/mbedtls
 PROGRAMS= \
 	cloud_init \
 	cloudwatch \
+	gcp \
 	ntp \
 	radar \
 	syslog \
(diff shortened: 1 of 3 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1811, opened 2023-01-19
Fix last_raw update race condition

sanderssj:
The vdso last_raw gets updated periodically, and if it were updated on one processor while another is running now(), there can be a race condition where the now() cpu fetches monotonic time, then the other cpu updates last_raw with a slightly later monotonic time, then the now() cpu fetches the saved last_raw but then fails the assertion check comparing the two timestamps.

This change fixes the problem by reading the last_raw value before fetching the current monotonic time.


francescolavra (commented):
LGTM

wjhun (approved):
LGTM

PR #1812, opened 2023-01-21
Lookup interp in paths pointed by LD_LIBRARY_PATH

### LINKED ISSUE #1810: Add support for LD_LIBRARY_PATH for intrep
`ops` now recognizes the `LD_LIBRARY_PATH` env variable in config: https://github.com/nanovms/ops/commit/a535c7864cc574da0dad9eca4b6b617ea0ab11b9

Similar to regular libs, ops also allow resolving interpreter path in one of the paths pointed by `LD_LIBRARY_PATH` https://github.com/nanovms/ops/commit/8e0f29f8f959f0899897cafe8ddb7d392a0a276f

However, nanos ignores `LD_LIBRARY_PATH`; when I launch nanos with interp at non-standard location and `LD_LIBRARY_PATH` set, it halts from https://github.com/nanovms/nanos/blob/c90e9ebd4743e4e13ce4408291623685910bf9a0/src/unix/exec.c#L251-L253

```sh
booting /Users/am11/.ops/images/myApp ...
en1: assigned 10.0.2.15
couldn't find program interpreter /lib64/ld-linux-x86-64.so.2
```

am11:
Fixes #1810.

The app is running on M1 with this fix, however, I am getting exit code 1 from qemu after otherwise a clean run. 🤔

### DIFF
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -259,7 +259,6 @@ process exec_elf(buffer ex, process kp)
                     vector v = split(heap_locked(kh), libPaths, ':');
                     buffer i;
                     vector_foreach(v, i) {
-                            exec_debug("pt_intrep 3.0\n");
                         /* null entries ("//") are skipped in path */
                         if (buffer_length(i) == 0)
                             continue;
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -248,9 +248,38 @@ process exec_elf(buffer ex, process kp)
     foreach_phdr(e, p) {
         if (p->p_type == PT_INTERP) {
             char *n = (void *)e + p->p_offset;
-            interp = resolve_path(root, split(heap_locked(kh), alloca_wrap_buffer(n, runtime_strlen(n)), '/'));
-            if (!interp) 
-                halt("couldn't find program interpreter %s\n", n);
+            vector interp_vec = split(heap_locked(kh), alloca_wrap_buffer(n, runtime_strlen(n)), '/');
+            interp = resolve_path(root, interp_vec);
+            if (!interp)  {
+                // check if LD_LIBRARY_PATH is set and interpreter exists in one
+                // of the given paths (separated by colon)
+                buffer libPaths = get(get_environment(), sym(LD_LIBRARY_PATH));
+                if (libPaths) {
+                    char* intrep_name = vector_get(interp_vec, vector_length(interp_vec) - 1);
+                    vector v = split(heap_locked(kh), libPaths, ':');
+                    buffer i;
+                    vector_foreach(v, i) {
+                            exec_debug("pt_intrep 3.0\n");
+                        /* null entries ("//") are skipped in path */
+                        if (buffer_length(i) == 0)
+                            continue;
+
+                        vector vec = allocate_vector(heap_locked(kh), 10);
+                        vector_push(vec, i);
+                        vector_push(vec, intrep_name);
+
+                        i = join(heap_locked(kh), vec, '/');
+                        interp = resolve_path(root, split(heap_locked(kh), i, '/'));
+                        if (interp)  {
+                            break;
+                        }
+                    }
+                }
+
+                if (!interp)  {
+                    halt("couldn't find program interpreter %s\n", n);
+                }
+            }
         } else if (p->p_type == PT_LOAD) {
             if (p->p_vaddr < load_range.start)
                 load_range.start = p->p_vaddr;
@@ -335,4 +364,3 @@ process exec_elf(buffer ex, process kp)
     start_process(t, entry);
     return proc;    
 }
-
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -253,10 +253,10 @@ process exec_elf(buffer ex, process kp)
             if (!interp)  {
                 // check if LD_LIBRARY_PATH is set and interpreter exists in one
                 // of the given paths (separated by colon)
-                buffer libPaths = get(get_environment(), sym(LD_LIBRARY_PATH));
-                if (libPaths) {
+                buffer lib_paths = get(get_environment(), sym(LD_LIBRARY_PATH));
+                if (lib_paths) {
                     char* intrep_name = vector_get(interp_vec, vector_length(interp_vec) - 1);
-                    vector v = split(heap_locked(kh), libPaths, ':');
+                    vector v = split(heap_locked(kh), lib_paths, ':');
                     buffer i;
                     vector_foreach(v, i) {
                         /* null entries ("//") are skipped in path */


src/unix/exec.c
```diff
@@ -248,9 +248,38 @@ process exec_elf(buffer ex, process kp)
...
+            if (!interp)  {
+                // check if LD_LIBRARY_PATH is set and interpreter exists in one
+                // of the given paths (separated by colon)
+                buffer libPaths = get(get_environment(), sym(LD_LIBRARY_PATH));
+                if (libPaths) {
+                    char* intrep_name = vector_get(interp_vec, vector_length(interp_vec) - 1);
+                    vector v = split(heap_locked(kh), libPaths, ':');
+                    buffer i;
+                    vector_foreach(v, i) {
+                            exec_debug("pt_intrep 3.0\n");
```
am11, line 262:
```suggestion
```

am11:
Local testing is _possible_ on M1 machine, but it is quite hacky right now. Here is what I did:

```zsh
# brew install binutils
# then build anos and ops as x86_64
# (because of reasons like: our configs are using 'aarch64' and Apple uses 'arm64')
% PATH=$PATH:/opt/homebrew/Cellar/x86_64-elf-binutils/2.40/x86_64-elf/bin \
    arch -x86_64 make run-noaccel

% cp output/platform/pc/bin/kernel.img output/platform/pc/boot/bootx64.efi \
     output/platform/pc/boot/boot.img output/tools/bin/mkfs \
     ~/.ops/0.1.43/

# then hack ops to:
# 1. stop overwriting the files in ~/.ops/0.1.43
# 2. stop passing hvf and rdtscp to qemu:
#     i've a patch for rosetta2 detection for this, but hopefully we
#     won't be needing roesetta emulation once this goes through
% arch -x86_64 ~/git/ops/ops run --show-debug -v -c config.json publish/myApp
...
Hello, World!
exit status 1
```

rinor:
Most probably than not I'm lacking some context here, but do we really need this?
I would opt for making `ops` a bit smarter and maybe enhancing `ops pkg from-docker` if possible.

am11:
`ops` already takes `LD_LIBRARY_PATH` into account when looking up the interpreter and libs. `nanos` fails to run that image (when loading the ELF) due to this condition. Anyways, this doesn't hurt the happy path when interpreter is at the exact location where the header is pointing to.

src/unix/exec.c
```diff
@@ -248,9 +248,37 @@ process exec_elf(buffer ex, process kp)
...
-                halt("couldn't find program interpreter %s\n", n);
+            vector interp_vec = split(heap_locked(kh), alloca_wrap_buffer(n, runtime_strlen(n)), '/');
+            interp = resolve_path(root, interp_vec);
+            if (!interp)  {
+                // check if LD_LIBRARY_PATH is set and interpreter exists in one
+                // of the given paths (separated by colon)
+                buffer libPaths = get(get_environment(), sym(LD_LIBRARY_PATH));
+                if (libPaths) {
+                    char* intrep_name = vector_get(interp_vec, vector_length(interp_vec) - 1);
+                    vector v = split(heap_locked(kh), libPaths, ':');
```
am11, line 259:
```suggestion
                buffer lib_paths = get(get_environment(), sym(LD_LIBRARY_PATH));
                if (lib_paths) {
                    char* intrep_name = vector_get(interp_vec, vector_length(interp_vec) - 1);
                    vector v = split(heap_locked(kh), lib_paths, ':');
```

eyberg:
@am11 seems like a few issues here; prob worth moving the ops related stuff to a diff tkt/issue but fwiw we do support both arm && x86 for both nanos && ops and I feel the defaults should always be x86/with hardware accel - as for m1s since you can't have hardware accel for x86 payloads i don't have a great answer for what the defaults should be but the vast majority of end-users are still going to be deploying to x86 so my assumption would be that the default (on arm m1s) would be to have ops/qemu be arm but have hardware accel off and use x86 (and of course we should be able to flag it on/off as desired, as you can run arm natively with hardware accel and then deploy to arm servers too (we can run on arm-based GCP/AWS today)), as for this particular change i'm not opposed to it but i'll let the others chime in instead;

as for the root issue - running .net as desired in https://github.com/nanovms/ops-examples/issues/47#issuecomment-1399235109 - it might make the most sense to have a pre-made package for .net (or several) that people can use and that way they are only adding their own libs as necessary) - i don't know how much the env actually changes but starting with a base package prob makes the most sense here - happy to help get that up on https://repo.ops.city if that's a route to take

am11:
@eyberg, this PR is the continuation of the bugfix: https://github.com/nanovms/ops/commit/8e0f29f8f959f0899897cafe8ddb7d392a0a276f. The "bug" was that we were finding interpreter with absolute path and getting back the exact absolute path back every single time. So it was either that the find call was redundant or we were suppoed to find by interpreter name. I chose the latter to fix the bug.

Now that the ops PR is merged, this PR is complementing it by making nanos also respect for what ops has allowed. Not taking this change means we should revert the other one because it is either both or none. I don't see how this can be a problem with this very simply patch: changing `panic` to `try another location before panic`.

francescolavra:
@am11 this change would make Nanos behave differently from the Linux kernel when executing an ELF program: on Linux, LD_LIBRARY path is not used to locate the interpreter file, it is only used by the interpreter itself to locate any other dynamic libraries needed by the ELF program. So I don't think we should make this change: the interpreter in the VM image must be located where the PT_INTERP header says. On a Linux machine, if you move the /lib64/ld-linux-x86-64.so.2 file to some other location you won't be able to run dynamically linked executables by invoking them with their file path, regardless of what is in LD_LIBRARY_PATH.
As for https://github.com/nanovms/ops/commit/8e0f29f8f959f0899897cafe8ddb7d392a0a276f, it seems to me that change to Ops is not incompatible with the current code in Nanos: Ops may need to retrieve both the interpreter and the dynamic libs from a non-standard location (e.g. /tmp/linux/) in the host filesystem, but the user must make sure (with an appropriate config using the "Files"", Dirs" and MapDirs" attributes) that these files end up at the correct location in the guest filesystem.

francescolavra:
> however, I am getting exit code 1 from qemu after otherwise a clean run

Exit code 1 from Qemu indicates that the program executed successfully (i.e. its exit code was 0). This is the expected behavior, and is due to the fact that we use the isa-debug-exit Qemu device to be able to return an exit code from the guest program to Qemu. The way this isa-debug-exit device works is it takes the program exit code `val` and does `exit((val << 1) | 1)`

am11:
> this change would make Nanos behave differently from the Linux kernel when executing an ELF program

OK. I didn't knew that we want to keep compatibility with Linux kernel at all costs (even in the exceptional cases?).

> "Files"", Dirs" and MapDirs" attributes

Unfortunately, MapDirs does not work in this case. The target directory exists in the image (without the desired interpreter) so ops refuses to overwrite it. That was the first thing I tried before sending the ops patch last week.

francescolavra:
We don't necessarily want to keep compatibility with the Linux kernel at all costs, but before adding new code to the kernel that makes it deviate from Linux, if there are other ways to make the whole thing work we tend to prefer those.
In this case, I think Ops could improved to deal with such edge cases, or alternatively, it seems that a manual copy of the ld-linux-x86-64.so.2 file to the lib64 folder (meaning the folder on the host that will end up at /lib64 in the guest) before creating the image would suffice.

PR #1817, opened 2023-02-07
pagecache_commit_dirty_ranges(): do not write past node length

francescolavra:
Writing to a file-backed memory area must not cause the length of the underlying file to change, even if the write happens to a location beyond the current file length (e.g. when the file length is not a multiple of the page size). The kernel page cache code that implements writeback of dirty pages written by userspace via mmapped memory is invoking the fs_write closure with a range argument corresponding to a full page size (or a multiple of the page size), which may cause the filesystem code to change the file length.
This change fixes the above issue, and adds to the mmap runtime tests a test case that would fail without this fix.


wjhun (approved):
LGTM

PR #1818, opened 2023-02-14
Filesystem: add support for 9P, using virtIO-9P devices

francescolavra:
This change adds support for the 9P filesystem exported by virtIO-9P devices, which allows the kernel to access files on the host filesystem (which can be made accessible to the guest by QEMU with the `-virtfs` command line option).
The numerical value of the `mount_tag` flag in the virtfs command line option is reflected in the attachment point that is associated to the volume created by the kernel when the relevant virtIO-9P device is probed, so that it is possible to instruct the kernel to mount a 9P filesystem by providing a `mounts:(%<attachment_id>: <mount_dir>)` directive in the manifest.

The Makefile for the PC platform has been amended to support the VIRTFS Makefile variable, which when non-empty is used as the path of the host folder to be exported to the guest: example to export the folder /mnt: `make run VIRTFS=/mnt`.


sanderssj (commented):
Looks good to me. I tried a few tests too and it works well. Just curious, what's the motivation for the vqmsg_commit_sync function instead of using a completion? While the host system should generally be fast, it could still theoretically access a network volume on the host which may be pretty slow, making a lot of busy waiting in the vm.

francescolavra:
> what's the motivation for the vqmsg_commit_sync function instead of using a completion?

Using a completion would mean that nearly all the callback functions in the filesystem struct would have to be asynchronous (e.g. take a status_handler argument to be invoked on completion), which would complicate the generic filesystem code (e.g. the lookup callback being asynchronous means that filesystem_resolve_cstring() also has to be asynchronous, so all the code where file paths are resolved needs to be changed).
Alternatively, we would need to implement some sort of suspend/resume functionality (similar to Linux waitqueues) where the current context is suspended after sending a command to the virtio device, and resumed when the response is received. If you think we should do that, let me know.

wjhun:
> Alternatively, we would need to implement some sort of suspend/resume functionality (similar to Linux waitqueues) where the current context is suspended after sending a command to the virtio device, and resumed when the response is received. If you think we should do that, let me know.

Such functionality exists with context_suspend() and context_schedule_return(). The mutex code avails itself of it; see mutex_lock_internal() and mutex_unlock().

francescolavra:
> Such functionality exists with context_suspend() and context_schedule_return().

OK, I will use that. A consequence of the filesystem code being subject to context suspend/resume is that we can't use a spinlock as filesystem lock, because it could lead to a deadlock if a context is suspended while holding the lock. So I will change the filesystem lock to use a mutex instead.

francescolavra:
PR updated. Changes from previous version:
- v9p_request() and v9p_readdir() now suspend the running context instead of busy-waiting for a response from the virtio device
- the filesystem lock is now a mutex instead of a spinlock, so that a context can be safely suspended with the filesystem lock held (https://github.com/nanovms/nanos/pull/1818/commits/13f608c3b98c9dc7ae85424b417e30c99cbb40e7)
- context acquire/resume behavior during kernel shutdown has been fixed, so that it is possible for a shutdown completion to suspend its context (https://github.com/nanovms/nanos/pull/1818/commits/f6ac009b9670f73b949bc470ac1c3f36c872cb90)
- in pagecache_write_sg(), the fs_reserve closure is applied without the pagecache node lock held, so that it can safely suspend the running context (https://github.com/nanovms/nanos/pull/1818/commits/014f1f3b17d12cab8eceb0e27c06a499911b120d)
- in p9_fsf_free(), the metadata tuple associated to a dentry is deleted immediately (instead of after the pagecache sync) so that it cannot be looked up (e.g. via its parent tuple) before being deleted
- in p9_readdir(), the use of sym_this() has been removed because it allocates stack memory, which causes stack overflow problems when reading directories with a high number of files

src/fs/9p.c
```diff
@@ -0,0 +1,813 @@
...
+                else
+                    t = 0;
+                p9_dentry dentry;
+                if (t) {
+                    dentry = p9_get_dentry_from_md(fs, t);
+                } else {
+                    t = allocate_tuple();
+                    dentry = 0;
+                }
+                p9_debug("  dir entry '%s', type 0x%x, md %p\n", name, entry->type, t);
```
sanderssj, line 386: 'name' no longer exists, have to use name_sym and a different format specifier

sanderssj (approved):
LGTM

src/fs/9p.c
francescolavra: Thanks, I replaced 'name' with 'symbol_string(name_sym)', using %b as format specifier.

PR #1819, opened 2023-02-14
Add option to run user program after getting ip4 address

sanderssj:
Some user programs assume the network is ready to go when executing, but
if dhcp takes a long time then the user program may hit network errors
if an address and gateway isn't acquired yet. This change adds a new manifest
option 'exec_wait_for_ip4_secs', which will delay running the user program
until either an ip4 address is acquired or the timeout specified in the option
has elapsed. If a static address is set then the program is executed immediately.

Additionally, the translation of lwip to errno codes has been updated to more
accurately reflect the nature of the lwip errors.

### DIFF
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -13,6 +13,14 @@
 BSS_RO_AFTER_INIT static heap lwip_heap;
 BSS_RO_AFTER_INIT mutex lwip_mutex;
 
+declare_closure_struct(0, 2, void, net_timeout_handler, u64, expiry, u64, overruns);
+
+static struct {
+    status_handler ip4_complete;
+    struct timer timeout;
+    closure_struct(net_timeout_handler, timeout_handler);
+} net;
+
 /* Pretty silly. LWIP offers lwip_cyclic_timers for use elsewhere, but
    says to use LWIP_ARRAYSIZE(), which isn't possible with an
    incomplete type. Plus there's no terminator to the array. So we
@@ -93,6 +101,17 @@ void lwip_deallocate(void *x)
     deallocate(lwip_heap, x, -1ull);
 }
 
+static boolean netif_ready(struct netif *n)
+{
+    return !ip_addr_isany(netif_ip_addr4(n));
+}
+
+static void check_netif_ready(struct netif *netif)
+{
+    if (netif_ready(netif))
+        remove_timer(kernel_timers, &net.timeout, 0);
+}
+
 static void lwip_ext_callback(struct netif* netif, netif_nsc_reason_t reason,
                               const netif_ext_callback_args_t* args)
 {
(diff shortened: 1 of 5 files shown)


src/net/lwipopts.h
```diff
@@ -77,6 +77,7 @@ typedef unsigned long size_t;
 #define LWIP_DHCP_BOOTP_FILE 1
 #define LWIP_DNS 1
 #define LWIP_INCLUDED_POLARSSL_MD5  1
+#define LWIP_NETIF_STATUS_CALLBACK 1
```
francescolavra, line 80: Instead of enabling an additional lwip option, we could just use the existing lwip_ext_callback function, i.e. add the ip4_complete logic in the code that handles the (reason & LWIP_NSC_IPV4_ADDRESS_CHANGED) condition

src/net/net.c
```diff
@@ -276,6 +285,60 @@ static boolean get_static_ip6_config(tuple t, struct netif *n, const char *ifnam
...
+}
+
+void ip4_when_ready(status_handler complete, timestamp timeout)
+{
+    struct netif *n;
+    if ((n = netif_get_default()) && netif_ready(n)) {
+        apply(complete, STATUS_OK);
+        return;
+    }
+    spin_lock(&net.lock);
```
francescolavra, line 315: If the interface becomes ready after the above check but before we can acquire this lock, the completion will only be invoked at the timeout. We could acquire the lock before checking for netif_ready(), or alternatively we could get rid of the spinlock altogether and do `atomic_swap_64(&net.ip4_complete, 0)` to handle concurrency.

src/unix/system_structs.h
```diff
@@ -83,6 +83,7 @@ typedef struct iovec {
 #define EMSGSIZE        90		/* Message too long */
 #define EPROTOTYPE      91		/* Wrong protocol type for socket */
 #define EOPNOTSUPP      95		/* Operation not supported */
+#define ECONNABORTED    103     /* Software caused connection abort */
```
francescolavra, line 86: Nit: alignment (I know, the presence of tabs in the preceding lines doesn't help)

francescolavra (commented):
LGTM

src/net/net.c
```diff
@@ -93,6 +101,21 @@ void lwip_deallocate(void *x)
...
+static boolean netif_ready(struct netif *n)
+{
+    return !ip_addr_isany(netif_ip_addr4(n));
+}
+
+static void check_netif_ready(struct netif *netif)
+{
+    if (netif_ready(netif)) {
+        remove_timer(kernel_timers, &net.timeout, 0);
+        status_handler c = pointer_from_u64(atomic_swap_64((u64 *)&net.ip4_complete, 0));
```
wjhun, line 113: Since cancelling a timer always invokes the timer handler, a further simplification would be to only invoke the status handler from net_timeout_handler, removing the check for ```overruns == timer_disabled```. Then this function only needs to call remove_timer() and nothing else, obviating the atomic swaps. The status handler may simply be stored in the timer handler closure.

wjhun (approved):
LGTM, with one simplification suggested

PR #1820, opened 2023-02-21
Go runtime tests: update RunConfig struct field names

francescolavra:
The `Imagename` field has been renamed to `ImageName`.


rinor:
my bad, missed that in here

Issue #1813, opened 2023-01-23
feat(klib - cloud_init): HTTP -> Environment (setenv)


rinor:
It would be useful to have the possibility to get and inject environment variables from `cloud_init` download functionality, in addition to file download capability, ex:

- **config.json**
```json
  "ManifestPassthrough": {
    "cloud_init": {
      "download_env": [
        {
          "src": "https://config.provisioner/env01",
          "auth": "Basic dXNlcjpwYXNzd2Q="
        }
      ],
      "download": [
        {
          "src": "https://config.provisioner/e.cfg",
          "dest": "/file.cfg",
          "overwrite": "t",
          "auth": "Basic dXNlcjpwYXNzd2Q="
        }
      ]
    }
  }
```
---

- **https://config.provisioner/env01** **output** can be whatever is easier/faster to implement, ex (just one format can be supported, not all of them, and can deviate from the std shell usage if needed to make the implementation easier):
```ini
VAR_NAME_01="value 01"
VAR_NAME_02="value_02"
VAR_NAME_03=123
```
or
```json
{
  "VAR_NAME_01": "value 01",
  "VAR_NAME_02": "value_02",
  "VAR_NAME_03": 123
}

```

or specific env section (for more advanced usage in the future)
```json
{
  "Env": {
    "VAR_NAME_01": "value 01",
    "VAR_NAME_02": "value_02",
    "VAR_NAME_03": 123
  }
}
```
---

- **expected** actions/results
  - `cloud_init` issues http request to `download_env.src`
  - response content is processed/parsed and if correct env(s) are injected, but response content is not saved to any file, but zeroed
  - injected env(s) are available to `program`
---

- case scenario where such feature is nice to have
  - application uses encrypted config/files from the provisioner (cloud_init file(s) download) and applications requires the password/key/.... to be present on an `env` var upon boot. Since password/key/... (not necessarily the config/file(s) content) is generated and rotated upon each init request  (when the image boots), we can't set it in advance through the static env config during image creation.

francescolavra:
@rinor we are working on this feature. Stay tuned!

francescolavra:
#1822 implements support for populating environment variables retrieved from an HTTP(S) server using JSON syntax.
Example Ops configuration to populate the environment when the server response follows the format below:
```
{
  "VAR_NAME_01": "value 01",
  "VAR_NAME_02": "value_02",
  "VAR_NAME_03": 123
}
```
```
{
  "ManifestPassthrough": {
[...]

PR #1822, opened 2023-03-01
Klibs: cloud_init: add support for environment variables

francescolavra:
This change adds to the cloud_init klib support for populating the environment of the user process with variables whose name and value are retrieved from an HTTP(S) server during startup.
The configuration syntax in the manifest to use this functionality is as in the following example:
```
download_env:[
  (
    src: "http://10.0.2.2:8200/v1/secret/data/hello"
    auth: "Bearer hvs.6v6yY5yZf32uZzaa7HhiX6AZ"
    path: "attr1/attr2"
  )
]
```
The `download_env` attribute is an array where each element specifies a download source URL, an optional authentication header, and an optional attribute path.
For each element in the `download_env` attribute, the klib executes an HTTP request to the specified URL, and if the peer responds successfully, the response body is parsed as a JSON object, and from this object an "environment object" is retrieved. If the attribute path (i.e. the `path` element in the manifest) is not present (or is empty), the environment object corresponds to the root JSON object of the response body; otherwise, for each element in the attribute path (where the element separator is the / character), a nested JSON object is retrieved from the response body by looking up an attribute named after the element (the corresponding attribute value must be a JSON object): the environment object is the nested object corresponding to the last element of the attribute path. Once the environment object is identified, all string-valued attributes in this object are converted to environment variables (any non-string-valued attributes are ignored).

Examples:
- an empty attribute path can be used to retrieve the environment variables from the following response body:
```
{
  "VAR1": "value1",
  "VAR2": "value2"
}
```
- an attribute path set to "obj1/obj2" can be used to retrieve the environment variables from the following response body:
```
{
  "obj1: {
    "obj2": {
      "VAR1": "value1",
      "VAR2": "value2"
    }
  }
}
```

Closes https://github.com/nanovms/nanos/pull/1813.

### DIFF
--- a/klib/cloud_init.c
+++ b/klib/cloud_init.c
@@ -14,6 +14,18 @@ enum cloud {
     CLOUD_UNKNOWN
 };
 
+enum cloud_init_task_op {
+    CLOUD_INIT_TASK_OP_START,
+    CLOUD_INIT_TASK_OP_DELETE,
+};
+
+typedef closure_type(cloud_init_task, void, enum cloud_init_task_op, void *);
+
+typedef closure_type(download_recv, boolean, buffer_handler, buffer);
+
+declare_closure_struct(2, 2, void, cloud_download_task,
+                       download_recv, recv, thunk, cleanup,
+                       int, op, void *, arg);
 declare_closure_struct(2, 1, void, cloud_download_done,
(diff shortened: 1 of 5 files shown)


sanderssj (commented):
LGTM

wjhun (approved):
LGTM

PR #1830, opened 2023-03-19
sched_getaffinity: check cpusetsize prior to calling lookup_thread

wjhun:
A call to sched_getaffinity with a cpusetsize that would result in an -EINVAL returned could mistakenly hold a thread reference. This moves the check for cpusetsize in advance of the call to lookup_thread.


sanderssj (commented):
LGTM

PR #1831, opened 2023-03-19
get_robust_list: fix misuse of thread_release

wjhun:
In get_robust_list, thread_from_tid is called if pid != 0, yet thread_release is called unconditionally, resulting in an incorrect thread reference count. This conditionalizes the release, as well as a check only necessary after the call to thread_from_tid, on pid != 0.


sanderssj (commented):
LGTM

PR #1832, opened 2023-03-19
unixsock_close: close rather than deallocate sockets in conn_q

wjhun:
The existing code in unixsock_close() could create a condition whereby a
reference for a deleted child socket would remain in the process file
descriptor vector. This could materialize as a use-after-free and crash upon
process exit.

The remedy presented here allows the child socket to be created upon
connection, but defers the allocation of an fd until the child is accepted
from the listening socket. When a listening socket is closed, its child
sockets are closed via the usual method rather than deleted outright.

### DIFF
--- a/src/unix/socket.c
+++ b/src/unix/socket.c
@@ -136,7 +136,7 @@ static void unixsock_disconnect(unixsock s)
     }
 }
 
-static unixsock unixsock_alloc(heap h, int type, u32 flags);
+static unixsock unixsock_alloc(heap h, int type, u32 flags, boolean alloc_fd);
 
 /* Called with lock acquired, returns with lock released. */
 static void unixsock_dealloc(unixsock s)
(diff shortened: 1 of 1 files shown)


src/unix/socket.c
```diff
@@ -496,13 +496,10 @@ closure_function(1, 2, sysreturn, unixsock_close,
...
-        /* Deallocate any sockets in the connection queue. */
+        /* Close any sockets in the connection queue. */
         unixsock child;
-        while ((child = dequeue(s->conn_q)) != INVALID_ADDRESS) {
-            unixsock_lock(child);
-            unixsock_dealloc(child);
-        }
-
+        while ((child = dequeue(s->conn_q)) != INVALID_ADDRESS)
+            fdesc_close_internal(t->p, &s->sock.f, s->sock.fd);
```
francescolavra, line 502: It's the child socket that should be closed, so the second and third arguments should be &child->sock.f and child->sock.fd.
An alternative approach to solve the problem would be to not allocate a file descriptor when a socket is enqueued in conn_q, but only if/when it's dequeued during an accept().

wjhun: Ok, I changed the code to defer allocating an fd until the enqueued socket is accepted. I removed fdesc_close_internal() as there is no need for it now.

francescolavra (commented):
Looks good to me

PR #1834, opened 2023-03-20
Http and user socket improvements

sanderssj:
This PR makes several improvements for web.c performance and compatibility by
updating web.c to use the http listener for better http support and then updating
the http.c code to support proper connection persistence based on the http
version of the request. The http response fuctions now take an http_responder pointer
that keeps track of whether or not to close the connection after the response and
automatically adds an appropriate connection header if necessary. These changes
allow apache-bench to work web.c, where previously it was timing out.

Also, the user socket code has been updated to set TCP_NODELAY after accepting
a connection. This drastically improves performance in web.c, particularly in a bridged
network situation, where the client's acks would take 20-40ms causing unnecessary
delays during tcp conversations.

### DIFF
--- a/src/http/http.c
+++ b/src/http/http.c
@@ -1,6 +1,10 @@
 #include <runtime.h>
 #include <http.h>
 
+#define HTTP_VER(x, y) (((x)<<16)|((y)&MASK(16)))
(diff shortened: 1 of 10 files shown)


wjhun (approved):
LGTM

PR #1837, opened 2023-04-04
syslog: ensure initial and final messages are delivered

### LINKED ISSUE #1836: syslog not reliably pushing logs to Rsyslog server
# Summary
I setup a basic rsyslog server to consume logs from unikernel instances in our production infrastructure.  I have noticed that on first start and when the unikernel crashes are completely missing from the rsyslog server.

specifically the "I crashed" logs
```
...
time="2023-03-31T18:22:19Z" level=fatal msg="disconnected from cluster"
...
```

and the start logs:

```
en1: assigned 10.0.1.7
{"level":"info","msg":"..."}
en1: assigned FE80::38:1BFF:FE70:68A9
```

basic configs are as you would want and once everything is up and stable, logs flow in. I just really need the system crash and start logs to give my dev team confidence in debugging a uniernel in  a production setting

francescolavra:
Since the network stack needs to do an initial ARP query in order to retrieve the MAC address to be used for packets with a given destination, if many such packets are sent to the same destination before a response to the ARP query is received, the ARP driver queues only a limited number (currently 3) of them. As a result, if the user application prints many messages in short succession during startup and expects these messages to be delivered to a syslog server, from the initial burst of messages, only the 3 most recent ones are delivered. To work around this issue, the first commit changes the syslog klib so that the number of UDP packets sent in the first batch does not exceed the size of the ARP queue.
The second commit amends the syslog klib so that any buffered messages are flushed when a kernel shutdown is requested. This fixes the issue of messages not delivered to the syslog destination if the user application prints one or more messages just before terminating.

Closes #1836

### DIFF
--- a/klib/syslog.c
+++ b/klib/syslog.c
@@ -58,6 +58,7 @@ static struct {
     buffer program;
     char *server;
     boolean dns_in_progress;
+    boolean arp_init;
     timestamp dns_backoff;
     timestamp dns_req_next;
     ip_addr_t server_ip;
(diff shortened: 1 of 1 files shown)


klib/syslog.c
```diff
@@ -327,11 +343,18 @@ define_closure_function(0, 2, void, syslog_timer_func,
...
+        syslog_file_flush(0);
         syslog_udp_flush();
     }
 }
 
+define_closure_function(0, 2, void, syslog_shutdown_completion,
+                        int, status, merge, m)
+{
+    syslog_file_flush(apply_merge(m));
+    syslog_udp_flush();
```
sanderssj, line 355: I think it's theoretically possible for the final completion from the merge (which shuts off the vm) to run before the udp flush, so maybe the udp flush should be first here?

francescolavra: The completion invoked by syslog_file_flush() can theoretically run before the udp flush, but the final completion from the merge won't run until after all shutdown completions are applied: kernel_shutdown() keeps the merge open until the call to apply(sh, STATUS_OK).

sanderssj: Right, I see. Looks good!

wjhun (approved):
LGTM

PR #1839, opened 2023-04-14
vendor(acpica): upgrade - R09_30_21 -> R03_31_23

rinor:


rinor:
@francescolavra is this smth we are interested in, or not worth it?

francescolavra:
> @francescolavra is this smth we are interested in, or not worth it?

I think we should do this change if it brings practical benefits, e.g. it adds a new feature to nanos or solves an existing issue, or decreases the kernel binary size.

rinor:
> I think we should do this change if it brings practical benefits, 

atm I don't have enough information/knowledge to provide an informed answer/feedback, however I do have some analysis:

> it adds a new feature to nanos

- I'm not aware of any **new public feature** that requires such upgrade

> solves an existing issue

- I'm not aware of any **existing public issue** that requires such upgrade

> decreases the kernel binary size

- stripped `kernel.img` size is **increased** by **88 bytes**
  - **R09_30_21** `1_666_080 bytes`
  - **R03_31_23** `1_666_168 bytes`

Nevertheless, cleaned up the changelog between the two releases, keeping only core and kernel-resident changes, so posting it here for convenience in case you can make better use of it.

<details><summary>Changes to <code>R03_31_23</code> since <code>R09_30_21</code></summary>

```
----------------------------------------
31 March 2023. Summary of changes for version 20230331:

0) Global changes:

Update all copyright notices to the year 2023. This effects all source 
modules, as well as utility signons.
Add OS-specific support for Zephyr RTOS.
Fix miscellaneous code that accesses various objects.
Remove some dead code.
Add C ?Flexible Array? support.
Add support for 64 bit LoongArch compilation.
Add first batch of RISC-V related definitions.
Performed a global automated update to remove tabs, fix indentation 
issues, etc.

1) ACPICA kernel-resident subsystem:

hwvalid: Drop port I/O validation.

----------------------------------------
20 October 2022. Summary of changes for version 20221020:

0) Global changes:

Allow disabling of -Werror. For distro maintainers having `-Werror` can 
delay update of GCC. Since every GCC release might add new warnings that 
were not yet captured, it might break the build of packages. With this 
change, distros can now build with `NOWERROR=TRUE` instead of patching 
either the errors or the makefiles. The default behavior keeps on using 
`-Werror`.

1) ACPICA kernel-resident subsystem:

Added support for FFH Operation Region special context data. FFH(Fixed 
Function Hardware) Opregion is approved to be added in ACPI 6.5 via code 
first approach[1]. It requires special context data similar to GPIO and 
Generic Serial Bus as it needs to know platform specific offset and 
length.

Reverted this commit "executer/exsystem: Warn about sleeps greater than 
10 ms." Due to user complaints about valid sleeps greater than 10ms seen 
in some existing machines -- generating lots of warnings.

Do not touch VGA memory when EBDA < 1KiB. The ACPICA code assumes that 
EBDA region must be at least 1KiB in size. Because this is not 
guaranteed, it might happen that while scanning the memory for RSDP 
pointer, the kernel touches memory above 640KiB. This is unwanted as the 
VGA memory range may not be decoded or even present when running under 
virtualization.

Check that EBDA pointer is in valid memory. If the memory at 0x40e is 
uninitialized, the retrieved physical memory address of EBDA may be 
beyond the low memory (i.e. above 640K). If so, the kernel may 
unintentionally access the VGA memory, that might not be decoded or even 
present in case of virtualization.

----------------------------------------
31 March 2022. Summary of changes for version 20220331:

0) Global changes:

Update all copyright notices to the year 2022. This effects all source 
modules, as well as utility signons.

1) ACPICA kernel-resident subsystem:

For the ASL Sleep() operator, issue a warning if the sleep value is 
greater than 10 Milliseconds. Quick boottime is important, so warn about 
sleeps greater than 10 ms. Distribution Linux kernels reach initrd in 350 
ms, so excessive delays should be called out. 10 ms is chosen randomly, 
but three of such delays would already make up ten percent of the 
boottime.

Namespace: Avoid attempting to walk the Namespace if the Namespace does 
not exist.

AML interpreter/iASL compiler: Add new Acpi 6.4 semantics for the 
LoadTable and Load operators. DDB_HANDLE is gone, now loadtable returns a 
pass/fail integer. Now load returns a pass/fail integer, as well as 
storing the return value in an optional 2nd argument.

Headers: Use uintptr_t and offsetof() in Linux kernel builds. To avoid 
"performing pointer subtraction with a null pointer has undefined 
behavior" compiler warnings, use uintptr_t and offsetof() that are always 
available during Linux kernel builds to define ACPI_UINTPTR_T and the 
ACPI_TO_INTEGER() and ACPI_OFFSET() macros when building the ACPICA code 
in the Linux kernel.

Added support for the Windows 11 _OSI string ("Windows 2021"). Submitted 
by superm1.

executer/exsystem: Inform users about ACPI spec violation for the Stall() 
operator. Values greater than 100 microseconds violate the ACPI 
specification, so warn users about it. From the ACPI Specification 
version 6.2 Errata A, 19.6.128 *Stall (Stall for a Short Time)*:
> The implementation of Stall is OS-specific, but must not relinquish
> control of the processor. Because of this, delays longer than 100
> microseconds must use Sleep instead of Stall.

----------------------------------------
17 December 2021. Summary of changes for version 20211217:

1) ACPICA kernel-resident subsystem:

Hardware: Do not flush CPU cache when entering S4 and S5. According to 
ACPI 6.4, Section 16.2, the CPU cache flushing is required on entering to 
S1, S2, and S3, but the ACPICA code flushes the CPU cache regardless of 
the sleep state. Blind cache flush on entering S5 causes problems for 
TDX.

Avoid subobject buffer overflow when validating RSDP signature. Since the 
Signature member is accessed through an ACPI_TABLE_HEADER, the pointer to 
it is only to a 4-char array, and so trying to read past the 4th 
character, as will be done when it is an RSDP, reads beyond the bounds of 
the accessed member. Contributed by jrtc27.

Add support for PCC Opregion special context data. PCC Opregion added in 
ACPIC 6.3 requires special context data similar to GPIO and Generic 
Serial Bus as it needs to know the internal PCC buffer and its length as 
well as the PCC channel index when the opregion handler is being executed 
by the OSPM. Adds support for the special context data needed by PCC 
Opregion. Submitted by Sudeep Holla <sudeep.holla@arm.com>
```
</details>

PR #1842, opened 2023-04-26
Disallow overwriting of program and interpreter files by default

francescolavra:
With this change, the user application by default is not allowed to overwrite the program (and interpreter, if present) binary file.
This default behavior can be overridden by inserting a "program_overwrite" attribute in the root tuple of the manifest.
In the write runtime test, the "exec_protection" flag has been removed in order to test that the executable file by default cannot be overwritten even without exec protection.
File access permissions are implemented in a generic way so that it is possible to mark an arbitrary file as read-only by inserting a "readonly" attribute in the file tuple in the manifest.
In addition, when exec protection is enabled (which protects the program file from overwriting, independently of this new feature), the kernel disallows overwriting of the interpreter file as well, if present.

### DIFF
--- a/src/kernel/stage3.c
+++ b/src/kernel/stage3.c
@@ -174,8 +174,7 @@ closure_function(6, 0, void, startup,
     value p = get(root, sym(program));
     assert(p);
     tuple pro = resolve_path(root, split(general, p, '/'));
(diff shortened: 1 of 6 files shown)


wjhun (approved):
LGTM

PR #1843, opened 2023-04-28
Unix: implement RELATIME (relaxed update of file access time)

francescolavra:
With this policy, when a file is accessed, the file access time is updated only if its current value is less than or equal to the file modification time (or if it is more than 1 day old). This is the default behavior on Linux, and makes file reads more efficient by avoiding frequent file metadata updates.
In addition, the file access time update is being removed from the filesystem_get_node() function, because the access time should be updated only when file contents are accessed (or, in the case of directories, when directory contents are listed).


wjhun (approved):
LGTM

PR #1844, opened 2023-04-28
Network sockets: implement sg_write closure

francescolavra:
This allows all data submitted by the program via a writev() syscall to be sent at once (instead of in as many chunks as the number of iovec entries) to the network stack, which improves efficiency, especially when using TCP sockets with the TCP_NODELAY option enabled (which would otherwise cause a TCP packet to be sent for each iovec entry).


sanderssj:
LGTM

sanderssj (approved):

PR #1846, opened 2023-05-02
vmxnet3 receive fixes and improvements

sanderssj:
In the vmxnet3 driver, When a new replacement buffer is added while the
current completion descriptor is being processed, it can sometimes cause the
receive descriptor to reset some of its fields, including the eop flag. This
can occur regularly once the rx ring has been run through once, and the effect
is that the eop flag gets reset to 0, which causes the packet in process to be
saved as if the first part of a large packet. Then, the next packet received
has the sop bit set, causing this packet to be discarded.

This behavior can cause up to half the received packets to be discarded,
which then causes retransmissions and increased retransmit times from
the client.

This change fixes the problem by delaying the replacement of the current
receive buffer until after the current receive completion descriptor has been
completely processed.

The rx ring size of the vmxnet3 has also been increased to help improve performance
in some cases.


francescolavra (commented):
Works for me!

PR #1847, opened 2023-05-05
vmxnet3: fix allocation size for rx queue completion descriptors

francescolavra:
The memory allocated for the rx queue completion ring was half the size required for the number of completion descriptors as set in the comp_ring_len field of the vmxnet3_rxq_shared struct. This was causing memory corruption as the vmxnet3 device would write past the memory range allocated for the completion ring.

In addition, this change set fixes the assert() statement in vmxnet3_receive() to check the rx queue ID.

Closes https://github.com/nanovms/ops/issues/1454


sanderssj:
LGTM!

wjhun (approved):
LGTM

PR #1848, opened 2023-05-05
vmxnet3: disable Large Receive Offload (LRO) feature

francescolavra:
If the LRO feature is enabled, when the vmwnet3 device coalesces multiple input TCP packets into a single packet, it does not update the checksum in the TCP header of the large packet (because it can indicate that all the checksums are correct by setting the appropriate fields in the completion descriptor); therefore, when the packet is processed by the lwIP TCP stack, the checksum calculation fails, and the packet is discarded. This causes a high performance penalty, since the application needs to wait for the discarded packet to be re-transmitted by the remote peer. On the other hand, checksum calculation by lwIP should be kept enabled because lwIP does not support disabling the calculation on a per-packet basis, and it is not possible for the driver to know beforehand whether the checksum for incoming packets will be calculated by the device.
To fix the above issue, this change disables the LRO feature; in addition, the RX checksum feature is also disabled because we cannot exploit it, since we cannot avoid the checksum calculation in lwIP.
The MTU size is being updated to match the size of the RX buffers (the vmxnet3 device discards any incoming packets larger than the configured MTU value).

### DIFF
--- a/src/vmware/vmxnet3_net.c
+++ b/src/vmware/vmxnet3_net.c
@@ -101,13 +101,13 @@ void init_vmxnet3_driver_shared(vmxnet3_pci dev)
 
     vmx_ds->vmxnet3_revision = VMXNET3_REVISION;
     vmx_ds->upt_version = VMXNET3_UPT_VERSION;
-    vmx_ds->upt_features = UPT1_F_CSUM | UPT1_F_LRO;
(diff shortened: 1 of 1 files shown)


sanderssj:
LGTM!

wjhun (approved):
LGTM

PR #1849, opened 2023-05-09
Aarch64: enable CPU cache operation from user program

francescolavra:
Frameworks such as Node.js expect to be able to retrieve CPU cache characteristics such as cache line sizes, and execute cache maintenance operations (e.g. flushing the instruction cache). This change modifies the settings of the SCTLR_EL1 register so that cache access instructions from the user program do not generate exceptions.


wjhun (approved):
LGTM

PR #1852, opened 2023-05-10
ESXi: add support for booting via UEFI

francescolavra:
This change set allows Nanos to run on ESXi instances configured to boot via UEFI instead of BIOS. Tested on ESXi v7.0.
Note: the VGA console does not work on UEFI instances, because the video device, although being VGA-compatible, is not configured in text mode by the UEFI firmware, and the the kernel VGA console driver supports text mode only. The serial console continues to work as before.

Credits go to @sanderssj for the GDT- and ACPI-related changes.


sanderssj (commented):
LGTM

sanderssj (approved):

PR #1855, opened 2023-05-17
X86_64: PVH entry point: add stack memory configuration

### LINKED ISSUE #1838: feat(vmm): cloud-hypervisor support
Was trying to play around with nanos on https://github.com/cloud-hypervisor/cloud-hypervisor , but it looks like there are some issues. Assuming it's not a local problem on my environment, still troubleshooting, do we know what is preventing nanos to work there? 

i.e:

```sh
cloud-hypervisor \
  --cpus "boot=1,max=1,max_phys_bits=48" \
  --memory "size=1024M" \
  --rng "src=/dev/urandom" \
  --kernel nanos/output/platform/pc/bin/kernel.img \
  --disk "path=/home/ops/.ops/images/ch-nanos,direct=on,id=nanos" \
  --cmdline "console=ttyS0 reboot=k panic=1 pci=off" \
  --net "offload_tso=off,offload_ufo=off,offload_csum=off,tap=,mac=,ip=,mask=" \
  --seccomp false \
  --console tty \
  --log-file ./ch-nanos.log \
  -v -v -v
```

While looking around stumbled across https://github.com/cloudius-systems/osv/commit/4fa14833629489bef538a8d40039d0b317edfc03, jic it's relevant in here.

francescolavra:
The pvh_start32 assembly code is a 32-bit entry point for the kernel, used by hypervisors such as QEMU (with the microvm machine type) and cloud-hypervisor (https://github.com/cloud-hypervisor/cloud-hypervisor). The current code was missing the initialization of the stack, which prevents the kernel from booting under cloud-hypervisor.
This change adds the code to set the stack pointer register and to map the memory area to be used as initial stack during boot; this allows booting under cloud-hypervisor.

Closes #1838.

### DIFF
--- a/src/x86_64/crt0.s
+++ b/src/x86_64/crt0.s
@@ -414,6 +414,10 @@ bits 32
         mov dword [edx + esi], ecx
         or dword [edx + esi], 0x83
         mov dword [edx + esi + 4], 0
+        ; set stack pointer to INITIAL_MAP_SIZE, and map stack memory
+        mov esp, 0xa000
+        mov dword [edx], 0x83
+        mov dword [edx + 4], 0
         ; map kernel code
         mov dword [edx + 8], 0x200000 | 0x83
         mov dword [edx + 12], 0


wjhun (approved):
LGTM

PR #1856, opened 2023-05-18
Aarch64 init_mmu(): call a non-locking version of map()

francescolavra:
When a new mapping is set up via the map() function, the page table spinlock is acquired in order to protect the page tables from concurrent access. On aarch64, spinlock operations are implemented by using the Load-Exclusive/Store-Exclusive mechanism; the Arm architecture reference manual (version H.a, section B2.9.2 "Exclusive access instructions and Shareable memory locations") says that support for this mechanism when address translation is disabled is not an architectural requirement, and architecturally compliant software must not rely on using it.
The init_mmu() function sets up some mappings before enabling the MMU (i.e. when address translation is disabled); in order for this code to run on machines where exclusive access instructions don't work without address translation (such as Apple M2, where the `ldxr` instruction triggers an exception), the page table lock must not be acquired. This change introduces a map_nolock() function, which is a variant of the map() function that does not acquire the page table lock; init_mmu() now calls map_nolock() instead of map(), which is safe because this code is only run during initialization by the boot CPU, when any additional CPUs have not been enabled yet.
This allows Nanos to run on Apple silicon with HVF acceleration.

### DIFF
--- a/src/aarch64/page.c
+++ b/src/aarch64/page.c
@@ -89,14 +89,14 @@ void init_mmu(range init_pt, u64 vtarget)
     page_init_debug("kernel_size ");
     page_init_debug_u64(kernel_size);
     page_init_debug("\n");
-    map(KERNEL_BASE, KERNEL_PHYS, kernel_size, pageflags_writable(pageflags_exec(pageflags_memory())));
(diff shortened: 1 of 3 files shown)


wjhun (approved):
LGTM

PR #1861, opened 2023-06-02
Page cache and filesystem fixes and improvements

### LINKED ISSUE #1851: node crashing on pagecache writes
hello! looks like there is a write issue. tried latest nanos and nightly:
```
INFO [05-09|23:53:39.332] Syncing beacon headers                   downloaded=13,731,840 left=3,494,356  eta=3m50.097s
ERROR[05-09|23:53:39.574] Empty cpu stats 
ERROR[05-09|23:53:42.577] Empty cpu stats 

frame trace: 
ffffc000027ffcd0:   ffffffff8008ac28    (extents_range_handler + 00000000000005f8/00000000000007bf)
ffffc000027ffdc0:   ffffffff8008af91    (filesystem_check_or_reserve_extent + 0000000000000051/00000000000000a9)
ffffc000027ffde0:   ffffffff8004de6b    (pagecache_write_sg + 00000000000000eb/000000000000054e)
ffffc000027fff00:   ffffffff800cbc43    (file_write + 0000000000000183/000000000000045b)
ffffc000027fff80:   ffffffff800ce9d9    (write + 0000000000000189/0000000000000277)
ffffc000027fffb0:   ffffffff800d758a    (syscall_handler + 00000000000002ca/00000000000005b9)

loaded klibs: 
assertion oldval failed at /home/circleci/project/src/tfs/tfs.c:877  in update_extent(); halt
```

If I can provide anymore information please lmk and I will be glad to

francescolavra:
This change set fixes a few issues that have been identified in the page cache and TFS code, and improves memory handling so that the kernel is able to operate correctly under memory pressure.
Closes #1851.

Issues fixed:
- refcounting of pagecache pages was leaking a reference when the touch_or_fill_page_nodelocked() function is called with a
null merge argument for a non-filled page 
- the range to be allocated on the TFS storage heap when extending a file extent was calculated incorrectly, thereby wasting a portion of the allocated range
- refcounting of TFS log structures was implemented in a way that prevents theses structures from being deallocated after a log is retired
- the log compaction code now prevents triggering a new compaction when the previous compaction is not complete

New features:
- the id heap being used as TFS storage heap has been replaced with a rangemap-based heap, for improved memory efficiency on instances with large volumes
- the page cache draining implementation is now able to deallocate not only the page-sized memory chunks that make up the cache, but also their associated structures that link pages to the other pagecache data structures, for improved scalability when running applications that create large files and/or a large amount of files
- the free memory threshold used to trigger memory cleaning is now calculated dynamically based on the total amount of physical memory available in an instance


src/runtime/range.c
```diff
@@ -90,6 +90,39 @@ boolean rangemap_insert_range(rangemap rm, range r)
...
+                if (n == INVALID_ADDRESS)
+                    return false;
+                rmnode_init(n, irange(r.end, curr->r.end));
+                curr->r.end = r.start;
+                rangemap_insert(rm, n);
+            } else {
+                curr->r.end = r.start;
+            }
+        } else if (curr->r.end > r.end) {
+            curr->r.start = r.end;
```
wjhun, line 117: Changing curr->r should a rangemap_reinsert(). In this particular case, it's possible that the adjustment of start would not actually cause any reordering in the underlying data structure (rbtree), but this isn't a guarantee at the interface level.

Similarly, changing curr->r.end above should not effect a change in the tree - but, again, there could be subtle breakage should the rangemap implementation be changed in the future.

src/kernel/pagecache.c
```diff
@@ -317,7 +317,8 @@ closure_function(3, 1, void, pagecache_read_page_complete,
     closure_finish();
 }
 
-/* The page refcount is incremented (unless a free page could not be re-allocated). */
+/* The page refcount is incremented (unless the page is not filled and either a read request is not
+ * ongoing, or the merge argument is null. */
```
wjhun, line 321: Nit: parens

wjhun (approved):
Looks good ... a few notes

src/kernel/pagecache.c
```diff
@@ -398,14 +398,28 @@ static void pagecache_page_release_locked(pagecache pc, pagecache_page pp)
     u64 pre = fetch_and_add(&pc->total_pages, -1);
     assert(pre > 0);
     pagecache_debug("%s: total pages now %ld\n", __func__, pre - 1);
+#ifdef KERNEL
+    if (full_delete) {
```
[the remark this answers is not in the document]
francescolavra: How would you implement a mem_cleaner callback for this? If we walk the list of free pages with the pagecache state lock held, we can't safely lock the node of a given page, for the same reason why we have a spin_try() here. On the other hand, we don't have a global list of pagecache_nodes; even if we had one, it seems inefficient to walk through all the nodes in the cache (and walk through all pages in each node) to find page structs that can be deallocated.

PR #1862, opened 2023-06-02
TFS create_extent(): fix assignment of file range

francescolavra:
When a request to create an extent of a certain size cannot be fulfilled due to missing free contiguous storage space, and the extent is created with a smaller size than requested, the existing code is assigning the originally requested file range (instead of the range actually allocated) to the extent being created, which leads to filesystem corruption.
Regression introduced in commit 65434ad7 (#1765).


sanderssj (approved):
LGTM

PR #1863, opened 2023-06-02
HyperV netvsc_m_append(): fix setup of pbuf length

francescolavra:
The maximum payload length for a struct pbuf allocated when receiving a packet in the netvsc driver is given by the `rxbuflen` field of the hn_softc_t struct, without adding sizeof(struct xpbuf).


sanderssj (approved):
LGTM

PR #1865, opened 2023-06-07
netlink RTM_GETADDR: add support for AF_INET6 and AF_UNSPEC

wjhun:
This adds netlink support for reporting IPv6 interface addresses via the RTM_GETADDR request. It also handles the AF_UNSPEC address family by reporting both IPv4 and IPv6 addresses.


test/runtime/netlink.c
```diff
@@ -287,16 +287,19 @@ static void test_getaddr(void)
...
-
+    const int afs[3] = { AF_INET, AF_INET6, AF_UNSPEC };
+    for (int i = 0; i < 3; i++) {
+        req.msg.rtgen_family = afs[i];
+        memcpy(iov.iov_base, &req, sizeof(req));
+        iov.iov_len = sizeof(req);
+        ret = sendmsg(fd, &msg, 0);
+        test_assert(ret == sizeof(req));
+        iov.iov_len = sizeof(buf);
+        recv_resp(fd, &msg, sizeof(struct ifaddrmsg), RTM_NEWADDR, NLM_F_MULTI);
```
francescolavra, line 298: This makes the test fail when run on Linux, with "Error: ret >= NLMSG_LENGTH(len) -- failed at /usr/src/nanos/test/runtime/netlink.c:68".
Specifically, for multipart messages such as the response to this RTM_GETADDR request there is no guarantee that all chunks are received in a single recvmsg() call: in my Linux box, the NLMSG_DONE chunk is not included in the data received by the first recvmsg() call, and is returned in the following call, which makes the test fail at the second iteration of the for() loop.
One way to get around this issue would be to use different netlink sockets to make different requests.

wjhun: Ok. I changed the loop to open and close a netlink socket for each request. The test now passes under Linux.

francescolavra (commented):
LGTM

PR #1868, opened 2023-06-09
Sockets/eventfd: fix memory leaks

francescolavra:
This change set fixes memory leaks occurring when some types of file descriptors (network and Unix domain sockets and eventfd) are closed and when network packets are sent on a TCP socket.
New macros have been defined to allow attaching different functions to different instantiations of a closure, and some heap-allocated closures have been converted to closures (declared is in unix_internal.h but instantiated with file descriptor-specific functions) embedded in other structs.


francescolavra:
In the last commit, added a missing deallocation of struct udp_entry in socket_close().

Issue #1870, opened 2023-06-11
investigate(build):  warnings - gcc-12.2, ld-2.40


rinor:
Starting with release 2.39 of the GNU Binutils several new warnings have been added to the linker.
Also gcc-12 may have new rules/checks.

In order to continue building nanos successfully ` -Wno-error=stringop-overflow -Wno-error=maybe-uninitialized` may need to be added to build flags. 

Posting all warnings, after adding those build flags, just to check if there is anything that's relevant and needs to be handled on code.

- **tools**
```
$ ld --version
GNU ld (GNU Binutils for Debian) 2.40
...
$ gcc --version
gcc (Debian 12.2.0-14) 12.2.0
```

- **build logs**
```
$ make                
...
ld: warning: /ops/nanos/output/klib/bin/firewall.dbg has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/klib/bin/gcp.dbg has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/klib/bin/ntp.dbg has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/klib/bin/radar.dbg has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/klib/bin/test/page_table.dbg has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/platform/pc/boot/platform/pc/boot/service32.o: missing .note.GNU-stack section implies executable stack
ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
ld: warning: /ops/nanos/output/platform/pc/boot/bin/stage2.elf has a LOAD segment with RWX permissions
...
ld: warning: /ops/nanos/output/platform/pc/boot/uefi/x86_64/uefi-crt0.o: missing .note.GNU-stack section implies executable stack
ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
...
CC	/ops/nanos/output/platform/pc/src/unix/aio.o
In file included from /ops/nanos/src/runtime/runtime.h:3,
                 from /ops/nanos/src/kernel/kernel.h:2,
                 from /ops/nanos/src/unix/unix_internal.h:3,
                 from /ops/nanos/src/unix/aio.c:1:
In function ‘fetch_and_add’,
    inlined from ‘refcount_reserve’ at /ops/nanos/src/runtime/refcount.h:15:5,
[...]

PR #1872, opened 2023-06-12
fix warnings when compiling with gcc 12

wjhun:
This addresses some of the compiler warnings found in #1870.


francescolavra (commented):
LGTM

PR #1873, opened 2023-06-14
Real-time clock: add support for running without an RTC peripheral

### LINKED ISSUE #1869: Firecracker incorrect clock
I'm not sure how to debug this, there aren't many docs about firecracker usage, but the clock is very wrong when using firecracker.

It appears to be stuck on Aug 16th 2021, on boot. I can't figure out why exactly. Firecracker supports KVM clock (pv), but I don't know if nanos is detecting it.

```
❯ ops profile
Ops version: 0.1.37
Nanos version: 0.0
Qemu version: 8.0.0
Arch: linux
Virtualized: false
```

```
❯ firecracker --version
Firecracker v0.24.6
```

Using the C program defined here: https://github.com/nanovms/nanos/issues/357#issuecomment-464216122

I get:

```
❯ ops build hello -c ops.json --ip-address 10.0.2.2 --type firecracker && sudo rm /tmp/firecracker.socket; sudo firecracker --config-file fc.json --api-sock /tmp/firecracker.socket
 100% |████████████████████████████████████████|         
Bootable image file:/home/jerome/.ops/images/hello
warning: ACPI MADT not found, default to 1 processor
en1: assigned 10.0.2.2
Boot time   = Mon Aug 16 00:00:00 2021

Monotonic Time  = Thu Jan  1 00:00:00 1970
Current (boot)  = 1629072000
Current (mono)  = 0
Current time    = Mon Aug 16 00:00:00 2021
Real Time       = Mon Aug 16 00:00:00 2021

en1: assigned FE80::A8FC:FF:FE00:1
Monotonic Time  = Thu Jan  1 00:00:02 1970
Current (boot)  = 1629072000
Current (mono)  = 2
Current time    = Mon Aug 16 00:00:02 2021
Real Time       = Mon Aug 16 00:00:02 2021

Monotonic Time  = Thu Jan  1 00:00:04 1970
Current (boot)  = 1629072000
Current (mono)  = 4
Current time    = Mon Aug 16 00:00:04 2021
Real Time       = Mon Aug 16 00:00:04 2021

Monotonic Time  = Thu Jan  1 00:00:06 1970
Current (boot)  = 1629072000
Current (mono)  = 6
Current time    = Mon Aug 16 00:00:06 2021
Real Time       = Mon Aug 16 00:00:06 2021
```

<details>
<summary>fc.json</summary>

```json
{
    "boot-source": {
      "kernel_image_path": "/home/jerome/.ops/0.1.43/kernel.img",
      "boot_args": "console=ttyS0 reboot=k panic=1 pci=off random.trust_cpu=on"
    },
    "drives": [
      {
        "drive_id": "rootfs",
        "path_on_host": "/home/jerome/.ops/images/hello",
        "is_root_device": true,
        "is_read_only": false
      }
    ],
    "network-interfaces": [
      {
        "iface_id": "eth0",
        "guest_mac": "AA:FC:00:00:00:01",
        "host_dev_name": "tap0"
      }
    ],
    "machine-config": {
      "vcpu_count": 1,
      "mem_size_mib": 512,
      "ht_enabled": false
    },
    "logger": {
        "log_path": "logs.fifo",
        "level": "info"
      }
  }
```
</details>

<details>
<summary>ops.json</summary>

```json
{
    "Env": {
      "PORT": "8080"
    },
    "Klibs": ["ntp.dbg"],
    "ManifestPassthrough": {
      "ntp_servers": ["0.us.pool.ntp.org:123"],
      "ntp_poll_min": "4",
      "ntp_poll_max": "6"
    }
  }
```
</details>

I'm only trying the `ntp` klib as an attempt to fix it, but that doesn't help either.

A few questions:
- Is there a way to debug this in a firecracker environment? I couldn't get any of the debugging examples to work for me in there.
- What is the list of supported `boot_args` for the kernel?

francescolavra:
Nanos can run on platforms where the RTC peripheral is read-only (e.g. riscv-virt), is not initialized with the system time from the host (e.g. cloud-hypervisor), or is not available at all (e.g. AWS Firecracker). This change set allows the system time to be retrieved during boot from an alternative source (e.g. the wall clock time made available by the paravirtualized clock), and allows changing the system time even when the RTC peripheral is unavailable (in this case, the time update will not persist across a VM reset).
The register_platform_clock_now() function now takes an additional parameter with the value of the offset between the system time and the monotonic time: if such offset is available (i.e. non-zero), it is used to initialize the rtc_offset field of the VDSO structure, otherwise the kernel falls back to trying to derive the offset from the RTC time. On platforms with the pvclock peripheral, the RTC offset is available via the wall clock time.
The x86 RTC driver is being amended to be able to handle cases where the RTC is not available: in these cases, rtc_gettimeofday() returns 0, and rtc_settimeofday() is a no-op.

Closes #1869

### DIFF
--- a/src/kernel/pvclock.c
+++ b/src/kernel/pvclock.c
@@ -5,6 +5,19 @@
 BSS_RO_AFTER_INIT static heap pvclock_heap;
 BSS_RO_AFTER_INIT static volatile struct pvclock_vcpu_time_info *vclock;
 
+static timestamp pvclock_get_rtc_offset(volatile struct pvclock_wall_clock *wc)
+{
+    u32 version;
+    timestamp rtc_offset;
+    do {
+        version = wc->version & ~1;
+        read_barrier();
+        rtc_offset = seconds(wc->sec) + nanoseconds(wc->nsec);
+        read_barrier();
+    } while (version != wc->version);
+    return rtc_offset;
+}
+
 u64 pvclock_now_ns(void)
 {
     return vdso_pvclock_now_ns(vclock);
@@ -15,12 +28,13 @@ closure_function(0, 0, timestamp, pvclock_now)
     return nanoseconds(pvclock_now_ns());
 }
 
-void init_pvclock(heap h, struct pvclock_vcpu_time_info *vti)
+void init_pvclock(heap h, struct pvclock_vcpu_time_info *vti, struct pvclock_wall_clock *wc)
 {
     assert(vti);
     vclock = vti;
     pvclock_heap = h;
-    register_platform_clock_now(closure(h, pvclock_now), VDSO_CLOCK_PVCLOCK);
+    register_platform_clock_now(closure(h, pvclock_now), VDSO_CLOCK_PVCLOCK,
+                                pvclock_get_rtc_offset(wc));
 }
 
 physical pvclock_get_physaddr(void)
--- a/src/x86_64/rtc.c
+++ b/src/x86_64/rtc.c
@@ -15,6 +15,8 @@
 #define RTC_YEAR 0x09
 #define RTC_STATUS_A 0x0a
 #define RTC_UIP (1 << 7)
+#define RTC_STATUS_D 0x0d
+#define RTC_RAM_POWER   (1 << 7)
 
 static unsigned int bcd2bin(unsigned int bcd) {
     return ((bcd >> 4) & 0x0f) * 10 + (bcd & 0x0f);
@@ -36,6 +38,8 @@ static void rtc_write(u8 reg, u8 val) {
 }
 
 u64 rtc_gettimeofday(void) {
+    if (rtc_read(RTC_STATUS_D) != RTC_RAM_POWER)    /* RTC not available */
+        return 0;
     while (rtc_read(RTC_STATUS_A) & RTC_UIP) {
         continue;
     }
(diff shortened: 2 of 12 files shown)


rinor:
tested on both firecracker/cloud-hypervisor and works as expected. Thank you.

src/kernel/pvclock.c
```diff
@@ -5,6 +5,19 @@
 BSS_RO_AFTER_INIT static heap pvclock_heap;
 BSS_RO_AFTER_INIT static volatile struct pvclock_vcpu_time_info *vclock;
 
+static timestamp pvclock_get_rtc_offset(struct pvclock_wall_clock *wc)
```
wjhun, line 8: Maybe it doesn't matter because of the read_barriers acting as compiler barriers, but it seems that the struct at wc should be treated as volatile.

wjhun (approved):
LGTM

src/x86_64/rtc.c
```diff
@@ -49,14 +53,17 @@ u64 rtc_gettimeofday(void) {
...
 
 void rtc_settimeofday(u64 seconds) {
+    u64 rtc_seconds = rtc_gettimeofday();
+    if (!rtc_seconds)   /* RTC not available */
+        return;
+
     /* Fix any possible inconsistencies due to RTC register concurrent updates by looping until the
      * timestamp read from the RTC equals the timestamp being written. */
-    while (rtc_gettimeofday() != seconds) {
+    while (rtc_seconds != seconds) {
```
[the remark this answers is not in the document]
francescolavra: OK, I added a limit on the number of retries.

src/kernel/pvclock.c
francescolavra: OK, changed the parameter of this function to `volatile struct pvclock_wall_clock *wc`

PR #1878, opened 2023-06-16
RTM_GETADDR: report multiple ipv6 addresses

wjhun:
This changes the behavior of nl_rtm_getaddr6() to iterate through all IPv6 addresses on an interface and enqueue an RTM_NEWADDR entry for each valid one.


francescolavra (commented):
LGTM

PR #1879, opened 2023-06-16
Klibs: gcp: add support for sending metrics to monitoring service

francescolavra:
This change set enhances the gcp klib so that it can now be configured for sending memory usage metrics to the GCP monitoring service, thus emulating the GCP ops agent.
The ID of the running instance and the zone where it is running (which are necessary to be able to send API requests to the
monitoring server) are retrieved from the instance metadata server.
The metrics being sent are related to memory usage (see https://cloud.google.com/monitoring/api/metrics_opsagent#agent-memory), and more specifically are the `bytes_used` and `bytes_percent` metric types; for each type, a value is sent for each of the "cached", "free" and "used" states.
Sending of memory metrics is enabled by inserting a "metrics" tuple in the "gcp" configuration tuple; it is possible to modify the time interval at which metrics are sent by inserting an "interval" attribute in the "metrics" tuple, with the interval value expressed
in seconds; the default (and minimum allowed) interval value is 60 seconds.

In order for the gcp klib to retrieve the appropriate credentials needed to communicate with the GCP monitoring server, the instance on which it runs must be associated to a service account (see https://cloud.google.com/compute/docs/access/service-accounts). Instances created by Ops are associated to a service account via the CloudConfig.InstanceProfile configuration parameter.

Example Ops configuration to enable sending memory metrics every 2 minutes:
```
{
    "CloudConfig" :{
        "ProjectID" :"my-proj",
        "Zone": "us-west1-a",
        "BucketName":"my-bucket",
        "InstanceProfile":"default"
    },
    "RunConfig": {
        "Klibs": ["gcp", "tls"]
    },
  "ManifestPassthrough": {
    "gcp": {
      "metrics": {"interval":"120"}
    }
  }
}
```

### DIFF
--- a/klib/gcp.c
+++ b/klib/gcp.c
@@ -3,12 +3,15 @@
 #include <http.h>
 #include <lwip.h>
 #include <mktime.h>
+#include <pagecache.h>
 #include <tls.h>
 
 #define GCP_MD_SERVER_ADDR  IPADDR4_INIT_BYTES(169, 254, 169, 254)
 
 #define GCP_LOG_SERVER_NAME "logging.googleapis.com"
 #define GCP_LOG_MAX_ENTRIES 8192
+
+#define GCP_METRICS_SERVER_NAME "monitoring.googleapis.com"
 
 declare_closure_struct(1, 1, void, gcp_setup_complete,
                        timestamp, retry_backoff,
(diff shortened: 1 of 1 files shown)


sanderssj (approved):
LGTM

PR #1880, opened 2023-06-21
Signal user program when receiving shutdown notification

sanderssj:
When an external shutdown notification is received, such as the acpi poweroff event, nanos
should signal the user program to exit before shutting down to allow it time to quiesce. This
PR implements this with a new unix_shutdown() function that sends a SIGTERM to the user
process. Programs that handle SIGTERM can perform shutdown tasks for up to 
UNIX_SHUTDOWN_TIMEOUT_SECS seconds, and programs without a SIGTERM handler will
be shut down immediately. The kernel_shutdown() is called when the user program exits or
when the timer expires.

Support has also been added to Xen for receiving shutdown notifications so the unix_shutdown
can be performed there as well.

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -644,6 +644,40 @@ out:
     return INVALID_ADDRESS;
 }
 
+closure_function(1, 2, void, shutdown_timeout,
+                 struct timer, shutdown_timer,
+                 u64, expiry, u64, overruns)
+{
+    closure_finish();
+    if (overruns == timer_disabled)
+        return;
+    if (!shutting_down)
+        kernel_shutdown(0);
+}
+
+void unix_shutdown(void)
+{
+    heap h = heap_locked(get_kernel_heaps());
+    /* Skip kernel unix process for shutdown */
+    for (int i = 1; i < MAX_PROCESSES; i++) {
+        process p = processes[i];
+        if (p == 0)
+            continue;
+        struct siginfo *si = allocate_zero(h, sizeof(struct siginfo));
+        if (si == INVALID_ADDRESS) {
+            kernel_shutdown(0);
+            return;
+        }
+        si->si_signo = SIGTERM;
+        deliver_signal_to_process(p, si);
+    }
+    struct timer shutdown_timer = {0};
+    init_timer(&shutdown_timer);
+    timer_handler th = closure(h, shutdown_timeout, shutdown_timer);
+    register_timer(kernel_timers, &closure_member(shutdown_timeout, th, shutdown_timer), 
+        CLOCK_ID_MONOTONIC, seconds(UNIX_SHUTDOWN_TIMEOUT_SECS), false, 0, th);
+}
+
 void program_set_perms(tuple root, tuple prog)
 {
     if (get(root, sym(exec_protection)))
--- a/src/xen/xen.c
+++ b/src/xen/xen.c
@@ -27,6 +27,12 @@
 declare_closure_struct(0, 1, void, xen_watch_handler,
                        const char *, path);
 declare_closure_struct(0, 0, void, xen_scan_service);
+declare_closure_struct(0, 1, void, xen_shutdown_watcher,
+                       const char *, path);
+declare_closure_struct(1, 0, void, xen_shutdown_handler,
+                       buffer, b);
+
+status xenstore_watch(buffer path, xenstore_watch_handler handler, boolean watch);
 
 typedef struct xen_platform_info {
     heap    h;                  /* general heap for internal use */
@@ -50,6 +56,8 @@ typedef struct xen_platform_info {
 
     closure_struct(xen_watch_handler, watch_handler);
     closure_struct(xen_scan_service, scan_service);
+    closure_struct(xen_shutdown_watcher, shutdown_watcher);
+    closure_struct(xen_shutdown_handler, shutdown_handler);
     u64 scanning;
 
     /* grant table */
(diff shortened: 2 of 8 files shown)


src/xen/xen.c
```diff
@@ -780,14 +819,9 @@ status xenstore_sync_printf(u32 tx_id, buffer path, const char *node, const char
 
 status xenstore_read_u64(u32 tx_id, buffer path, const char *node, u64 *result)
 {
-    buffer request = allocate_buffer(xen_info.h, 64);
-    assert(push_buffer(request, path));
-    push_u8(request, '/');
-    assert(buffer_write(request, node, runtime_strlen(node)));
-    push_u8(request, 0);
-
+    *result = 0;
```
francescolavra, line 822: We can remove this (the result variable should not be touched in case of an error)

src/unix/unix.c
```diff
@@ -644,6 +644,42 @@ process init_unix(kernel_heaps kh, tuple root, filesystem fs)
...
+            continue;
+        struct siginfo *si = allocate_zero(h, sizeof(struct siginfo));
+        if (si == INVALID_ADDRESS) {
+            kernel_shutdown(0);
+            return;
+        }
+        si->si_signo = SIGTERM;
+        process_lock(p);
+        deliver_signal_to_process(p, si);
+        process_unlock(p);
```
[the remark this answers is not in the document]
sanderssj: Agreed, I have removed the lock.

src/xen/xen.c
```diff
@@ -379,6 +387,33 @@ closure_function(1, 0, void, xen_per_cpu_init,
...
+define_closure_function(1, 0, void, xen_shutdown_handler,
+                 buffer, b)
+{
+    buffer rsp = bound(b);
+    buffer_clear(rsp);
+    u32 txid;
+    status s = xenstore_transaction_start(&txid);
+    if (!is_ok(s))
+        return;
+    s = xenstore_read_string(txid, alloca_wrap_cstring("control"), "shutdown", rsp);
```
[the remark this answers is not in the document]
sanderssj: Interestingly, gcc 11 doesn't complain about this one, but does complain about uninitialized result in xenstore_read_u64 below, which is why result is initialized there. I added tx_id initialization in xenstore_transaction_start instead of changing its callers.

src/xen/xen.c
sanderssj: This line is there to eliminate a compiler uninitialized warning in the callers.

rinor:
Will test in a few, but just curious, is this expected to work also on `cloud-hypervisor`, or additional handling is needed to make it work?

On `firecracker` this won't work for already known reasons mentioned also on theirs docs
> - https://github.com/firecracker-microvm/firecracker/blob/main/FAQ.md#how-can-i-gracefully-reboot-the-guest-how-can-i-gracefully-poweroff-the-guest
>
> ### How can I gracefully reboot the guest? How can I gracefully poweroff the guest?
>
> Firecracker does not implement ACPI and PM devices, therefore operations like
gracefully rebooting or powering off the guest are supported in unconventional ways.
>
> Running the `poweroff` or `halt` commands inside a Linux guest will bring it
down but Firecracker process remains unaware of the guest shutdown so it lives
on.
>
> Running the `reboot` command in a Linux guest will gracefully bring down the guest
system and also bring a graceful end to the Firecracker process.
>
> On `x86_64` systems, issuing a `SendCtrlAltDel` action command through the
Firecracker API will generate a `Ctrl + Alt + Del` keyboard event in the guest
which triggers a behavior identical to running the `reboot` command. This is,
however, not supported on `aarch64` systems.

francescolavra:
> Will test in a few, but just curious, is this expected to work also on `cloud-hypervisor`, or additional handling is needed to make it work?

Yes, I expect it to work also with cloud-hypervisor. According to their docs, there are 2 ways to trigger an ACPI power button event: one is via the [REST API](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/api.md) (/vm.power-button endpoint), and the other is via [ch-remote](https://www.cloudhypervisor.org/docs/prologue/commands/#ch-remote-binary) (power-button subcommand)

rinor:
> Yes, I expect it to work also with cloud-hypervisor. According to their docs, there are 2 ways to trigger an ACPI power button event: one is via the [REST API](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/api.md) (/vm.power-button endpoint), and the other is via [ch-remote](https://www.cloudhypervisor.org/docs/prologue/commands/#ch-remote-binary) (power-button subcommand)

Can't detect anything, even activating debug on ACPI nothing happens (`log wise`) when `vm.power-button` is called

```go
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"
)

func main() {
	sigs := make(chan os.Signal, 1)
	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGINT, syscall.SIGSEGV)
	done := make(chan bool, 1)

	go func() {
		sig := <-sigs
		fmt.Printf("...got signal: %q\n", sig)
		done <- true
	}()

	fmt.Println("waiting for signal...")
	<-done
	fmt.Println("exiting")
}

```

src/xen/xen.c
francescolavra: OK, I see the uninitialized warning in the callers. This only happens for callers in xen.c, because apparently gcc inlines these calls but somehow doesn't detect that an uninitialized use cannot happen. But adding this line breaks the logic in xenblk_watch_service(), where the sector_size variable is used to read first "sector-size" and then "physical-sector-size": if the second read results in an error, the sector_size variable is now reset to 0, which causes xbd->capacity to be set to 0. So we should either put the initialization in the callers inside xen.c, or make sure that all callers don't rely on the result variable not being touched in case of an error. The same goes for xenstore_transaction_start().

francescolavra:
> Can't detect anything, even activating debug on ACPI nothing happens (`log wise`) when `vm.power-button` is called

I just tested the power-button command on cloud-hypervisor, and indeed the button event is not detected by the kernel. But this is a separate issue, independent from this SIGTERM feature. We'll take a look.

src/xen/xen.c
sanderssj: Thanks, I missed that usage before. I took out the initialization in those functions and initialized at the caller instead, but at least it seems to only be needed in xen.c.

francescolavra:
LGTM

nielsbergsma:
I verified the change this morning on GCP (through manual termination, instance template update through instance groups), and it works great! thanks for adding this feature so quickly.

francescolavra:
> I just tested the power-button command on cloud-hypervisor, and indeed the button event is not detected by the kernel

#1893 adds support for the ACPI power button on cloud-hypervisor, which allows the kernel to perform a clean VM shutdown.

rinor:
tested and works as expected. Thank you
- output of https://github.com/nanovms/nanos/pull/1880#issuecomment-1602289590
```sh
awaiting signal
en1: assigned FE80::A8C6:FF:FE00:1
got signal: "terminated"
exiting
```
- triggered by `curl --unix-socket ./ch-api.socket -i -X PUT 'http://localhost/api/v1/vm.power-button'`
```sh
curl --unix-socket ./ch-api.socket -i -X PUT 'http://localhost/api/v1/vm.power-button'

HTTP/1.1 204 
Server: Cloud Hypervisor API
Connection: keep-alive
```

PR #1881, opened 2023-06-21
socket_write_tcp_bh(): fix use of uninitialized variable tcp_lw

francescolavra:
This change fixes a regression introduced in 0e67b81ef8406d15d5eec0f6eafdda95d0f9cba8, which causes tcp_unref() to be called with an uninitialized argument if the `out_unlock` goto target is reached.


sanderssj (approved):
LGTM

PR #1882, opened 2023-06-21
tfs_rename(): fix writing of metadata to TFS log

francescolavra:
A directory entry corresponds to an attribute in the `children` tuple of the directory tuple, not in the directory tuple itself. This change fixes an error that causes a filesystem corruption when a file is renamed (this corruption is only detected after a reboot).
Regression introduced in a4afaa317f88e1406432f5dc691fb6b9e448b025


wjhun (approved):
LGTM

PR #1883, opened 2023-06-24
trim cache parameters for low memory operation

wjhun:
This change reduces baseline memory requirements for various memory caches
under "low memory" conditions. If the function is_low_memory_machine() returns
true, then the following parameters are minimized:

- the pagecache will use PAGECACHE_LOWMEM_CONTIGUOUS_PAGESIZE (default 128KB)
  as its objcache pagesize instead of PAGESIZE_2M
- the mcache high order is lowered to 11 (2KB), thus forcing PAGESIZE or
  larger allocations to be made directly from the backing heap - and thus not
  cached, nor available for situations that implement a malloc/free interface
- similiarly, allocations for lwIP are limited to a maximum size of 2KB

In addition:
- the runtime tests are stripped by default, saving the unstripped binaries
  under the ".dbg" extension
- in virtio-net, the size of the backing page allocations for the rx buffer and
  tx completion caches is set according to the rx and tx queue sizes

With these changes, it is possible to run a nanos instance capable of running the
'webs' web server in as little as 16MB of system memory.

### DIFF
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -75,9 +75,10 @@ void init_kernel_heaps(void)
 #endif
     assert(heaps.linear_backed != INVALID_ADDRESS);
 
-    bytes pagesize = is_low_memory_machine(&heaps) ?
-                     U64_FROM_BIT(MAX_MCACHE_ORDER + 1) : PAGESIZE_2M;
-    heaps.general = allocate_mcache(&bootstrap, (heap)heaps.linear_backed, 5, MAX_MCACHE_ORDER,
+    boolean is_lowmem = is_low_memory_machine();
+    int max_mcache_order = is_lowmem ? MAX_LOWMEM_MCACHE_ORDER : MAX_MCACHE_ORDER;
+    bytes pagesize = is_lowmem ? U64_FROM_BIT(max_mcache_order + 1) : PAGESIZE_2M;
+    heaps.general = allocate_mcache(&bootstrap, (heap)heaps.linear_backed, 5, max_mcache_order,
                                     pagesize);
     assert(heaps.general != INVALID_ADDRESS);
 
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -12,7 +12,6 @@
 
 BSS_RO_AFTER_INIT static heap lwip_heap;
 BSS_RO_AFTER_INIT int (*net_ip_input_filter)(struct pbuf *pbuf, struct netif *input_netif);
-BSS_RO_AFTER_INIT int lwip_alloc_order;
 
 declare_closure_struct(0, 2, void, net_timeout_handler, u64, expiry, u64, overruns);
 
@@ -86,9 +85,6 @@ void lwip_debug(char * format, ...)
 
 void *lwip_allocate(u64 size)
 {
-    /* To maintain the malloc/free interface with mcache, allocations must stay
-       within the range of objcaches and not fall back to parent allocs. */
-    assert(size <= U64_FROM_BIT(lwip_alloc_order));
     void *p = allocate(lwip_heap, size);
     return ((p != INVALID_ADDRESS) ? p : 0);
 }
@@ -390,7 +386,7 @@ void init_net(kernel_heaps kh)
     heap h = heap_general(kh);
     heap backed = (heap)heap_linear_backed(kh);
     boolean is_lowmem = is_low_memory_machine();
-    lwip_alloc_order = is_lowmem ? MAX_LOWMEM_LWIP_ALLOC_ORDER : MAX_LWIP_ALLOC_ORDER;
+    int lwip_alloc_order = is_lowmem ? MAX_LOWMEM_LWIP_ALLOC_ORDER : MAX_LWIP_ALLOC_ORDER;
     bytes pagesize = is_lowmem ? U64_FROM_BIT(lwip_alloc_order + 1) : PAGESIZE_2M;
     lwip_heap = allocate_mcache(h, backed, 5, lwip_alloc_order, pagesize);
     assert(lwip_heap != INVALID_ADDRESS);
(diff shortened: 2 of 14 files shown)


src/kernel/init.c
```diff
@@ -75,9 +75,10 @@ void init_kernel_heaps(void)
 #endif
     assert(heaps.linear_backed != INVALID_ADDRESS);
 
-    bytes pagesize = is_low_memory_machine(&heaps) ?
-                     U64_FROM_BIT(MAX_MCACHE_ORDER + 1) : PAGESIZE_2M;
-    heaps.general = allocate_mcache(&bootstrap, (heap)heaps.linear_backed, 5, MAX_MCACHE_ORDER,
+    boolean is_lowmem = is_low_memory_machine();
+    int max_mcache_order = is_lowmem ? MAX_LOWMEM_MCACHE_ORDER : MAX_MCACHE_ORDER;
+    bytes pagesize = is_lowmem ? U64_FROM_BIT(max_mcache_order + 1) : PAGESIZE_2M;
+    heaps.general = allocate_mcache(&bootstrap, (heap)heaps.linear_backed, 5, max_mcache_order,
```
francescolavra, line 81: The general heap is used by the tls klib for its malloc/free implementation; if I try to use this klib in a a low-memory VM, when mbedtls_calloc() in klib/mbedtls.c is called to allocate memory chunks larger than 2KB, memory is allocated directly from the linear backed heap, which causes a subsequent deallocation attempt to fail with something like "assertion f->avail <= o->objs_per_page failed at /usr/src/nanos/src/runtime/heap/objcache.c:176  in objcache_deallocate(); halt".
Since the tls klib can't be used if it can't allocate memory chunks larger than 2 KB, we could prevent it from being loaded in the first place when running in low-memory VMs, e.g. by adding a check in its init() function. But this would prevent us from using the cloud_init klib even if we don't need HTTPS. So we should probably add a check in tls_connect() instead, so that the klib loads correctly but returns an error when one tries to set up a TLS session.

francescolavra (commented):
As a heads up, with this change the tls klib cannot be used anymore in low-memory VMs, which means that such VMs won't be able to use the radar service, and won't be able to use HTTPS URLs to retrieve files or environment variables with the cloud_init klib.
Also, in low-memory VMs writes to network socket file descriptors will be limited to max 2KB (so the netsock and tun runtime tests will fail when run on a 64MB Qemu instance).

src/kernel/init.c
wjhun: I have addressed this issue by tracking such "fallback" allocations in a hash table stored within the mcache struct. When an allocation is made that exceeds the size of the largest contained mcache, the size is stored in said table with the address as the key. When deallocating objects of unknown (-1ull) size, a lookup/removal is made in this table and, if an entry is found, the size parameter is padded and used for the deallocate. Small objects are not recorded, so a lookup failure proceeds to bin the allocation according to objcache page membership as was done prior. Deallocation of objects with a specified size classified as fallback also requires removal from the table. The cost here is a table insertion and removal for all objects of fallback size as well as a lookup/removal for all deallocations of an unspecified size. Given that such uses are not typical during kernel runtime and are primarily for vendor code use, this should add no cost in most cases, and even a negligible one for said exceptions.

The upshot here is that the mcache may now be used with a "malloc/free"-type interface regardless of allocation size. We don't have an exhaustive mcache test (#1886), but I have observed such fallback deallocations working correctly with a test run using cloud-init.

src/net/net.c
```diff
@@ -87,7 +88,7 @@ void *lwip_allocate(u64 size)
 {
     /* To maintain the malloc/free interface with mcache, allocations must stay
        within the range of objcaches and not fall back to parent allocs. */
-    assert(size <= U64_FROM_BIT(MAX_LWIP_ALLOC_ORDER));
+    assert(size <= U64_FROM_BIT(lwip_alloc_order));
```
francescolavra, line 91: I think we can remove this assert() now, since it would make writes to network socket file descriptors unable to handle data sizes larger than 2KB in low-memory VMs (e.g. the netsock and tun runtime tests would fail in such VMs); so we don't even need the lwip_alloc_order global variable.
Likewise, the check against the max alloc order in mbedtls_calloc() in klibs/mbedtls.c could be removed.

src/runtime/heap/mcache.c
```diff
@@ -41,19 +42,29 @@ u64 mcache_alloc(heap h, bytes b)
     rputs(": ");
 #endif
     if (b > m->parent_threshold) {
+        if (!m->fallbacks) {
+            m->fallbacks = allocate_table(m->meta, identity_key, pointer_equal);
```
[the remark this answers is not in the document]
wjhun: Thanks for pointing this out; I completely overlooked it. I tried a few different approaches to formulating a key and ran some mock allocation tests in the kernel. Even shifting down by PAGELOG wouldn't be an improvement for our primary use, because it would still produce the same key whenever the allocation size is over 16KB (4*PAGESIZE), and the minimum fallback size for (non-lowmem) Nanos general mcache is 128KB. As allocation sizes increase, the address lsb goes even higher, so it seemed the right thing would be to mix every pair (default four buckets) of bits starting at bit 12 to some point. I chose 23 as the high bit as this helps capture some entropy from 2MB size allocations as well. In a test of allocating memory from 128KB through 4MB, this produced an even distribution among the four buckets.

Incidentally, it's kind of too bad the table interface doesn't use closures for the key and compare, because then then the key function could be tuned to the specific mcache. But I think the above, while adding 15 instructions on x86 (many of which I expect would issue in parallel), is sufficiently general.

src/net/net.c
wjhun: Removed, thanks!

francescolavra: The lwip_alloc_order global variable is no longer needed and can be removed.
Also, the check against the max alloc order in mbedtls_calloc() in klibs/mbedtls.c can be removed.

wjhun: Ok, update pushed.

PR #1887, opened 2023-07-06
Unix: add vsock support

### LINKED ISSUE #1876: virtio-vsock implementation for use with Firecracker
Firecracker supports vsocks, I don't know if the implementation is different from qemu's, but I thought I'd specify nonetheless.

I'm researching using `nanos` for very specific use cases and I noticed vsocks where not implemented yet. I assume that's not a widely used feature in VMs in general, but we use it for exporting logs from a VM. Assigning the serial console for stdout and stderr was far too slow for our own use case and instead we're passing them via a vsock.

Is a vsock implementation on a roadmap? I couldn't find much information about it.

francescolavra:
This change set adds support for the vsock socket family, which allows a program running in a virtual machine to communicate with programs running in the host or in other virtual machines.
The vsock transport mechanism currently supported is the mechanism implemented by virtio socket devices, used in both QEMU and AWS Firecracker. The current vsock implementation supports stream-type sockets.

In addition, an issue in the virtio PCI driver (discovered when running Nanos under QEMU with a "vhost-user-vsock-pci" device) is being fixed.

Closes #1876.


src/unix/vsock.c
```diff
@@ -0,0 +1,990 @@
...
+        if (notify)
+            notify_dispatch(s->sock.f.ns, vsock_events_connlocked(s));
+    }
+    return conn;
+}
+
+void vsock_buf_space_notify(vsock_connection conn, u64 buf_space)
+{
+    vsock s = conn->vsock;
+    blockq_wake_one(s->sock.txbq);
```
sanderssj, line 932: I got a few crashes here with a page fault that looked like txbq was null. As I turned on some of the vsock debugging, the problem went away and I haven't been able to reproduce it even after turning it back off. It's possible there was just a stale object file, but the crash happened after the "Hello from Nanos" message was delivered (following this article https://nanovms.com/dev/tutorials/what-is-vsock-why-use-with-unikernels), so I initially thought there was perhaps a race when cleaning up the socket. Still, I'm not sure it's a real problem and I unfortunately didn't capture the backtrace.

sanderssj (commented):
I have some questions about the MSI-X changes. I don't believe we should be mixing interrupt types on a single PCI device -- I believe it should be configured to be only MSI-X, MSI, or legacy for all interrupt sources on the device, and the device does not use both at the same time.

I think a better way to handle this might be to change how we deal with the configuration change interrupt -- the virtio spec allows for queue notifications and config notifications to be mapped to the same MSI-X vector (see 4.1.5.* in virtio spec 1.2), so I think it's unnecessary to reserve an entire vector just for config changes, the driver just needs to know which vector will require checking both a queue and the config space for changes.

Alternatively, the driver is allowed to write NO_VECTOR (-1) into a queue or config vector to disable notication for it. That might be an easy and appropriate action to take if the requested vector is more than the available MSI-X vectors.

I'm not sure I understand why it's necessary to reserve 0 for the config interrupt for all types (in vtpci_alloc_virtqueue), when perhaps it should be the driver's responsibility to map to a vector of its choosing? Perhaps @wjhun can offer additional insight?

src/unix/vsock.c
francescolavra: Although I have been unable to reproduce any crash here, I think I know what may have happened: the VIRTIO_VSOCK_OP_RST packet sent by the host in response to the VIRTIO_VSOCK_OP_SHUTDOWN packet sent by the guest (when the socket is closed by the guest program) was processed by the virtio driver just before terminating the VM, and so `conn->vsock'  was NULL, leading to a null pointer dereference.
I added a check for a non-null pointer, just like it's done in other similar functions called by the transport layer when processing a received packet.

francescolavra:
@sanderssj you are right about the MSI-X changes, we shouldn't be mixing interrupt types on a single PCI device. I replaced the fixed assignment of MSI-X slots with a dynamic allocation, so the config change interrupt is no longer statically assigned to slot 0, and the Nth virtqueue interrupt is no longer statically assigned to slot N+1; instead, slots are assigned dynamically based on whether the config change interrupt is used by a driver and based on which virtqueues are set up.
The assignment of slot 0 to the config interrupt was done in commit 91f86efea41494f00b8d2bad44febede4deb629f, I believe it was done in this way just to keep the implementation simple, but as you pointed out we can assign any available MSI-X slot to any virtqueue or to the config interrupt.
The kernel still doesn't support assigning the same slot to more than one interrupt handler, but it seems unlikely that we will ever need that.

sanderssj (approved):
LGTM!

Issue #1888, opened 2023-07-06
vdso calls appear to be unused when running -noaccel


wjhun:
This issue was discovered while investigating #1884 as well as attempting to run runtime-tests with acceleration. A call to vdso clock_gettime() would reveal a flaw in the time test, while tests running without acceleration (including CI tests) would mask the issue. This suggests that there may be an issue with our vdso implementation that prevents glibc from selecting vdso calls when running under TCG emulation, causing clock_gettime(2) and other glibc calls to fall back to invoking syscall implementations.

francescolavra:
This is a known issue (#1045) and applies to the time(), clock_gettime() and gettimeofday() vdso calls, which fall back to using their respective syscalls if a pvclock is not available (the vdso_get_now_fn() function returns vdso_now_none unless the registered clock source is VDSO_CLOCK_PVCLOCK).

wjhun:
Ok, makes sense. Closing.

PR #1889, opened 2023-07-06
time test: perform EFAULT tests through syscalls only

wjhun:
The time test may fail with a SIGSEGV if clock_gettime() is performed via VDSO with a faulting user address. The same failure occurs when running the test under Linux. To fix this, tests for EFAULT should directly invoke the syscall being tested, avoiding VDSO.

See #1888


sanderssj (approved):
LGTM

PR #1890, opened 2023-07-07
Additional tests for optional nanos features

sanderssj:
Nanos has a number of behaviors that are only activated when explicitly configured in the manifest. Many of these are informational or related to debugging, so they are not exercised in normal testing. Sometimes major changes in nanos can break these features and without testing coverage the problem goes unnoticed.

This PR creates a python script that performs tests on some of these features, and fixes some bugs that were discovered by doing this testing:

- The syscall_defer option was found to be obsolete and no longer did anything.
- The syscall_summary option was panicking and needed a thread pointer check
- The coredump feature had an assert failure because the failing context's fault handler wasn't cleared while waiting for the coredump to be written out
- The netconsole feature had a deadlock when trying to send a message from an lwip callback

Other changes include printing the IP in an assert failure message to make it easier to identify failed asserts in inlined functions, and an aslr runtime test that prints out addresses in order to verify the noaslr option. The test script has also been included in the build-pc CI checks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ jobs:
             OPS_DIR=$HOME/.ops
             PATH=$HOME/.ops/bin:$PATH
             make test-noaccel
+      - run: test/extended_tests.py
       - run:
           command: |
             if [[ ! -v GCLOUD_SERVICE_KEY ]]; then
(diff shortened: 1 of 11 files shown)


francescolavra (commented):
LGTM

PR #1894, opened 2023-07-13
TFS: fix directory cleanup when writing directory entries to log

francescolavra:
When a TFS metadata tuple is written to the TFS log, if it refers to a directory entry it should be cleaned up (recursively if the entry is a directory itself) by temporarily removing references to parent directories, otherwise the log will contain useless entries. This is not being done in the current code by the tfs_rename() function, which is not doing the cleanup recursively, with the result that renaming a non-empty directory produces spurious TFS log entries. In addition, if a TFS log rebuild is triggered during the first call to log_write_eav() by filesystem_write_eav(), the filesystem_log_rebuild() function fixes up the entire filesystem tree, which may cause the second call to log_write_eav() to attempt a log write of a tuple with a reference to the parent.

This change fixes the above two issues by adding the logic to clean up and fix up a directory tree to the filesystem_write_eav() function, which now takes an additional boolean argument indicating whether a cleanup (and subsequent fixup) should be performed. The rename runtime test is being amended with a new test case that would produce spurious TFS log entries without this fix.

### DIFF
--- a/src/fs/tfs.c
+++ b/src/fs/tfs.c
@@ -372,6 +372,8 @@ closure_function(2, 3, void, filesystem_storage_read,
 }
 
 #ifndef TFS_READ_ONLY
+static tuple cleanup_directory(tuple dir);
+
 fs_status filesystem_write_tuple(tfs fs, tuple t)
 {
     if (fs->fs.ro)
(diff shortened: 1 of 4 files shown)


src/fs/tlog.c
```diff
@@ -625,7 +625,10 @@ boolean log_write_eav(log tl, tuple e, symbol a, value v)
     u64 len = buffer_length(tl->tuple_staging);
     if ((tl->state == TLOG_STATE_FAILED) || len >= TFS_LOG_MAX_TUPLE_STAGING_BYTES)
         return false;
+    tuple parent = (v && is_tuple(v)) ? cleanup_directory(v) : 0;
```
wjhun, line 628: Although not likely to be an issue today - given that we're currently only writing non-filesystem metadata at once during mkfs, and not updating it during runtime (except for writing the booted flag) - it seems that moving the directory cleanup and fixup to this level might interfere with updating a non-tfs tuple that happens to contain a ".." attribute.

francescolavra: OK, I moved the logic to clean up and fix up a directory tree to the filesystem_write_eav() function, which now takes an additional boolean argument indicating whether a cleanup (and subsequent fixup) should be performed

src/fs/tfs.c
```diff
@@ -382,12 +384,22 @@ fs_status filesystem_write_tuple(tfs fs, tuple t)
...
+fs_status filesystem_write_eav(tfs fs, tuple t, symbol a, value v, boolean cleanup)
 {
     if (fs->fs.ro)
         return FS_STATUS_READONLY;
-    if (log_write_eav(fs->tl, t, a, v) &&
-            (!fs->temp_log || log_write_eav(fs->temp_log, t, a, v)))
+    tuple parent = cleanup ? cleanup_directory(v) : 0;
+    boolean success = log_write_eav(fs->tl, t, a, v);
+    if (success && fs->temp_log) {
+        /* The above log_write_eav) call might have started a TFS log rebuild, in which case the
```
wjhun, line 394: nit: parens

wjhun (approved):
LGTM

PR #1896, opened 2023-07-13
prevent infinite recursion when encoding tuples

wjhun:
This change modifies the tuple encoding functions to only encode a tuple's content once per call, thus avoiding infinite recursion in cases where there exists a cycle in a tuple structure (for instance, if a ".." directory entry is mistakenly encoded and the current directory metadata is revisited as a child of the parent directory). It achieves this by passing a table of visited tuples through the call stack, recording each encoded tuple (immediate or reference) in the table. Future attempts to encode any tuple found in the table (within the span of the encode_tuple(), encode_eav() or encode_value() call) will result in a reference to the tuple being encoded with a zero attribute count. The tuple_test unit test has been updated accordingly to reflect this behavior, and a check has been added to test the encoding of a tuple that contains a reference to itself.

Incidentally, this change prevents the VM crash in (albeit not the root cause of) #1894

### DIFF
--- a/src/runtime/tuple.c
+++ b/src/runtime/tuple.c
@@ -292,14 +292,14 @@ void encode_symbol(buffer dest, table dictionary, symbol s)
     }
 }
 
-void encode_tuple(buffer dest, table dictionary, tuple t, u64 *total);
-void encode_value(buffer dest, table dictionary, value v, u64 *total)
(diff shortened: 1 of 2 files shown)


francescolavra (commented):
LGTM

PR #1905, opened 2023-07-28
Fault in process stack and heap pages

wjhun:
Up to now, Nanos would allocate and map an unnecessarily physically-contiguous
2MB chunk of memory for the userspace stack. This could be wasteful for
low-memory configurations both in terms of allocated but potentially unused
stack space as well as memory fragmentation created by the large, contiguous
allocation. Similarly, extensions of the process heap, made via a call to brk(2), would
also result in physically-contiguous allocations and mappings of the extended area.

With this change, only a single page (PROCESS_STACK_PREALLOC_SIZE) is
allocated for stack usage at first, allowing further stack pages to be faulted
in as needed. A new vmap flag, VMAP_FLAG_STACK, has been added to identify the
stack mapping, a fault on which is handled in the same way as an anonymous
mmap. Following in suit, heap extensions in brk(2) result only in an extending of the
heap vmap (now marked with VMAP_FLAG_HEAP), and heap pages are faulted
in on demand.

These changes save over 2MB (2172KB) of memory allocated, compared to master,
when running the webs test in a low memory configuration (16MB). The instance is
still not able to load with only 14MB of system memory, however, possibly due to
memory fragmentation and alignment of other large allocations (e.g. program binary).

### DIFF
--- a/src/unix/exec.c
+++ b/src/unix/exec.c
@@ -14,21 +14,35 @@
 
 #define DEFAULT_PROG_ADDR       0x400000
 
-#define spush(__s, __w) *((--(__s))) = (u64)(__w)
+static void *stack_prealloc(void *start, u64 size)
+{
+    u64 sphys = allocate_u64((heap)heap_physical(get_kernel_heaps()), size);
+    assert(sphys != INVALID_PHYSICAL);
+    start -= size;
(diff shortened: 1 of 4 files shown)


francescolavra (commented):
LGTM

sanderssj (approved):
LGTM

PR #1906, opened 2023-07-28
test/go: run 'go mod tidy' in post-clean rule

wjhun:
Running a 'make clean' on a clean tree, or a tree that hasn't executed the test/go tests, may result in an error such as:

   go: github.com/nanovms/ops@v0.0.0-20210408190622-0dc76ec84cd2: missing go.sum entry; to add it:
   	go mod download github.com/nanovms/ops

To avoid this, the post-clean rule now runs 'go mod tidy' prior to 'go clean'.

[I'm not certain if this is the right fix; does anyone know if there's a simpler way to bypass the above error?]

### DIFF
--- a/test/go/Makefile
+++ b/test/go/Makefile
@@ -20,6 +20,7 @@ test: build
 CLEANFILES+=	$(OBJDIR)/kernel.img $(OBJDIR)/boot.img
 
 post-clean:
+	$(Q) $(GOMOD) tidy
 	$(Q) $(GOCLEAN)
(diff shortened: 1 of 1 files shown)


rinor:
Usually I use `GO111MODULE=off` to skip dealing with that (when not caring about tests)
- `GO111MODULE=off make clean`

francescolavra (commented):
LGTM (I don't know if there is a better way to deal with this error)

sanderssj (approved):
LGTM, I don't know a better solution either.

PR #1908, opened 2023-07-30
Klibs: ntp: add backoff time for missing response error messages

francescolavra:
Instead of printing a "failed to receive server response" message every time the response to an NTP query packet is not received (which is not uncommon, especially if the NTP server is geographically distant from the instance, given that NTP uses UDP as transport protocol), with this change the NTP klib prints a message only if no response is received in a given interval (which is initialized based on the configured NTP query interval, and is incremented after printing a message in order to avoid spamming the console).

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -117,6 +117,8 @@ static struct {
     s64 skew;
     s64 max_corr_freq;
     s64 max_base_freq;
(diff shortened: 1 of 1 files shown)


sanderssj (approved):
LGTM

PR #1909, opened 2023-07-30
Miscellaneous improvements and fixes

sanderssj:
This is a small collection of fixes and a couple of improvements from other work. The
clone_id_heap and reserve_backed_heap_wrapper are not currently used but will be useful
later, and the linker elides them from the final binary for now so they don't affect code size.

### DIFF
--- a/src/http/http.c
+++ b/src/http/http.c
@@ -431,7 +431,7 @@ closure_function(2, 1, void, each_http_request,
         if (!hl->default_handler)
             goto not_found;
         apply(hl->default_handler, method, hr, v);
-        return;
+        goto out;
     }
 
     buffer rel_uri = clone_buffer(hl->h, uri);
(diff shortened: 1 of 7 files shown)


src/runtime/heap/id.c
```diff
@@ -465,6 +465,46 @@ boolean id_heap_range_foreach(id_heap i, range_handler rh)
...
+        s->n.r = r->n.r;
+        assert(rangemap_insert(i->ranges, &s->n));
+    }
+    i->mgmt = 0;                /* regenerate */
+    return i;
+  fail_dealloc_ranges:
+    rangemap_foreach(i->ranges, n) {
+        id_range r = struct_from_field(n, id_range, n);
+        rangemap_remove_node(i->ranges, n);
+        deallocate(h, r, sizeof(*r));
```
francescolavra, line 500: We should also deallocate the bitmap, by calling `deallocate_bitmap(r->b)`.

src/http/http.c
```diff
@@ -461,14 +461,16 @@ closure_function(2, 1, void, each_http_request,
...
         set(v, sym(relative_uri), rel_uri);
-    else
-        deallocate_buffer(rel_uri);
 
-    if (match) {
+    if (match)
         apply(match->each, method, hr, v);
-        return;
-    }
+    deallocate_buffer(rel_uri);
```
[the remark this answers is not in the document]
sanderssj: Ok, I reverted to the original code.

src/runtime/heap/id.c
sanderssj: Added back in, thanks!

PR #1910, opened 2023-07-31
inotify_event_handler: fix event mask check

francescolavra:
The IN_ISDIR flag in the event mask for a given inotify event does not identify a specific event; instead, it is a qualifier that further describes an event identified by another flag in the mask. Therefore, the presence of the IN_ISDIR flag should not be considered to determine whether an event is of interest for an inotify file descriptor.
This change fixes a bug that causes events occurring to directories to be read from an inotify file descriptor regardless of the mask specified in the relevant inotify_add_watch() syscall; more specifically, this bug causes events that should not be returned when reading from the file descriptor to be returned with a mask value set to IN_ISDIR.
The inotify runtime test has been amended with a few test cases that would fail without this fix.


wjhun (approved):
LGTM

PR #1913, opened 2023-08-09
Additional e2e tests

sanderssj:
This PR expands the e2e tests beyond testing just for web servers. The
new e2e code can run an arbitrary program, and the new testing code checks
for the exit status message at the end of the output to determine a test
success or failure. The AsyncCmdStart code has been modified to use a
CommandContext with a custom Cancel function in order to provide a
reliable timeout and process status for the command. An additional
Postrun step is added in order to handle any cleanup required after a
test is run.

The new tests include several versions of large memory allocations with
several different language packages, a test for disk full conditions and
data corruption (thanks to @rinor), and a test of some of the cloud init
functionality.

The memory allocation tests revealed two problems in out-of-memory
handling. One is fixed by touching the sigframe memory to test that
any stack expansion can be allocated, and the other is by adding some
code to test if a user program is hitting frequent repeated out of memory
page faults which will terminate the program instead of letting it spin
forever trying to allocate memory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,7 @@ Makefile.local
 /test/go/image
 /test/runtime/soop.data
 .ruby
-test/e2e/go/main
-test/e2e/rust/main
+test/e2e/*/main
 
 # Default build directory
 /output
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -523,10 +523,21 @@ void vm_exit(u8 code)
 
     tuple root = get_root_tuple();
     if (root) {
-        u64 expected_code;
-        if (get_u64(root, sym(expected_exit_code), &expected_code) &&
-                expected_code == code)
-            code = 0;
+        value expected = get(root, sym(expected_exit_code));
+        if (expected) {
+            u64 expected_code;
+            if (is_string(expected) && u64_from_value(expected, &expected_code) &&
+                expected_code == code) {
+                code = 0;               
+            } else if (is_tuple(expected)) {
+                for (int i = 0; get_u64(expected, intern_u64(i), &expected_code); i++) {
+                    if (expected_code == code) {
+                        code = 0;
+                        break;
+                    }
+                }
+            }
+        }
         if ((code != 0) && get(root, sym(reboot_on_exit)))
             vm_reset();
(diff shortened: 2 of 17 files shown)


src/x86_64/unix_machine.c
```diff
@@ -136,15 +138,19 @@ boolean setup_sigframe(thread t, int signum, struct siginfo *si)
...
        but not 16 bytes; the ABI requires that stacks are aligned to 16 before
        a call, but the sigframe return into the function takes the place of a
        call, which would have pushed a return address. The function prologue
        typically pushes the frame pointer on the stack, thus re-aligning to 16
        before executing the function body. */
-    rsp -= pad(sizeof(struct rt_sigframe), 16) + 8; /* ra offset */
-
+    u64 sz = pad(sizeof(struct rt_sigframe), 16) + 8; /* ra offset */
+    rsp -= sz;
+    if (!fault_in_user_memory_for_process(t->p, (void *)rsp, sz, true))
```
francescolavra, line 152: It seems to me there is no need to to fault in explicitly the user memory, because we already have the `if (context_set_err(ctx)) return false;` instructions which ensure setup_sigframe() returns false if a page cannot be faulted in due to OOM conditions. So a call to fault_in_memory() would effectively do the same thing that's already being done by writing to the user stack page, e.g. in setup_ucontext(). The same goes for the aarch64 and riscv code.
I tried running the e2e tests after reverting this commit, and I couldn't reproduce the problem you describe where the VM gets stuck forever in a loop: e.g. the ruby_alloc test ends when either setup_sigframe() returns false (which as far as I can tell works correctly with the existing code), or a sequence of out-of-memory page faults occurs (in which case the new OOM kiiller takes care of shutting down the program).

sanderssj: When I first started looking at this, before I had the OOM killer written, I was seeing two different symptoms related to OOM (after adding additional debugging in), and this fix eliminated one of the symptoms. I also can't seem to reproduce what I was seeing before. I see the context_set_err should take care of it in exactly the same way with or without this fix, so now I'm questioning what I saw originally. In any case, you are right and this fix isn't needed, so I dropped the commit. Thanks!

PR #1914, opened 2023-08-09
Add support for building with clang on Linux

francescolavra:
With these changes, it is possible to build Nanos with the clang compiler (by typing e.g. `make CC=clang`), as an alternative to gcc.

Tested with clang version 15.0.7.


wjhun (approved):
LGTM

PR #1915, opened 2023-08-09
fault in user program pages

wjhun:
Until this point, user program executables have been loaded and mapped, in their entirety, using physically-contiguous backed allocations. This practice has the effect of creating pressure on the memory system in the form of fragmentation, which is especially problematic in low memory configurations.

With this change, the user program file and program loader, if any, are now demand paged. Rather than reading the program with filesystem_read_entire(), a small portion is read, which is enough to contain the ELF program headers. The program headers are parsed using load_elf(), but instead of mapping loaded data directly into user memory, the elf_map_handlers instead add pagecache node information to the vmaps covering program text and data, allowing pages for these areas to be faulted in. BSS mappings are flagged as such and treated as anonymous mappings (like stack and heap mappings).

To support this change, load_elf() now passes only file offsets (instead of buffer physical addresses) to elf_map_handlers as the second argument. The handlers must find the buffer physical address by enclosing the buffer start and finding the physical address based on the given offset. The logic for handling non-page-aligned starts to bss sections has been moved from load_elf() to the handlers themselves, and the bss portion of any loadable segment is expressed with a new 'bss_size' argument, rather than handled in two separate invocations of the elf_map_handler.

Note that this feature is disabled (in favor of a statically-mapped load in entirety) when ltrace is in use, because the ltrace code needs to be able to make instruction-level modifications to the program in memory. It may also be disabled with the 'static_map_program' config option.

exec_elf() now takes a path string for the program file and performs the file loading itself. If 'ltrace' or 'static_map_program' are specified, the whole program binary is loaded at once and kept resident in memory. Otherwise, only a minimum necessary to read the program headers (ELF_PROGRAM_LOAD_MIN_SIZE, or one page) is loaded at first. Additional ELF data may be subsequently read as necessary, for instance to access the path string for the program loader, or to read in symbol data if 'ingest_program_symbols' is specified. When running in the fault-in mode, buffered file data is released after ELF header parsing has completed.

### DIFF
--- a/src/kernel/klib.c
+++ b/src/kernel/klib.c
@@ -34,31 +34,64 @@ closure_function(1, 1, boolean, klib_elf_walk,
     return true;
 }
 
-closure_function(1, 4, u64, klib_elf_map,
-                 klib, kl,
-                 u64, vaddr, u64, paddr, u64, size, pageflags, flags)
+static klib_mapping add_klib_mapping(klib kl, u64 vaddr, u64 paddr, u64 size, pageflags flags)
 {
-    klib kl = bound(kl);
-    boolean is_bss = paddr == INVALID_PHYSICAL;
-    if (is_bss) {
-        paddr = allocate_u64((heap)heap_physical(klib_kh), size);
-        assert(paddr != INVALID_PHYSICAL);
-    }
-
     klib_mapping km = allocate(heap_locked(klib_kh), sizeof(struct klib_mapping));
-    assert(km != INVALID_ADDRESS);
+    if (km == INVALID_ADDRESS)
+        return km;
     km->n.r = irangel(vaddr, size);
     km->phys = paddr;
     km->flags = flags;
-
-    klib_debug("%s: kl %s, vaddr 0x%lx, paddr 0x%lx%s, size 0x%lx, flags 0x%lx\n",
-               __func__, kl->name, vaddr, paddr, is_bss ? " (bss)" : "", size, flags.w);
-
+    klib_debug("%s: vaddr 0x%lx, paddr 0x%lx, size 0x%lx, flags 0x%lx\n",
+               __func__, vaddr, paddr, size, flags.w);
     assert(rangemap_insert(kl->mappings, &km->n));
     map(vaddr, paddr, size, flags);
-    if (is_bss)
-        zero(pointer_from_u64(vaddr), size);
-    return vaddr;
+    return km;
+}
+
+closure_function(2, 5, boolean, klib_elf_map,
+                 klib, kl, buffer, b,
+                 u64, vaddr, u64, offset, u64, data_size, u64, bss_size, pageflags, flags)
+{
+    klib kl = bound(kl);
+    klib_debug("%s: kl %s, vaddr 0x%lx, offset 0x%lx, data_size 0x%lx, bss_size 0x%lx, flags 0x%lx\n",
+               __func__, kl->name, vaddr, offset, data_size, bss_size, flags);
+    u64 map_start = vaddr & ~PAGEMASK;
+    data_size += vaddr & PAGEMASK;
+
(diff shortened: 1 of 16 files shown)


src/kernel/klib.c
```diff
@@ -101,8 +134,11 @@ closure_function(3, 1, status, load_klib_complete,
     klib_debug("   loading klib @ %R, resolving relocations\n", kl->load_range);
     elf_apply_relocs(b, where);
 
+    klib_debug("   ingesting elf symbols\n");
+    add_elf_syms(kl->elf, kl->load_range.start);
```
francescolavra, line 138: Why is this being moved here? This code can run multiple times for a single klib (e.g. the first time it may fail with KLIB_MISSING_DEP, and then after another klib is loaded, the second time it succeeds); having this code run only when KLIB_INIT_OK or KLIB_INIT_IN_PROGRESS ensures elf symbols are ingested only once per klib.

platform/virt/Makefile
```diff
@@ -208,6 +208,7 @@ BOOT_OBJDIR=    $(OBJDIR)/boot
 UEFI_SRCS= \
 	$(SRCDIR)/aarch64/elf64.c \
 	$(SRCDIR)/aarch64/uefi.c \
+	$(SRCDIR)/boot/elf.c \
```
francescolavra, line 211: This file shouldn't actually be needed here, but in SRCS-uefi in platform/pc/boot/Makefile instead: the uefi_kernel_complete() function is only used if uefi_arch_options->load_to_physical is false, which is the case for the x86 UEFI loader, but not for the aarch64 one.

src/kernel/klib.c
wjhun: Prior to doing the tail copy, I initially had klib_elf_map map the last page of the data section and simply zero the non-page-aligned portion of the bss section. This would cause the elf symbol ingest to later fail, as I think this zeroing was affecting either the symtab or stringtab sections. I moved the elf ingest prior to the elf loading for this reason, but missed that it could be invoked multiple times.

Since I subsequently changed all the non-faulting elf map handlers do the tail copy, this is no longer necessary. I have now reverted this change.

src/unix/exec.c
```diff
@@ -326,24 +409,131 @@ process exec_elf(buffer ex, process kp)
...
+
+void exec_elf(process kp, string program_path, status_handler complete)
+{
+    exec_debug("%s: path \"%b\", complete %p (%F)\n", __func__, program_path, complete, complete);
+    kernel_heaps kh = (kernel_heaps)(kp->uh);
+    heap general = heap_locked(kh);
+    tuple root = kp->process_root;
+    fsfile f = fsfile_open_or_create(program_path, false);
+    if (!f) {
+        apply(complete, timm("result", "unable to open program file %v", program_path));
```
francescolavra, line 520: Using fsfile_open_or_create() means that if the file doesn't exist it will be created (with length 0), so instead of an "unable to open program file" error message there is a more cryptic "failed to parse elf file" message. The same goes for the interpreter, if the file doesn't exist we get a "failed to parse elf file" message instead of "couldn't find program interpreter".
Perhaps we could change fsfile_open_or_create() to a more generic fsfile_open() that takes an additional boolean argument indicating whether the file should be created if not found.

wjhun: Ok, I added an fsfile_open() function for opening existing files, but left fsfile_open_or_create() in place for existing uses. I also noticed that, prior to this PR, a missing program file would result in a similarly-cryptic message "cannot set on unknown value" (or an "assert(0)" failure and backtrace prior to new-tag-types merge) due to a set on a null tuple in startup(). I changed this to print an "unable to resolve program path" error if resolve_path() returns null.

francescolavra (commented):
LGTM

PR #1916, opened 2023-08-11
Unix file paths: properly enforce NAME_MAX and PATH_MAX limits

francescolavra:
This change set fixes handling of file names whose length exceeds NAME_MAX, so that -ENAMETOOLONG is returned when appropriate, and allows using symlinks whose target path length is within the PATH_MAX limit.


sanderssj (approved):
LGTM

PR #1917, opened 2023-08-11
Unix syscalls: implement utimensat(2)

francescolavra:
This change adds to the kernel an implementation of the utimensat syscall, which allows programs to use the utimensat() and futimens() POSIX functions.
The new code is being exercised in the runtime tests.

### DIFF
--- a/src/unix/filesystem.c
+++ b/src/unix/filesystem.c
@@ -212,6 +212,82 @@ sysreturn utimes(const char *filename, const struct timeval times[2])
     return utime_internal(filename, atime, mtime);
 }
 
+static boolean utimens_is_valid(const struct timespec *t)
+{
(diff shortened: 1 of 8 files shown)


src/unix/filesystem.c
```diff
@@ -212,6 +212,82 @@ sysreturn utimes(const char *filename, const struct timeval times[2])
     return utime_internal(filename, atime, mtime);
 }
 
+static boolean utimens_is_valid(const struct timespec *t)
+{
+    return (t->tv_nsec < BILLION) || (t->tv_nsec == UTIME_NOW)|| (t->tv_nsec == UTIME_OMIT);
```
sanderssj, line 217: nit - need space in front of second ||

sanderssj (approved):
LGTM

PR #1920, opened 2023-08-16
utime/utimes: fix handling of non-existent files

francescolavra:
If the program calls utime() or utimes() with a non-existent file path argument, the kernel crashes with "assertion ctx == m->turn failed at src/kernel/mutex.c:263 (IP 0xffffffff800491af)  in mutex_unlock(); halt".
This commit fixes the above issue and adds a test case that would fail without this fix.


sanderssj (approved):
LGTM

Issue #1911, opened 2023-08-04
issue(sqlite3): WAL - journal_mode and multi-thread support - err(522) SQLITE_IOERR_SHORT_READ


rinor:
Is there any known issue/gotcha to be aware of while testing? (i.e: not supported yet fs/kernel stuff)

eyberg:
i'm not aware of anything atm, if this needs more discussion perhaps we can mv it to the discussions area? - if you do come across anything feel free to open an issue

rinor:
I'm about to start testing in a few, and if issues found, the communication thread may be similar to https://github.com/nanovms/nanos/issues/1845.

If you prefer to have separate gh issue for each problem found, feel free to close this.

rinor:
<details><summary><code>Code</code> used to test</summary>

  - binary named `nanos-sqlite`

```sh
$ go mod init nanos-sqlite
$ go mod tidy
$ CGO_ENABLED=0 GO111MODULE=on go build --ldflags "-s -w" -trimpath -a -v -tags osusergo,netgo
```



```go
package main

import (
	"database/sql"
	"log"
	"math/rand"
	"os"
	"strconv"
	"sync"
	"time"

	_ "modernc.org/sqlite"
	//
	// _ "github.com/mattn/go-sqlite3" // to be tested next (requires cgo)
)

const (
	NANOS_SQLITE3_WRITE_CONNECTION_ENV     = "NANOS_SQLITE3_WRITE_CONNECTION"
	NANOS_SQLITE3_WRITE_CONNECTION_DEFAULT = "file:" + "nanos.sqlite3" + "?mode=rwc&_pragma=journal_mode=WAL&_pragma=synchronous=NORMAL&_pragma=busy_timeout=5000"

	NANOS_SQLITE3_READ_CONNECTION_ENV = "NANOS_SQLITE3_READ_CONNECTION"

	NANOS_SQLITE3_MAX_GOROUTINES_WRITE_ENV     = "NANOS_SQLITE3_MAX_GOROUTINES_WRITE"
	NANOS_SQLITE3_MAX_GOROUTINES_WRITE_DEFAULT = 1

	NANOS_SQLITE3_MAX_GOROUTINES_READ_ENV     = "NANOS_SQLITE3_MAX_GOROUTINES_READ"
	NANOS_SQLITE3_MAX_GOROUTINES_READ_DEFAULT = 1

[...]

PR #1921, opened 2023-08-16
Unix file struct: remove length field

francescolavra:
Using the length field of the file struct to determine the length of an open file can lead to incorrect results if the same file is opened via multiple file descriptors at the same time, because a change of file length resulting from an operation on a file descriptor will not be reflected in the length field of the other descriptors. This can cause problems such as for example the inability to read file contents that have been written via a different file descriptor (https://github.com/nanovms/nanos/issues/1911).
This change set removes the length field from the file struct, and changes the code so that fsfile_get_length() is used to determine the length of an open file. In the process of modifying the code that uses this struct filed, a few other minor issues (that were causing Nanos to deviate from the behavior of the Linux kernel) have been resolved.


sanderssj (approved):
LGTM

PR #1922, opened 2023-08-17
/proc/meminfo special file: add more records

francescolavra:
The sys-info Rust crate has a mem_info() function that reads the contents of the /proc/meminfo file and expects to find a few records (https://docs.rs/sys-info/latest/src/sys_info/lib.rs.html). This change adds the "Buffers", "SwapTotal" and "SwapFree" records so that the mem_info() function doesn't return an error.


sanderssj (approved):
LGTM

PR #1923, opened 2023-08-17
socket_write_tcp_bh(): fix check of TCP send buffer length

francescolavra:
PR #1761 increased the TCP send buffer size and TCP window size from 0xffff to 0x34000 bytes, which improved performance of large network transfers. However, the tcp_sndbuf() macro caps the returned value to 0xffff, and the maximum data length that can be passed to the tcp_write() function is still limited to 0xffff bytes; this means that the kernel can call tcp_write() multiple times with a data length set to the value returned by tcp_sndbuf(), while still remaining within the limit of the send buffer size. This change modifies the socket_write_tcp_bh() function so that the send buffer size is retrieved after each call to tcp_write(), in order to be able to use the entire send buffer when large write requests are made by the user program.
Beside improving the performance of socket write operations, this fixes a potential stall faced by user programs that use epoll with edge detection to monitor the status of network sockets, and assume that a short write (i.e. a write request that processes only part of the data passed to it) on a TCP socket implies that the EPOLLOUT flag has been reset on the socket file descriptor, thus requiring polling for an edge on the EPOLLOUT condition in order to be able to send the remaining data without blocking. For example, the Tokio runtime (a Rust library for asynchronous programming) makes this assumption (in the poll_write function at https://github.com/tokio-rs/tokio/blob/master/tokio/src/io/poll_evented.rs), which causes HTTP responses sent by web servers using this runtime (e.g. OpenObserve) to stall if the data being sent is larger than 64 KB.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -708,8 +708,8 @@ closure_function(6, 1, sysreturn, socket_write_tcp_bh,
         } else {
             n = remain;
         }
-        if (avail < rv + n) {
-            n = avail - rv;
+        if (avail < n) {
+            n = avail;
             apiflags |= TCP_WRITE_FLAG_MORE;
         }
 
@@ -717,8 +717,10 @@ closure_function(6, 1, sysreturn, socket_write_tcp_bh,
         if (err == ERR_OK) {
             if (sg)
                 sg_consume(sg, n);
+            else
+                buf += n;
             rv += n;
-            if (rv == avail)
+            if ((avail = tcp_sndbuf(tcp_lw)) == 0)
                 break;
             remain -= n;
             continue;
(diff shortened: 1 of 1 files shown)


rinor:
Maybe a bit out of topic, but since you are working with OpenObserve (rust) and the [crate](https://docs.rs/sys-info/latest/sys_info/) used to fetch sys info uses also the following file(s), are you planing to add them to nanos also?

I'm just mapping the `proc` dir locally, so not a big deal, just curious.

```sh
proc
└── sys
    └── kernel
        ├── osrelease   5.0-0.1.46
        └── ostype      Nanos
```

sanderssj (approved):
LGTM

PR #1924, opened 2023-08-18
filessytem_mkdir: properly handle existing directories and read-only filesystems

francescolavra:
With this change set, mkdir() returns -EEXIST instead of 0 if the user program tries to create a directory that already exists.
In addition, -EROFS is now properly returned when creating a directory or binding a Unix domain socket on a read-only filesystem.


sanderssj (approved):
LGTM

PR #1925, opened 2023-08-23
Klbs: add special_files

francescolavra:
This new klib will implement miscellaneous special files (e.g. files whose content is dynamically generated, or files that trigger specific actions when written to).
Each special file is enabled by a specific attribute in the "special_files" configuration tuple.
Currently supported attributes:
- "disks" (tuple): special file at /sys/devices/disks, whose contents are a set of records, one for each disk attached to the instance; each record contains the volume name and UUID of the disk, in text format

### DIFF
--- a/klib/Makefile
+++ b/klib/Makefile
@@ -9,6 +9,7 @@ PROGRAMS= \
 	ntp \
 	radar \
 	sandbox \
+	special_files \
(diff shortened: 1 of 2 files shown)


wjhun (approved):
LGTM

PR #1928, opened 2023-08-25
vm_exit(): fix parsing of multiple expected exit codes

francescolavra:
The current version of Ops encodes vector values using the new vector type introduced in TFS version 5; therefore, if multiple expected exit values are specified in the manifest of a given image, the corresponding value decoded by the kernel from the root tuple is a vector.
This change modifies the vm_exit() function so that it can correctly retrieve the expected exit codes when there are multiple such codes; in addition, the checks for string and integer values are being removed because they are redundant: if a value is not a string nor an integer, u64_from_value() returns false.


sanderssj (approved):
LGTM

PR #1929, opened 2023-08-25
PCI add support for multiple root bridges

francescolavra:
A PCI root bridge is a bridge that does not sit downstream another bridge, and as such cannot be detected by probing the host bridge. PCI root bridges are listed in the ACPI tables, where a bridge device handle contains information not only on the memory ranges managed by a given bridge, but also on the "bridge window", i.e. the range of PCI bus numbers managed (directly or indirectly) by the bridge.
This commit amends the ACPI code so that it retrieves both the memory ranges and the bridge windows for all PCI root bridges, and amends the PCI code so that it stores (instead of a vector of PCI buses) a map of bus number ranges (with I/O memory associated to the corresponding root bridge).
The pci_discover() function probes, beside the host bridge, any additional bridges that have been detected via ACPI; the pci_bus_get_iomem() function has been amended so that it can retrieve the I/O memory to be assigned to a given PCI device even if the device is on a different bus than the primary bus of its root bridge.
These changes allow the kernel to detect the  network adapter on AWS instances with multiple root bridges, such as m7a instance types.

### DIFF
--- a/src/drivers/acpi.c
+++ b/src/drivers/acpi.c
@@ -12,6 +12,11 @@
 
 #define acpi_heap   heap_locked(get_kernel_heaps())
 
+typedef struct acpi_pci_res_ctx {
+    range bridge_window;
+    id_heap iomem;
+} *acpi_pci_res_ctx;
+
 boolean acpi_walk_madt(madt_handler mh)
 {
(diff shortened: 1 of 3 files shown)


wjhun (approved):
LGTM

PR #1931, opened 2023-08-31
File-backed mappings: request pagecache fill asynchronously

francescolavra:
Requesting a pagecache fill to resolve a page fault for a file-backed mapping can cause the current context to be suspended, e.g. to acquire the filesystem mutex. The context_switch() function, which is called to switch to a kernel context before calling pagecache_map_page(), does not move the stack pointer to the stack of the kernel context; instead, the context will run on the exception/interrupt stack of the current CPU; if the context is suspended, the current CPU can pick up and run a user thread, which can cause a new exception, whose handler will run on the same stack; as a result, when the suspended context is later resumed, it will run on a corrupted stack, leading to a kernel crash.
This change fixes the above issue by removing the use of the context_switch() function (which is unnecessary because a switch to a kernel context happens when kern_yield() is called) when resolving a page fault, and using an asynchronous thunk instead of calling pagecache_map_page() directly.

### DIFF
--- a/src/unix/mmap.c
+++ b/src/unix/mmap.c
@@ -174,9 +174,13 @@ static status demand_anonymous_page(pending_fault pf, vmap vm, u64 vaddr)
     return STATUS_OK;
 }
 
-static void demand_file_page(pending_fault pf, vmap vm, u64 node_offset, u64 page_addr,
-                             pageflags flags)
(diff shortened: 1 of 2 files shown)


wjhun (approved):
LGTM

PR #1932, opened 2023-08-31
Support building with ARM-based Macs

sanderssj:
This adds support for building nanos on an M1 or M2 based Mac. Several rules
have been rewritten to be more friendly for cross compiling or for determining
CFLAGS based on compiler type and not just platform.

Aarch64 apple clang has several warnings and errors when processing nanos
code and some of them did not have a reasonable solution for making the compiler
happy. Most of these problems were related to the inline assembly. Since homebrew
provides an up-to-date aarch64-elf-gcc, this change uses that instead of clang to
compile the kernel.

While aarch64-elf-binutils works fine for working with the kernel, the linker
appears to have a problem compiling the static runtime tests where the
automatic symbol __ehdr_start is not found. For this reason, a custom
aarch64-linux-binutils formula is provided which allows proper linking of
those static user programs.

The OBJS_CRT definitions that were removed in a previous change have
been restored as they are required for linking user programs on either
architecture of Macs, but they have been relocated to the runtime
Makefile as that is the only place they are used.

A handful of small changes were necessary to silence new compiler warnings.

### DIFF
--- a/src/aarch64/machine.h
+++ b/src/aarch64/machine.h
@@ -45,8 +45,10 @@ typedef u64 bytes;
 #define mask_and_set_field(u, f, v) (clear_field(u, f) | u64_from_field(f, v))
 
 #define DIV(__x, __by, __q, __r) \
-    do { asm("udiv %0, %2, %3; msub %1, %0, %3, %2" :           \
-             "=&r"(__q), "=r"(__r) : "r"(__x), "r"(__by)); } while(0)
+    do { asm("udiv %0, %1, %2" :           \
+            "=r"(__q): "r"(__x), "r"((u64)__by)); \
+        asm("msub %0, %2, %3, %1" :           \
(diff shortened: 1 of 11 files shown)


src/aarch64/machine.h
```diff
@@ -45,8 +45,10 @@ typedef u64 bytes;
 #define mask_and_set_field(u, f, v) (clear_field(u, f) | u64_from_field(f, v))
 
 #define DIV(__x, __by, __q, __r) \
-    do { asm("udiv %0, %2, %3; msub %1, %0, %3, %2" :           \
-             "=&r"(__q), "=r"(__r) : "r"(__x), "r"(__by)); } while(0)
+    do { asm("udiv %0, %2, %3" :           \
+            "=r"(__q), "=r"(__r) : "r"(__x), "r"((u64)__by)); \
```
francescolavra, line 49: The __r variable is not used in this instruction, so it can be removed from the list of operands.

PR #1935, opened 2023-09-06
Volumes: print console message when a filesystem cannot be probed

### LINKED ISSUE #1926: chore(storage): warn when volume name/label length > nanos(VOLUME_LABEL_MAX_LEN)
https://github.com/nanovms/nanos/blob/41c61e23db0d358383fdc8c95b1c046eceb4d147/src/runtime/storage.h#L23

when the non boot volume `name/label length > 31` the kernel won't load/mount it, so a warning log upon disk/volume detection would be very helpful to understand why the volume is not being mounted.

https://github.com/nanovms/ops/issues/1517

francescolavra:
Instead of silently ignoring volumes whose filesystem cannot be probed, a message is now printed on the console, which makes it easier to identify the reason why an attached volume is not mounted by the kernel (e.g. tfs magic mismatch, tfs version mismatch, or invalid label).
Closes #1926.

### DIFF
--- a/src/fs/tfs.h
+++ b/src/fs/tfs.h
@@ -9,7 +9,7 @@ extern io_status_handler ignore_io_status;
 #define MAX_EXTENT_SIZE (PAGECACHE_MAX_SG_ENTRIES * PAGESIZE)
 #define MIN_EXTENT_ALLOC_SIZE   (1 * MB)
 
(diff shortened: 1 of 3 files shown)


sanderssj (approved):
LGTM

PR #1936, opened 2023-09-07
virt/riscv-virt: use virtio-scsi-pci as QEMU virtio SCSI device

### LINKED ISSUE #1482: virtio-scsi fails to probe on aarch64 / virt platform

francescolavra:
The virtio-scsi-device QEMU device sits on the virtio bus, and since the virtio_scsi driver in the kernel does not register itself as a virtio-mmio driver, the root volume is not detected when an image is run with `STORAGE=virtio-scsi`.
This change modifies the virt and riscv-virt platform Makefiles so that a virtio SCSI device is instantiated with the virtio-scsi-pci command line option (which creates a device sitting on the PCI bus) instead of virtio-scsi-device: this allows the root volume to be properly detected when `STORAGE=virtio-scsi` is specified in the `make` command line.
Closes #1482.


wjhun (approved):
LGTM

PR #1938, opened 2023-09-07
reduce stage2 size

wjhun:
Changes introduced in #1903 caused the size of the stage2 bootloader to increase, exceeding 64KB in some build environments. As described in commit ed5a2f77, stage2 binaries larger than this size may cause boot failures when running under Xen (e.g. AWS EC2).

This PR includes a few commits that ameliorate this inflation. A build-time check is also added which causes the build to fail if the stage2 binary size exceeds 64KB. (This is disabled when building debug images with MEMDEBUG or UBSAN enabled.)

### DIFF
--- a/src/runtime/runtime.h
+++ b/src/runtime/runtime.h
@@ -73,7 +73,7 @@ void print_frame_trace_from_here();
         if (!(x)) {                                 \
             __label__ __here;                       \
             print_frame_trace_from_here();          \
-            __here: halt("assertion " #x " failed at " __FILE__ ":%d (IP %p)  in %s(); halt\n", __LINE__, &&__here, __func__); \
+            __here: halt("assertion " #x " failed at %s:%d (IP %p)  in %s(); halt\n", __FILE__, __LINE__, &&__here, __func__); \
         }                                           \
     } while(0)
 #endif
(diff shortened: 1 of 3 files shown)


src/runtime/runtime.h
```diff
@@ -73,7 +73,7 @@ void print_frame_trace_from_here();
         if (!(x)) {                                 \
             __label__ __here;                       \
             print_frame_trace_from_here();          \
-            __here: halt("assertion " #x " failed at " __FILE__ ":%d (IP %p)  in %s(); halt\n", __LINE__, &&__here, __func__); \
+            __here: halt("assertion " #x " failed at %s :%d (IP %p)  in %s(); halt\n", __FILE__, __LINE__, &&__here, __func__); \
```
sanderssj, line 76: It would be useful to remove the space between "%s" and ":%d" so the file:line is contiguous for tools that use it for navigation.

sanderssj (approved):
LGTM, though looking through the assert history, I bet the format specifier was avoided so the assert had at least some useful information when triggered before the runtime is initialized. Is it worthwhile to include a version of assert with the old concatenation hidden behind a define for debugging?

francescolavra (commented):
LGTM

PR #1939, opened 2023-09-08
Socket ioctl: add support for SIOCGIFHWADDR

francescolavra:
This change adds support for SIOCGIFHWADDR ioctl requests to retrieve the hardware address from a network interface. The netif_get_type() macro is being defined to derive the device type from an lwIP network interface.


sanderssj (approved):
LGTM

PR #1940, opened 2023-09-12
VGA: check display buffer setting when probing device

sanderssj:
In UEFI systems using a framebuffer display there may still be a working vga pci device, but since it is not set up for text mode, writes to the display buffer can cause the system to hang. This change adds an additional check when probing a vga device to read the miscellaneous output register and check the "RAM Enable" bit, which is required to be set for accessing the display buffer from the system. This bit is cleared in UEFI systems with a framebuffer/GOP display.


francescolavra (commented):
LGTM. Tested with Qemu and OVMF UEFI firmware, I can confirm that bit 1 "RAM Enable" of misc output register is cleared

PR #1941, opened 2023-09-12
Klog: Ensure sufficient alignment for dump

sanderssj:
The klog_dump used by the kernel is a buffer passed directly to a storage handler, but some storage drivers like xen require at least sector alignment for the transaction's buffer. The dump was located in the bss at the end of the klog struct, so it had no alignment guarantees. This change moves it to the front of the struct and adds a 512-byte alignment requirement to the struct. This fixes an assertion failure when running klog_save on a xen instance.


francescolavra (commented):
LGTM

PR #1942, opened 2023-09-13
Network socket close/shutdown: fix race with lwIP error callback

francescolavra:
In the existing TCP network socket code there is a race condition that can be triggered when running on a multi-CPU machine, as shown by the following scenario:
- a thread on CPU0 calls close() on a connected TCP socket, the code in socket_close() retrieves the tcp_pcb struct and tries to lock it via netsock_tcp_get()
- the remote end of the TCP link closes the connection, which causes the lwip_tcp_conn_err callback function to be run on CPU1, which decrements the tcp_pcb reference count via tcp_unref()
- the code running on CPU0 acquires the tcp_pcb lock, calls tcp_close() (which is wrong because this function must not be called on a tcp_pcb that has been disconnected), then calls tcp_unref() (which is wrong because it duplicates the tcp_unref() call made by the callback function)

A similar race condition happens when shutdown(SHUT_RDWR) is called instead of close() (with the difference that the refcount is not decremented during the shutdown call, which is another problem that causes a memory leak when the race condition does not occur).

This change fixes the above issues by adding a netsock_tcp_close() helper function, which calls tcp_close() only if a tcp_pcb has not been disconnected (i.e. the error callback has not been called); this helper function, which acquires the netsock lock to avoid concurrent execution with the error callback, is called by both socket_close() and netsock_shutdown() (when the `how` argument is set to SHUT_RDWR). The error callback no longer drops the reference to the tcp_pcb, so this reference is only dropped when the socket is closed. In addition, the tcp_pcb state checks in netsock_shutdown() are now done with the tcp_pcb struct locked, for multi-thread safety.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -371,6 +371,17 @@ static void netsock_tcp_put(struct tcp_pcb * tcp_lw)
     tcp_unref(tcp_lw);
 }
 
+static void netsock_tcp_close(netsock s, struct tcp_pcb *tcp_lw)
+{
+    netsock_lock(s);
+    if (s->info.tcp.state != TCP_SOCK_UNDEFINED) {
+        tcp_close(tcp_lw);
+        tcp_arg(tcp_lw, 0);
+        s->info.tcp.state = TCP_SOCK_UNDEFINED;
+    }
+    netsock_unlock(s);
+}
+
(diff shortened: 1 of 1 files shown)


wjhun (approved):
LGTM

PR #1943, opened 2023-09-13
UEFI fixes

francescolavra:
This change set fixes a couple of issues preventing the kernel from booting on UEFI-enabled machines.

1. Since the addition of the string and vector tag types, the UEFI bootloader needs to include the code from the string.c and vector.c source files
2. The existing code in the pci_platform_init_bar() function for the PC platform is not handling 64-bit addresses, thus fails to properly read and assign addresses when the 64-bit flag is set for a given PCI BAR


sanderssj (approved):
LGTM

PR #1945, opened 2023-09-14
fix double free in each_http_request()

wjhun:
Commit 5301c6a7 introduced a regression whereby the "start_line" vector attribute of an http request would be explicitly freed, which would then later cause a double free during parser cleanup. This change corrects this issue by treating the vector as an immutable value (whereas previously it was cloned and thus consumable).


francescolavra (commented):
LGTM

PR #1947, opened 2023-09-19
feat(gcp): disk metrics

rinor:
**todo**:
- [x] attempt dynamic metrics body memory allocation based on number of volumes
- [x] maybe add manifest config to enable it (in case it's heavy and may not be interesting for stateless apps that don't use any storage at all)
  - _default_ **disabled** - `"disk": {}` to enable.
  - _default_ **readonly mount(s) are not reported** - `"include_readonly": "t"` to enable.
```json
  "ManifestPassthrough": {
    "gcp": {
      "metrics": {
        "interval": "60",
        "disk": {
          "include_readonly": "true"
        }
      }
    }
  }
```
- [ ] add ops docs

### DIFF
--- a/klib/gcp.c
+++ b/klib/gcp.c
@@ -5,6 +5,8 @@
 #include <mktime.h>
 #include <pagecache.h>
 #include <tls.h>
+#include <storage.h>
+#include <fs.h>
 
 #define GCP_MD_SERVER_ADDR  IPADDR4_INIT_BYTES(169, 254, 169, 254)
 
@@ -54,6 +56,8 @@ static struct gcp {
     closure_struct(gcp_log_vh, log_vh);
     buffer_handler log_resp_parser;
     timestamp metrics_interval;
+    boolean metrics_disk;
+    boolean metrics_disk_include_readonly;
     buffer metrics_url;
     boolean metrics_pending;
     struct timer metrics_timer;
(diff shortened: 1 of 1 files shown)


klib/gcp.c
```diff
@@ -711,7 +739,7 @@ static boolean gcp_metrics_post(void)
     set(req, sym(Accept), app_json);
     set(req, sym(Content-Type), app_json);
     boolean success = false;
-    buffer body = allocate_buffer(gcp.h, 2 * KB);
+    buffer body = allocate_buffer(gcp.h, 4 * KB);
```
rinor, line 742: handle this better:
- `memory: 1_712 bytes`
- `disk: ~ 1_200 bytes` x volume

rinor:
is this acceptable to be added to `kernel/storage` ?
```c
u64 storage_count(boolean rw, boolean ro)
{
    u64 count = 0;
    boolean fs_readonly = filesystem_is_readonly(storage.root_fs);
    if ((fs_readonly && ro) || (!fs_readonly && rw))
        count++;
    storage_lock();
    list_foreach(&storage.volumes, e) {
        volume v = struct_from_list(e, volume, l);
        if (v->fs) {
            fs_readonly = filesystem_is_readonly(v->fs);
            if ((fs_readonly && ro) || (!fs_readonly && rw))
                count++;
        }
    }
    storage_unlock();
    return count;
}
```
Alternatively I can implement this only where I need it by using `storage_iterate`

francescolavra:
> is this acceptable to be added to `kernel/storage` ?
> ...
> 
> Alternatively I can implement this only where I need it by using `storage_iterate`

No strong opinion, but I would rather not add it to the main kernel code if it's only used by a single klib.

klib/gcp.c
```diff
@@ -698,6 +701,45 @@ static void gcp_metrics_add_memory(buffer body, boolean first, const char *type,
...
+    int block_size = fs_blocksize(fs);
+    u64 total_blocks = fs_totalblocks(fs);
+    u64 free_blocks = fs_freeblocks(fs);
+    u64 percent_used_free = free_blocks * 100 / total_blocks;
+    buffer b = bound(b);
+    const char *interval = bound(interval);
+    gcp_metrics_add_disk(b, false, "bytes_used", "free", free_blocks * block_size, interval, label);
+    gcp_metrics_add_disk(b, false, "bytes_used", "used", (total_blocks - free_blocks) * block_size, interval, label);
+    gcp_metrics_add_disk(b, false, "percent_used", "free", percent_used_free, interval, label);
+    gcp_metrics_add_disk(b, false, "percent_used", "used", 100 - percent_used_free, interval, label);
```
francescolavra, line 731: A volume may have an empty label; in this case, I would use the volume UUID as device identifier, instead of an empty string. See for example the `mounts_handler` closure in src/unix/special.c

klib/gcp.c
```diff
@@ -698,6 +701,45 @@ static void gcp_metrics_add_memory(buffer body, boolean first, const char *type,
...
+}
+
+closure_function(2, 4, void, gcp_disk_count_vh,
+                u64 *, count, boolean, skip_readonly,
+                 u8 *, uuid, const char *, label, filesystem, fs, inode, mount_point)
+{
+    if (filesystem_is_readonly(fs) && bound(skip_readonly))
+        return;
+    bound(count)++;
```
francescolavra, line 740: This is incrementing the pointer, not the value being pointed to; to increment the value, you can write `(*bound(count))++;`

rinor: thanks, fixed.

klib/gcp.c
```diff
@@ -698,6 +702,57 @@ static void gcp_metrics_add_memory(buffer body, boolean first, const char *type,
...
+    u64 total_blocks = fs_totalblocks(fs);
+    u64 free_blocks = fs_freeblocks(fs);
+    u64 percent_used_free = free_blocks * 100 / total_blocks;
+    buffer b = bound(b);
+    const char *interval = bound(interval);
+    const char *device = 0;
+    if (label[0]) {
+        device = label;
+    } else if (uuid) {
+        buffer b_uuid = allocate_buffer(gcp.h, 2 * UUID_LEN + 4);
```
rinor, line 732: Is this correct/recommended buffer size in this case
- `2(hex string) * 16(UUID_LEN) + 4(dashes) = 36 bytes`
- `9e7db206-f94d-f0ca-0c98-dd1c404d47a8`? 

What is the rationale/best practice behind allocation sizes, since one way or another it ends up calling `buffer_extend`

klib/gcp.c
rinor: Checked that initially, but was not sure about the correct implementation, thanks for the example. Pushed some changes.

klib/gcp.c
rinor: Appreciate you taking the time to explain. Changes pushed. Thank you.

klib/gcp.c
```diff
@@ -698,6 +702,51 @@ static void gcp_metrics_add_memory(buffer body, boolean first, const char *type,
...
+    u64 total_blocks = fs_totalblocks(fs);
+    u64 free_blocks = fs_freeblocks(fs);
+    u64 percent_used_free = free_blocks * 100 / total_blocks;
+    buffer b = bound(b);
+    const char *interval = bound(interval);
+    buffer device = NULL;
+    if (label[0]) {
+        device = alloca_wrap_cstring(label);
+    } else if (uuid) {
+        buffer device = little_stack_buffer(2 * UUID_LEN + 4);
```
francescolavra, line 732: This is creating a new `device` local variable whose scope is limited to the `else if` block, instead of assigning a value to the outer `device` variable. It should be `device = little_stack_buffer(2 * UUID_LEN + 4);`
Also, no need for the `if (uuid)` part, and no need to initialize device to NULL (when the label is empty, the uuid is never null, and we cannot pass a null buffer to bprintf() anyway).

francescolavra (commented):
LGTM

PR #1948, opened 2023-09-25
Create only dual-stack udp pcbs

sanderssj:
If a udp pcb is created with udp_new(), it appears to default to being an ip4 pcb, which will then reject sending any ip6 traffic with an invalid argument because the pcb doesn't support ip6. This changes all udp_new calls to udp_new_ip_type(IPADDR_TYPE_ANY) which makes the pcb dual-stack. Ip6 udp sockets will now properly transmit
ip6 traffic. This also allows the udp netconsole and syslog to work with ip6 addresses.


francescolavra (commented):
LGTM

PR #1949, opened 2023-09-27
Aarch64 cpu_init(): enable userspace access to CNTVCT_EL0 register

francescolavra:
Aarch64 user programs may execute code that accesses system registers such as CNTVCT_EL0 (Counter-timer Virtual Count Register); for example the _armv7_tick function in the OpenSSL library (at https://github.com/openssl/openssl/blob/master/crypto/arm64cpuid.pl) reads such register.
With the existing kernel code, accessing the CNTVCT_EL0 register from the user program triggers an exception; this change modifies the aarch64 CPU initialization code so that EL0 accesses to the frequency register and virtual counter registers are not trapped; this fixes an issue where NodeJS v18.10.0 would fail during startup when run on aarch64.


wjhun (approved):
LGTM

PR #1956, opened 2023-10-24
chore(kernel/pagecache): PAGECACHE_NODE_DIRTY_COMMITS_QUEUE_SIZE 8

rinor:
Just to make it a bit more visible


rinor:
Queue is being removed and this makes no sense anymore.

PR #1969, opened 2023-11-20
x86: add AMD SEV (Secure Encrypted Virtualization) support

francescolavra:
This changeset adds support for confidential computing, by making the kernel able to run on AMD CPUs where the hypervisor enables SEV (memory encryption with a guest-private key).
Use of kernel heaps has been revised in order to differentiate between DMA-compatible and non-DMA-compatible memory (when running with SEV enabled, kernel and userspace memory is encrypted, while memory used for DMA transfers cannot be encrypted).
In addition, the gVNIC (Google Virtual Ethernet) device driver is being added to the x86 build of the kernel, because confidential VM instances in Google Cloud Platform are equipped with a gVNIC network adapter.

### DIFF
--- a/klib/mbedtls.c
+++ b/klib/mbedtls.c
@@ -273,7 +273,7 @@ int tls_connect(ip_addr_t *addr, u16 port, connection_handler ch)
 
 int init(status_handler complete)
 {
-    tls.h = heap_locked(get_kernel_heaps());
+    tls.h = heap_malloc();
     mbedtls_ssl_config_init(&tls.conf);
     mbedtls_ctr_drbg_init(&tls.ctr_drbg);
     mbedtls_entropy_init(&tls.entropy);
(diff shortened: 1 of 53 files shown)


eyberg (approved):

rinor:
are there any side-effects to be expected on **non SEV** deployments, like increased memory usage or anything else?

francescolavra:
The changes for non-SEV deployments should be minimal: we have a few more static variables and struct member fields here and there, but the increase in memory usage should be less than 100 bytes overall.
As for CPU usage, the only differences should be in the calculation of the bitmask to be applied to page table flags (which is no longer a hardcoded value but depends on the memory encryption mask that is set during initialization), and in the code to sync the page cache with disk contents (which needs to check whether bounce buffering should be done); but again the increase in resource usage should be negligible.

PR #1974, opened 2023-11-28
debugflags(reboot_on_exit_zero): reboot_on_exit_zero

rinor:
Needed this, because "some" may decide that `exit(0)` on some app failures is a good idea, and while testing, instances kept going down, even though `reboot_on_exit` was configured.

Nothing else changes and `idle_on_exit` **has priority over this** `reboot_on_exit_zero`

Feel free to drop it if not useful to others though.

### DIFF
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -606,7 +606,7 @@ void vm_exit(u8 code)
             u64 expected_code;
             if (u64_from_value(expected, &expected_code) &&
                 expected_code == code) {
-                code = 0;               
+                code = 0;
             } else if (is_vector(expected)) {
                 for (int i = 0; get_u64(expected, intern_u64(i), &expected_code); i++) {
                     if (expected_code == code) {
@@ -616,11 +616,15 @@ void vm_exit(u8 code)
                 }
             }
         }
-        if ((code != 0) && get(root, sym(reboot_on_exit)))
-            vm_reset();
-        if ((code == 0) && get(root, sym(idle_on_exit))) {
-            send_ipi(TARGET_EXCLUSIVE_BROADCAST, shutdown_vector);
-            machine_halt();
+        if (!shutting_down) {
+            if ((code != 0) && get(root, sym(reboot_on_exit)))
+                vm_reset();
+            if ((code == 0) && get(root, sym(idle_on_exit))) {
+                send_ipi(TARGET_EXCLUSIVE_BROADCAST, shutdown_vector);
+                machine_halt();
+            }
+            if ((code == 0) && get(root, sym(reboot_on_exit_zero)))
+                vm_reset();
         }
     }
     if (vm_halt && (!root || !get(root, sym(debug_exit)))) {


eyberg:
I'm ok w/this although the docs would need to be very clear that operations such as 'stop' and 'shutdown' aren't going to work w/this enabled, assuming you can still delete the vm on GCP with this enabled?

rinor:
> I'm ok w/this although the docs would need to be very clear that operations such as 'stop' and 'shutdown' aren't going to work w/this enabled, assuming you can still delete the vm on GCP with this enabled?

Right, all operations on clouds should "work" since instances will be "forcefully stopped" after a specific timeout. However, and this applies also to `reboot_on_exit`, upon shutdown the program may be started again, for a short time, and this may cause unnecessary long instance shutdown times, and at times program data corruption upon the final forced terminate.

Added a check to execute that code only when we are not `shutting_down`, I had this patch already deployed on the test instances earlier (because of the double program start), but missed it on the pr.

francescolavra:
Instead of adding another option, I would expand the usage of the existing reboot_on_exit option, turning it from a boolean flag to a more generic option that is used as a pattern against which the exit code is matched; this would reuse (and enhance) the code that is already there to parse the expected_exit_code option. So we could have for example reboot_on_exit=!0 to make the instance reboot for non-zero exit codes, reboot_on_exit=* to make the instance reboot regardless of the exit code, and so on. For backwards compatibility, reboot_on_exit=t should behave the same as reboot_on_exit=!0.
As for avoiding long instance shutdown time and forced instance stopping, using the shutting_down value isn't going to work in all cases, for example if the application doesn't have a SIGTERM signal handler it will be forcefully terminated by the kernel when an instance stop command is executed by the user, and shutting_down will not be set. I think the existing code should be modified so that an instance is always stopped when the user wants to stop it, regardless of the exit code and the presence of the reboot_on_exit option. I can come up with something to implement this.

rinor:
> Instead of adding another option, I would expand the usage of the existing reboot_on_exit option, turning it from a boolean flag to a more generic option that is used as a pattern against which the exit code is matched; this would reuse (and enhance) the code that is already there to parse the expected_exit_code option. So we could have for example reboot_on_exit=!0 to make the instance reboot for non-zero exit codes, reboot_on_exit=* to make the instance reboot regardless of the exit code, and so on. For backwards compatibility, reboot_on_exit=t should behave the same as reboot_on_exit=!0.

Since we want to keep compatibility with previous behavior, do we really care about `reboot_on_exit=!0`?
Can't we just handle `reboot_on_exit=*` and anything else would work as always?
I mean smth like this:

```c
            buffer reboot_on_exit = get(root, sym(reboot_on_exit));
            if (reboot_on_exit && (code || !buffer_strcmp(reboot_on_exit, "*")))
                vm_reset();

```

the whole block would look smth like:

```c
        if (!(shutting_down & SHUTDOWN_POWER)) {
            buffer reboot_on_exit = get(root, sym(reboot_on_exit));
            if (reboot_on_exit && (code || !buffer_strcmp(reboot_on_exit, "*")))
                vm_reset();
            if ((code == 0) && get(root, sym(idle_on_exit))) {
                send_ipi(TARGET_EXCLUSIVE_BROADCAST, shutdown_vector);
                machine_halt();
            }
        }
```

PR #1975, opened 2023-11-29
feat(shutdown): record shutdown sources

rinor:
Not sure if this is the best way, but it should work for clean differentiation of shutdown origin when checking/executing  hooks like `reboot_on_exit` so they don't interfere on "normal" shutdown/stop requests

### DIFF
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -548,7 +548,6 @@ closure_function(0, 2, void, storage_shutdown, int, status, merge, m)
     storage_sync(apply_merge(m));
 }
 
-extern boolean shutting_down;
 void __attribute__((noreturn)) kernel_shutdown(int status)
 {
     heap locked = heap_locked(init_heaps);
@@ -557,7 +556,11 @@ void __attribute__((noreturn)) kernel_shutdown(int status)
     status_handler sh = apply_merge(m);
     shutdown_handler h;
 
-    shutting_down = true;
+    /* Set shutting_down to prevent user threads from being scheduled
+     * and then try to interrupt the other cpus back into runloop
+     * so they will idle while running kernel_shutdown */
+    shutting_down |= SHUTDOWN_ONGOING;
+    wakeup_or_interrupt_cpu_all();
 
     if (root_fs)
         vector_push(shutdown_completions,
@@ -606,7 +609,7 @@ void vm_exit(u8 code)
             u64 expected_code;
             if (u64_from_value(expected, &expected_code) &&
                 expected_code == code) {
-                code = 0;               
+                code = 0;
             } else if (is_vector(expected)) {
                 for (int i = 0; get_u64(expected, intern_u64(i), &expected_code); i++) {
                     if (expected_code == code) {
@@ -616,11 +619,13 @@ void vm_exit(u8 code)
                 }
             }
         }
-        if ((code != 0) && get(root, sym(reboot_on_exit)))
-            vm_reset();
-        if ((code == 0) && get(root, sym(idle_on_exit))) {
-            send_ipi(TARGET_EXCLUSIVE_BROADCAST, shutdown_vector);
-            machine_halt();
+        if (!(shutting_down & SHUTDOWN_POWER)) {
+            if ((code != 0) && get(root, sym(reboot_on_exit)))
+                vm_reset();
+            if ((code == 0) && get(root, sym(idle_on_exit))) {
+                send_ipi(TARGET_EXCLUSIVE_BROADCAST, shutdown_vector);
+                machine_halt();
+            }
         }
     }
     if (vm_halt && (!root || !get(root, sym(debug_exit)))) {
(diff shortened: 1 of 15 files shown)


src/kernel/init.c
```diff
@@ -579,6 +579,7 @@ void __attribute__((noreturn)) kernel_shutdown(int status)
     while(1);
 }
 
+extern boolean unix_shutting_down;
```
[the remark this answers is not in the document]
rinor: make sense, thanks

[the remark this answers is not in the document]
rinor: Pushed some changes to make sure we are on the same page.

src/kernel/kernel.h
```diff
@@ -55,7 +55,11 @@ void init_kernel_context(kernel_context kc, int type, int size, queue free_ctx_q
 #define cpu_interrupt 3
 #define cpu_user 4
 
-extern boolean shutting_down;
+#define SHUTDOWN_POWER (1<<0) // shutdown triggered externally
+#define SHUTDOWN_TERM  (1<<1) // process termination triggered by the kernel as a result of a signal
+#define SHUTDOWN_EXIT  (1<<2) // process termination triggered by the process itself
```
francescolavra, line 60: Thinking more about it, we can do without the SHUTDOWN_TERM and SHUTDOWN_EXIT flags (in the future, perhaps we could incorporate this info in the kernel_shutdown() arguments); instead, we need a SHUTDOWN_ONGOING flag, which will be set in kernel_shutdown(), and will be checked instead of checking the existing boolean variable.
So all the callers to kernel_shutdown() don't need to touch the `shutting_down` variable.

src/xen/xen.c
```diff
@@ -388,7 +388,7 @@ closure_function(1, 0, void, xen_per_cpu_init,
     xen_setup_vcpu(current_cpu()->id, bound(shared_info_phys));
 }
 
-void unix_shutdown(void);
+void kernel_powerdown(void);
```
francescolavra, line 391: Ditto.

src/unix/coredump.c
```diff
@@ -226,7 +226,7 @@ void coredump(thread t, struct siginfo *si, status_handler complete)
     status s = STATUS_OK;
 
     /* stop running any other user threads */
-    shutting_down = true;
+    shutting_down |= SHUTDOWN_TERM;
     wakeup_or_interrupt_cpu_all();
```
francescolavra, line 230: These lines should be removed, similarly to what is done in exit_group(). But in order for the core dump to be executed after ensuring that user threads are not running, the core dump needs to be executed as a shutdown completion. So default_signal_action(), instead of calling directly coredump(), should instantiate a shutdown_handler (which will call coredump()) and call add_shutdown_completion(), without calling runloop() afterwards; so these lines:
```
        coredump(t, &qs->si, closure(h, after_dump_halt, t, signum));
        clear_fault_handler();
        runloop();
        return;
```
will become something like:
```
        add_shutdown_completion(closure(h, coredump_handler, t, &qs->si));
        clear_fault_handler();
        break;
```

rinor: this is the last part yet to be implemented

src/kernel/init.c
```diff
@@ -557,7 +556,11 @@ void __attribute__((noreturn)) kernel_shutdown(int status)
     status_handler sh = apply_merge(m);
     shutdown_handler h;
 
-    shutting_down = true;
+    shutting_down |= SHUTDOWN_ONGOING;
+    /* Set shutting_down to prevent user threads from being scheduled
+     * and then try to interrupt the other cpus back into runloop
+     * so they will idle while running kernel_shutdown */
```
francescolavra, line 562: This comment should go above the `shutting_down |= SHUTDOWN_ONGOING` line

src/unix/coredump.c
rinor: would smth like this be correct
```diff
@@ -1036,6 +1036,18 @@ closure_function(2, 1, void, after_dump_halt,
     closure_finish();
 }
 
+closure_function(4, 2, void, coredump_shutdown_handler,
+                 heap, h, thread, t, struct siginfo *, si, int, signum,
+                 int, status, merge, m)
+{
+    heap h = bound(h);
+    thread t = bound(t);
+    struct siginfo *si = bound(si);
+    int signum = bound(signum);
+
+    coredump(t, si, closure(h, after_dump_halt, t, signum));
+}
+
 static void default_signal_action(thread t, queued_signal qs)
 {
     heap h = heap_locked(get_kernel_heaps());
@@ -1070,10 +1082,9 @@ static void default_signal_action(thread t, queued_signal qs)
     case SIGTRAP:
     case SIGXCPU:
     case SIGXFSZ:
-        coredump(t, &qs->si, closure(h, after_dump_halt, t, signum));
+        add_shutdown_completion(closure(h, coredump_shutdown_handler, h, t, &qs->si, signum));
         clear_fault_handler();
-        runloop();
-        return;
+        break;
 
     case SIGSTOP:
     case SIGTSTP:
```

rinor: do we need `clear_fault_handler();` there still?

francescolavra (approved):

rinor:
@francescolavra thank you for the feedback, guidance and patience. Learned a lot in the process.

PR #1977, opened 2023-12-02
feat(reboot_on_exit): additional triggering patterns

rinor:
based on comments from https://github.com/nanovms/nanos/pull/1974

- reboot on **non zero exit code** - use `"t"` or `"!0"`
```json
{
  "ManifestPassthrough": {
    "reboot_on_exit": "t"
  }
}
```
or
```json
{
  "ManifestPassthrough": {
    "reboot_on_exit": "!0"
  }
}
```
- reboot on **any exit code** - use `"*"`
```json
{
  "ManifestPassthrough": {
    "reboot_on_exit": "*"
  }
}
```
- reboot on a **single exit code** - use `"replace_me_with_exit_code"`
```json
{
  "ManifestPassthrough": {
    "reboot_on_exit": "0"
  }
}
```
- reboot on a **list of exit codes**
```json
{
  "ManifestPassthrough": {
    "reboot_on_exit": ["0","1","6"]
  }
}
```
---

**Note**: `ops` behavior remains unchanged for `"RebootOnExit": true` or   `"Debugflags": ["reboot_on_exit"]`

---

**Question**: Do we want to add other negating possibilities other than `!0`, i.e: exit on any code other than`!code` where code can be any single value? - **yes**

### DIFF
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -582,6 +582,39 @@ void __attribute__((noreturn)) kernel_shutdown(int status)
     while(1);
 }
 
+static boolean vm_exit_match(u8 exit_code, tuple config, symbol option, boolean match_on_t)
+{
+    if (!config)
+        return false;
+
+    value config_option = get(config, option);
+    if (!config_option)
+        return false;
+
+    u64 action_code;
+    if (is_string(config_option)) {
+        if (!buffer_strcmp(config_option, "*"))
+           return true;
+
+        boolean neq = false;
+        if (peek_char(config_option) == '!') {
+            neq = true;
+            buffer_consume(config_option, 1);
+        }
+        if (u64_from_value(config_option, &action_code)) {
+            return (((action_code == exit_code) && !neq) ||
+                    ((action_code != exit_code) && neq));
+        }
+    } else if (is_vector(config_option)) {
+        for (int i = 0; get_u64(config_option, intern_u64(i), &action_code); i++) {
+            if (action_code == exit_code)
+                return true;
+        }
+        return false;
+    }
+    return match_on_t;
+}
+
 void vm_exit(u8 code)
 {
 #ifdef SMP_DUMP_FRAME_RETURN_COUNT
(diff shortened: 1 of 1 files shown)


src/kernel/init.c
```diff
@@ -620,8 +621,27 @@ void vm_exit(u8 code)
...
+            // reboot_on_exit
+            value reboot = get(root, sym(reboot_on_exit));
+            if (reboot) {
+                u64 reboot_code;
+                if (u64_from_value(reboot, &reboot_code)) {
+                    if (reboot_code == code)
+                        vm_reset();
+                } else if (is_vector(reboot)) {
+                    for (int i = 0; get_u64(reboot, intern_u64(i), &reboot_code); i++) {
+                        if (reboot_code == code)
```
francescolavra, line 633: Instead of duplicating the code that parses the expected_exit_code option, I would create a `boolean vm_exit_match(u8 exit_code, tuple config, symbol option, boolean match_on_t)` generic function that does the parsing and returns where there is a match (if the option value is "t", it returns the value of `match_on_t`). So vm_exit() can do something like this:
```
if (vm_exit_match(code, root, sym(expected_exit_code), false))
    code = 0;
if (!(shutting_down & SHUTDOWN_POWER)) {
    if (vm_exit_match(code, root, sym(reboot_on_exit), code != 0))
        vm_reset();
```
And yes, I would make vm_exit_match() able to parse a generic `!code` expression where code can be any single value, not just 0.

src/kernel/init.c
```diff
@@ -604,6 +604,7 @@ void vm_exit(u8 code)
 
     tuple root = get_root_tuple();
     if (root) {
+        // expected_exit_code
```
francescolavra, line 607: I wouldn't add these comments, I don't think they make the code easier to read, they are just duplicating the `sym()` string that is in the next line.

francescolavra (changes requested):

src/kernel/init.c
rinor: thanks for the advice. Pushed some changes and now all 3 config options ( `expected_exit_code`, `reboot_on_exit` and `idle_on_exit`) benefit from the same enhanced logic (while retaining the existing default behavior).

While matching vector is supported ["0","1","6"], I did not implement negating vector ["!0", "!1", "!6"]. Do we want it?

francescolavra (approved):

PR #1979, opened 2023-12-05
AWS: add support for PTP hardware clock

francescolavra:
This change set enhances the AWS ENA network interface driver to add support for retrieving the current time from a PTP hardware clock, when supported by the network interface.
In addition, the NTP klib has been enhanced to be able to retrieve time from the PTP clock alternatively to using NTP; if the root tuple contains a "chrony" tuple that has a "refclock" attribute with "ptp" value, then the NTP klib uses the PTP clock for synchronizing the system time; if a PTP clock is not available, the klib falls back to NTP.
These features can be used on selected AWS instances (currently, from the R7g (ARM Graviton3) instance family in the Asia Pacific (Tokyo) region) to synchronize the system time with the PTP clock.
The first 5 commits fix a few issues encountered when running on AWS R7g instances, thus making the kernel able to run on ARM instances equipped with Graviton3 processors.
Example Ops configuration to enable synchronization of system time with the PTP hardware clock:
```
{
  "Klibs": ["ntp"]
  },
  "CloudConfig" : {
    "ProjectID": "my-project",
    "Zone": "ap-northeast-1a",
    "BucketName": "my-bucket",
    "Flavor": "r7g.medium"
  },
  "ManifestPassthrough": {
    "chrony": {"refclock": "ptp"}
  }
}
```

### DIFF
--- a/klib/ntp.c
+++ b/klib/ntp.c
@@ -79,21 +79,20 @@ typedef struct ntp_server {
     struct ntp_ts last_originate_time;
 } *ntp_server;
 
-declare_closure_struct(0, 2, void, ntp_query_func,
-                       u64, expiry, u64, overruns);
 declare_closure_struct(0, 2, void, ntp_slew_complete_func,
                        u64, expiry, u64, overruns);
 declare_closure_struct(0, 2, void, ntp_raw_update_func,
                        u64, expiry, u64, overruns);
 
 static struct {
+    boolean ptp_clock;
     vector servers;
     int current_server;
     struct udp_pcb *pcb;
     struct timer query_timer;
     struct timer slew_timer;
     struct timer raw_update_timer;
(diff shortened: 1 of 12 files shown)


ernestomedina17:
Then Graviton2-based instances like t4g.nano do not support this feature in AWS?

francescolavra:
> Then Graviton2-based instances like t4g.nano do not support this feature in AWS?

Correct, according to https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-time-sync-service-microsecond-accurate-time/ the PTP hardware clock is only available in the Asia Pacific (Tokyo) Region on R7g instances. Their statement says "we will be expanding support to additional AWS Regions and EC2 Instance types", but there are no further details as to what instance types will be covered.

eyberg (approved):

PR #1983, opened 2023-12-13
Kernel command line: add support for arbitrary configuration settings

### LINKED ISSUE #1976: allow passing in network settings via kernel boot args for firecracker
currently images produced for firecracker either need dhcp or if using static need to be re-built - this option would allow the user to supply static w/out having to re-build the image

https://github.com/nanovms/ops/discussions/1554

this should override what might already be in the manifest

francescolavra:
This patch set adds support for changing arbitrary settings in the root tuple via the kernel command line (which is retrieved by the kernel when booting under AWS Firecracker on x86).
With these changes, it is possible for example to override the network settings when starting an instance from a given image (without modifying the image itself) by specifying those settings in the kernel command line ("boot_args" parameter in the Firecracker configuration file), as in the following example:
"en1.ipaddr=10.3.3.6 en1.netmask=255.255.0.0 en1.gateway=10.3.0.1".
In the above example, "en1" identifies the first network interface; if multiple interfaces are used (en2, en3, etc.), each of them can be configured independently.
Example to configure a static IPv6 address on the first network interface:
"en1.ip6addr=20::A8FC:FF:7600:AA"
Example to add an environment variable (or override its value if the variable is already present in the image):
"environment.VAR_NAME=VAR_VALUE"
Example to modify the program arguments:
"arguments.0=/bin/my-program arguments.1=--my-option"

The first 3 commits are fixes for issues that have been exposed by the last 2 commits; the fix in the first commit allows all available physical memory ranges to be used by the kernel, without requiring 2MB alignment of the range start.

Closes #1976

### DIFF
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -613,7 +613,7 @@ static boolean vm_exit_match(u8 exit_code, tuple config, symbol option, boolean 
             return (((action_code == exit_code) && !neq) ||
                     ((action_code != exit_code) && neq));
         }
-    } else if (is_vector(config_option)) {
+    } else if (is_composite(config_option)) {
         for (int i = 0; get_u64(config_option, intern_u64(i), &action_code); i++) {
             if (action_code == exit_code)
                 return true;
--- a/src/kernel/init.c
+++ b/src/kernel/init.c
@@ -193,6 +193,8 @@ closure_function(2, 2, void, fsstarted,
     wrapped_root = tuple_notifier_wrap(fs_root, true);
     assert(wrapped_root != INVALID_ADDRESS);
     tuple root = (tuple)wrapped_root;
+    boot_params_apply(root);
+    reclaim_regions();  /* for pc: no accessing regions after this point */
     tuple mounts = get_tuple(root, sym(mounts));
     if (mounts)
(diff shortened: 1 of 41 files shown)


src/kernel/init.c
```diff
@@ -227,10 +229,111 @@ closure_function(2, 2, void, fsstarted,
...
+    int prefix_len = cmdline_get_prefix(opt_start, opt_len);
+    if (prefix_len && (prefix_len < name_end - opt_start)) {
+        symbol s = intern(alloca_wrap_buffer(opt_start, prefix_len));
+        value child = get(cfg, s);
+        if (!child) {
+            child = allocate_tuple();
+            assert(child != INVALID_ADDRESS);
+            set(cfg, s, child);
+        }
+        if (is_tuple(child) || is_vector(child)) {
```
rinor, line 306:
```suggestion
        if (is_composite(child)) {
```

francescolavra: Good catch, change applied.

francescolavra:
I added 2 commits to fix concurrency issues observed when running CI tests.

eyberg (approved):

PR #1985, opened 2023-12-20
virtio-net: add support for segmentation offload

francescolavra:
This patchset enhances the virtio-net driver to support TCP segmentation and UDP fragmentation offload, which allows the virtIO device to merge multiple received packets into one large packet; in addition, the driver now supports the MRG_RXBUF feature, so that it is not necessary for the driver to allocate large RX buffers in order to receive large packets.
These changes bring a performance improvement to applications receiving large amounts of data over a network link, e.g. when a VM uses virtual network interfaces backed by tap interfaces on the host.
The first commit is a fix to the network socket code that was causing a performance penalty in data transfers on TCP connections with the window scaling option enabled.

### DIFF
--- a/src/net/direct.c
+++ b/src/net/direct.c
@@ -72,12 +72,26 @@ define_closure_function(1, 0, void, direct_receive_service,
             dc->receive_bh = bh;
         }
         boolean client = (d->p == 0);
+        input_buffer_handler ibh = dc->receive_bh;
         while (true) {
             struct pbuf *p = dequeue(dc->receive_queue);
             if (p == INVALID_ADDRESS)
                 break;
             if (p) {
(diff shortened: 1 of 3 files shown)


eyberg (approved):

rinor:
will investigate further, but it looks like this is breaking some stuff on clouds. 
GCP metrics for one is not working anymore (no errors, no warnings).

francescolavra:
> GCP metrics for one is not working anymore (no errors, no warnings).

Thanks for reporting the issue, The code used by the GCP klib to communicate with the monitoring server was indeed not prepared to handle input packets with segmentation offload. This is now fixed in the updated PR, and specifically in commit https://github.com/nanovms/nanos/pull/1985/commits/3ef3181a34dff24adf4d010f32fc9ecf9ebf802a

Issue #1885, opened 2023-06-30
question: idiomatic/safe method to detect nanos & klibs at runtime


rinor:
1. what would be an alternate idiomatic/safe method for an app (process) to **detect at runtime if it's running on nanos**, without making use of the environment variable(s) injected by ops/manually? - `not a feature request and not important`

2. what would be an alternate idiomatic/safe method for an app (process) to **detect at runtime the enabled klibs with the respective features**, without making use of the environment variable(s) injected by ops/manually? - `not a feature request and not important`

francescolavra:
1. A process can detect whether it's running on Nanos via the [uname()](https://man7.org/linux/man-pages/man2/uname.2.html) syscall; more specifically, the `sysname` field of `struct utsname` will contain the string "Nanos". Example:
```
$ ops run /bin/uname -a "-s"
running local instance
booting /home/francesco/.ops/images/uname ...
en1: assigned 10.0.2.15
Nanos
```
2. Currenly the kernel does not expose information about enabled klibs to the process

rinor:
thank you

rinor:
```sh
running local instance
...
en1: assigned 10.0.2.15
Nanos version: "Nanos 5.0-0.1.45 888809a7a8f0003b66a8ec2af2f5e8dd749866d0 x86_64"
```
- sample code - `go`
```go
package main

import (
	"fmt"
	"syscall"
)

func main() {
	fmt.Printf("Nanos version: %q\n", Version())
}

func Version() string {
	var uts syscall.Utsname
	if err := syscall.Uname(&uts); err != nil {
		return ""
	}

	sysname := int8ToString(uts.Sysname[:])
	release := int8ToString(uts.Release[:])
	version := int8ToString(uts.Version[:])
	machine := int8ToString(uts.Machine[:])

	return sysname + " " + release + " " + version + " " + machine
}

func int8ToString(s []int8) string {
	b := make([]byte, 0, len(s))
	for _, v := range s {
		if v == 0x00 {
			break
		}
		b = append(b, byte(v))
	}
	return string(b)
}
```

PR #2008, opened 2024-03-11
uname: add configurability of sysname and release

francescolavra:
With this change, it is possible to configure the kernel to insert arbitrary strings in the `sysname` and `release` fields of the struct used in the uname syscall. This allows running user programs that expect these fields to have specific values (e.g. programs that expect the sysname to be "Linux", or programs that parse the release string to infer a Linux kernel version).
To configure the sysname and release fields, a "uname" configuration tuple is used: this tuple can contain a "sysname" string and a "release" string, which allow configuring the respective fields; if a given string is not present in the configuration, its default value is used (i.e., "Nanos" for the sysname field, and "5.0-<NANOS_VERSION>" for the release field). Example snippet of Ops configuration file:
```
"ManifestPassthrough": {
  "uname": {
    "sysname": "Linux",
    "release": "5.0.5"
  }
}
```

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -1718,7 +1718,6 @@ sysreturn lseek(int fd, s64 offset, int whence)
 
 sysreturn uname(struct utsname *v)
 {
-    char sysname[] = "Nanos";
     char machine[] =
 #ifdef __x86_64__
         "x86_64";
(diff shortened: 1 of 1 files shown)


rinor:
How would we detect that we are running on **nanos** if `sysname` gets changed then, since this https://github.com/nanovms/nanos/issues/1885#issuecomment-1614336887 is not reliable anymore

francescolavra:
The default behavior is not changing, i.e. if the user doesn't override the sysname, the kernel will continue to output "Nanos".
If the user does override the sysname, I would assume they know what they are doing.

PR #2024, opened 2024-05-17
scheduler: add support for CPU affinity (thread pinning)

francescolavra:
This change set enhances the scheduler so that it honors the CPU affinity of user threads (set via the sched_setaffinity syscall).
In addition, it amends the CircleCI configuration so that automated tests are run on 2-vCPU instances, in order to increase test
coverage of SMP-related kernel code. This increased test coverage revealed a few issues in the RISC-V-specific code and in the Unix code (Unix domain sockets, poll/select, and robust futexes), which have been fixed as part of this change set.


francescolavra:
I added 2 commits that fix SMP-related issues found in the poll/select implementation and the robust futex handling logic.

eyberg (approved):

PR #2031, opened 2024-06-16
rm nightly x86 mac runner; change default mac runner to arm

eyberg:


.circleci/config.yml
```diff
@@ -95,7 +95,8 @@ jobs:
 
   build-mac:
     macos:
-        xcode: "12.2.0"
+        xcode: "15.4.0"
+    resource_class: macos.m1.medium.gen1
```
francescolavra, line 99: This CI job won't run successfully on ARM, it was written to run on x86 (e.g. it installs the x86 assembler and binutils, and uses an x86 target root filesystem).
Anyway, this job is currently never triggered (it's not part of the commit workflow), so this change won't break the CI tests.

eyberg: k, i think we can just yank this for now

PR #2032, opened 2024-06-23
On-demand program file paging: fix initialization of BSS areas

francescolavra:
When on-demand paging of the program file is enabled, BSS areas in pages faulted-in on demand are zeroed in-place, i.e. a newly mapped page retrieved via the page cache is zeroed starting from the BSS offset set up when initializing the relevant vmap. This creates a problem if the page contains other data (e.g. from another loadable section of the program) at or after the BSS offset, in which case this data would be overwritten.
This change fixes the above issue by using a separate page (instead of the page from the page cache) where the initialized program data (located before the BSS offset) is copied from the page cache page, and the rest of the page (starting at the BSS offset) is zeroed out. Closes https://github.com/nanovms/ops/issues/1629.

The last commit fixes an assertion failure that occurs when a page fault during a kernel context cannot be resolved synchronously.

### DIFF
--- a/src/aarch64/interrupt.c
+++ b/src/aarch64/interrupt.c
@@ -240,7 +240,6 @@ void synchronous_handler(void)
             context_release_refcount(retctx);
             frame_return(retctx->frame);
         }
-        assert(!is_kernel_context(ctx));
         runloop();
     } else {
         console("\nno fault handler for frame ");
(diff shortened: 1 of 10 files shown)


francescolavra:
I fixed two issues that could potentially cause the kernel to fail to release a lock when under memory pressure.

Issue #2051, opened 2024-08-15
Create unnamed O_TMPFILE fails in VirtFS (`error 22(EINVAL): Invalid argument`)


aleasims:
# Description

When I'm trying to create unnamed temporary file `O_TMPFILE` in directory mounted with VirtFS, it fails with `error 22: invalid argument`.

# Reproduce

Consider following C code :

**`main.c`:**

```c
#define _GNU_SOURCE

#include <fcntl.h>
#include <stdio.h>

int main()
{
    int fd = open("/tmp", O_TMPFILE | O_RDWR);
    printf("fd = %d\n", fd);
    return 0;
}
```

Compile code and run with `ops`, mounting directory through VirtFS:

```
gcc -o main main.c && ops run --mounts /tmp:/tmp ./main
```

**Expected result**

Program works successfully, just like on native Linux platform: unnamed tmp file is created and removed at the end of execution. Printed `fd` is positive:

```
running local instance
booting /home/alea/.ops/images/main ...
FS: filesystem_read_entire: t 0x7fc601c4, bufheap 0x00017054, buffer_handler 0x7fbdd000, status_handler 0x7fc609b4
FS: read_entire_complete: status [invalid format %v], addr 0x00000000
en1: assigned 10.0.2.15
fd = 3
```

**Actual result**

Returned `fd` is `-1`:

```
running local instance
booting /home/alea/.ops/images/main ...
FS: filesystem_read_entire: t 0x7fc601c4, bufheap 0x00017054, buffer_handler 0x7fbdd000, status_handler 0x7fc609b4
FS: read_entire_complete: status [invalid format %v], addr 0x00000000
en1: assigned 10.0.2.15
fd = -1
```

# Setup

```
$ ops profile
Ops version: 0.1.41
Nanos version: 0.1.51
Qemu version: 6.2.0
OS: linux
Arch: amd64
Virtualized: false
```

I tested the same with --nightly and with local build of master branch and the same error occurs.

# What I found

It looks like to me that `open()` syscall results in [calling `filesystem_creat_unnamed` function](https://github.com/nanovms/nanos/blob/0.1.51/src/unix/syscall.c#L924-L925), and 9pfs implementation [fails immediately](https://github.com/nanovms/nanos/blob/0.1.51/src/fs/9p.c#L616-L617) if asked to create unnamed file.

[...]

PR #2054, opened 2024-08-21
Filesystem: retire fs_status enum in favor of standard error codes

francescolavra:
Having filesystem-specific error codes requires conversion between filesystem codes and POSIX standard codes every time a filesystem- related error is reported to userspace. For the 9P filesystem, in some cases a double conversion (from a standard code to a filesystem code and then back to a standard code) may need to be done.
This change removes the fs_status enum that declares the filesystem error codes, and uses standard POSIX codes to report filesystem errors; this allows eliminating the above conversions. Standard error codes are defined in the new kernel/errno.h header file; in order to prevent this file from being included from userspace code, the kernel directory is being removed from the include paths in the relevant Makefiles; this required some adjustments to other header files so that they can be included by both kernel and non- kernel code.
For the 9P filesystem, the p9_create() callback function is being modified so that EOPNOTSUPP is returned instead or EINVAL if the user program tries to open a file with the O_TMPFILE flag (#2051).

### DIFF
--- a/klib/shmem.c
+++ b/klib/shmem.c
@@ -16,18 +16,18 @@ sysreturn memfd_create(const char *name, unsigned int flags)
     filesystem fs = shmem.fs;
     fsfile fsf;
     filesystem_lock(fs);
-    fs_status fss = filesystem_creat_unnamed(fs, &fsf);
-    if (fss == FS_STATUS_OK) {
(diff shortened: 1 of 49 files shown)


aleasims:
Checked it locally, everything regarding error codes for O_TMPFILEs works perfect, thanks!

PR #2068, opened 2024-09-24
fix: correct socket state checks and handle listen errors in netsyscall

rinor:


src/net/netsyscall.c
```diff
@@ -2088,7 +2088,12 @@ static sysreturn netsock_listen(struct sock *sock, int backlog)
         }
         goto unlock_out;
     }
-    struct tcp_pcb * lw = tcp_listen_with_backlog(s->info.tcp.lw, backlog);
+    err_t err;
+    struct tcp_pcb * lw = tcp_listen_with_backlog_and_err(s->info.tcp.lw, backlog, &err);
+    if (!lw) {
+        rv = lwip_to_errno(err);
+        goto unlock_out;
+    }
```
francescolavra, line 2096: Right, if the SO_REUSEADDR socket option is enabled, the bind() system call doesn't return -EADDRINUSE if another socket is bound to the same address, so we have to check for in-use addresses in the listen() system call. This is the correct fix.

francescolavra (approved):

PR #2070, opened 2024-09-26
feat: add address handling for unspecified IPs in connect function

rinor:
Added a utility function to set loopback addresses for unspecified IPv4 and IPv6 addresses during socket connection attempts. This ensures proper handling of "0.0.0.0" and "::" addresses.

### DIFF
--- a/src/net/netsyscall.c
+++ b/src/net/netsyscall.c
@@ -1684,6 +1684,19 @@ static inline sysreturn connect_tcp(netsock s, const ip_addr_t* address,
     return rv;
 }
 
+static inline void set_loopback_if_any(ip_addr_t *ipaddr)
+{
+    if (ipaddr->type == IPADDR_TYPE_V4) {
+        // Check if the IPv4 address is "0.0.0.0"
+        if (ip4_addr_isany(ip_2_ip4(ipaddr)))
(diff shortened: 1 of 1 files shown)


francescolavra:
Connecting to 0.0.0.0 or [::] is non-standard practice, and as such I'm a little reluctant to add support for it. Is there an application out in the wild that uses this and you want to run on Nanos?

rinor:
> Connecting to 0.0.0.0 or [::] is non-standard practice, and as such I'm a little reluctant to add support for it. Is there an application out in the wild that uses this and you want to run on Nanos?

Right, strictly speaking those are not valid as a destination address,  only as a source address, but based on what i.e linux is doing (copies source address to destination address, and if the packet does not have a source address, both are set to the loopback address), it's being (ab)used.

I encountered this while trying to build some stuff with https://github.com/ergo-services/examples.

I understand the reluctance and agree, however we could put that behind a build flag leaving it disabled by default, assuming there is any interest at all, otherwise I don't mind closing this simce the correct solution would be fixing the application.

francescolavra:
I would prefer not to add this functionality for the time being. If its use by existing applications turns out to be more widespread, we could revisit this.
Thanks

PR #2071, opened 2024-09-30
fix doc and build issue (for linux)

peppergrayxyz:
**Build is not working due to following issues:**

- ```go``` is not documented as dependency in the description for Linux
```bash
GO	/home/user/repos/nanos/output/test/runtime/bin/webg
/bin/sh: line 1: go: command not found
```
- there is an issue with a new line that is not printed due to ```echo``` not processing ```\n```
```bash
VERSION	/home/user/repos/nanos/output/platform/pc/gitversion.c
CC	/home/user/repos/nanos/output/platform/pc/output/platform/pc/gitversion.o
/home/user/repos/nanos/output/platform/pc/gitversion.c:1:21: error: extra tokens at end of #include directive [-Werror]
    1 | #include <runtime.h>\nconst sstring gitversion = ss_static_init("243c5a2010916b8fd69f86b290d637ad0efc1dbd");
      |    
```

**Test-Build is not working due to following issues:**

- ```ops``` is not documented as dependency in the description for Linux
```bash
=== RUN   TestE2E/packages/node_alloc
    e2e.go:181: Waiting for command to complete...
    e2e.go:207: Output: /bin/bash: line 1: ops: command not found
    e2e.go:208: ops exit code 127
```

- ```udp_test.c``` doesn't check for parse errors 
```bash
CC	/home/user/repos/nanos/output/test/unit/test/unit/udp_test.o
In file included from /usr/include/endian.h:35,
                 from /usr/include/sys/types.h:176,
                 from /usr/include/bits/socket.h:29,
                 from /usr/include/sys/socket.h:33,
                 from /usr/include/netinet/in.h:23,
                 from /usr/include/arpa/inet.h:22,
                 from /home/user/repos/nanos/test/unit/udp_test.c:8:
In function ‘__bswap_16’,
    inlined from ‘main’ at /home/user/repos/nanos/test/unit/udp_test.c:57:21:
/usr/include/bits/byteswap.h:37:10: error: ‘dport’ may be used uninitialized [-Werror=maybe-uninitialized]
   37 |   return __builtin_bswap16 (__bsx);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
/home/user/repos/nanos/test/unit/udp_test.c: In function ‘main’:
/home/user/repos/nanos/test/unit/udp_test.c:31:9: note: ‘dport’ was declared here
   31 |     u16 dport;
      |         ^~~~~
```

**Debug-Build is not working due to following issues:**

- ```storage.c``` prints a parameter, that doesn't exist

 ```bash
CC	/home/user/repos/nanos/output/platform/pc/src/kernel/storage.o
/home/user/repos/nanos/src/kernel/storage.c: In function ‘volume_add’:
/home/user/repos/nanos/src/kernel/storage.c:274:45: error: ‘size’ undeclared (first use in this function)
  274 |     storage_debug("new volume (%ld bytes)", size);
      |                                             
```

- ```unix.c``` has an improper macro expansion 
 
 ```bash
CC	/home/pepper/dev/repos/nanos/output/platform/pc/src/unix/unix.o
/home/pepper/dev/repos/nanos/src/unix/unix.c: In function ‘handle_protection_fault’:
/home/pepper/dev/repos/nanos/src/unix/unix.c:241:61: error: macro "ss" passed 11 arguments, but takes just 1
  241 |         pf_debug(format_protection_violation(vaddr, ctx, vm));
      |                                                             ^
In file included from /home/pepper/dev/repos/nanos/src/runtime/runtime.h:54,
                 from /home/pepper/dev/repos/nanos/src/kernel/kernel.h:2,
                 from /home/pepper/dev/repos/nanos/src/unix/unix_internal.h:3,
                 from /home/pepper/dev/repos/nanos/src/unix/unix.c:1:
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,11 +109,12 @@ Note: This is only recommended for those that wish to make code changes to Nanos
 
 Nanos doesn't need too many dependencies on Linux.
 
-To build you need to install nasm && qemu:
+To build you need to install nasm, qemu and go (If you don't want to install go, choose a different target, see below):
 
 ```
-sudo apt-get install qemu-system-x86 nasm
+sudo apt-get install qemu-system-x86 nasm golang-go
 ```
+
 
 For tests you'll also need the following installed:
 
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ brew install x86_64-elf-binutils
 brew tap nanovms/homebrew-qemu
 brew install nanovms/homebrew-qemu/qemu
 ```
+(note: ```go``` is only needed for certain [examples](#examples), but not for building the kernel)
 
 For ARM-based Macs (M1/M2):
 
@@ -109,12 +110,12 @@ Note: This is only recommended for those that wish to make code changes to Nanos
 
 Nanos doesn't need too many dependencies on Linux.
 
-To build you need to install nasm, qemu and go (If you don't want to install go, choose a different target, see below):
+To build you need to install nasm, qemu and go:
 
 ```
 sudo apt-get install qemu-system-x86 nasm golang-go
 ```
-
+(note: ```go``` is only needed for certain [examples](#examples), but not for building the kernel)
 
 For tests you'll also need the following installed:
 
@@ -143,7 +144,11 @@ sudo apt-get install libfuse-dev fuse
 make kernel
 ```
 
-#### To run an example program from the test/runtime folder:
+#### Examples
+
+
+To run an example program from the test/runtime folder:
+
 With hardware acceleration:
 
 ```
@@ -154,6 +159,15 @@ Without hardware acceleration:
 ```
 make run-noaccel
 ```
+
+Set ```TARGET``` to run a specific example:
+```
+make TARGET=<example> run
+```
+
+Check out [test/runtime](test/runtime/README.md) for a list of examples. Certain examples require ```go``` to be built. 
+
+More examples can be found in [docs/examples#examples](https://docs.ops.city/ops/examples#examples).
 
 ### Documentation
 
--- a/README.md
+++ b/README.md
@@ -122,6 +122,11 @@ For tests you'll also need the following installed:
 sudo apt-get install ent ruby
 ```
 
+Ops:
+```
+curl https://ops.city/get.sh -sSfL | sh
+```
+
 Rust:
 ```
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
--- a/test/unit/udp_test.c
+++ b/test/unit/udp_test.c
@@ -30,7 +30,7 @@ int main(int argc, char ** argv)
     u32 daddr;
     u16 dport;
     if (!parse_v4_address_and_port(target, &daddr, &dport))
-        test_perror("parse");
+        test_error("failed to parse address and port");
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -109,12 +109,13 @@ Note: This is only recommended for those that wish to make code changes to Nanos
 
 Nanos doesn't need too many dependencies on Linux.
 
-To build you need to install nasm && qemu:
+To build you need to install nasm, qemu and go (If you don't want to install go, choose a different target, see below):
```
francescolavra, line 112: This doesn't tell what target doesn't need go, I would be a bit more explicit, something along the lines of "installing go is not necessary if you only want to build the kernel"

test/unit/udp_test.c
```diff
@@ -29,7 +29,8 @@ int main(int argc, char ** argv)
     buffer target = vector_pop(unassoc);
     u32 daddr;
     u16 dport;
-    parse_v4_address_and_port(target, &daddr, &dport);
+    if (!parse_v4_address_and_port(target, &daddr, &dport))
+        test_perror("parse");
```
francescolavra, line 33: test_perror() prints the string representation of the `errno` variable value, which is not set when a parsing error occurs. We should use test_error() instead, and the error string should be more informative, e.g. "failed to parse address and port".

francescolavra (commented):
Just a couple of comments
Thanks

README.md
peppergrayxyz: Go is also mentioned as dependency for Mac, so this is consistent with what is already there. 

Go is not strictly needed for all targets, but it's needed for the default target: 

In ´´´Makefile´´´ the default target uses go:
```makefile
5: # runtime tests / ready-to-use targets
6: TARGET=		webg
```

If you follow the instructions in Readme
```
sudo apt-get install qemu-system-x86 nasm
make run
```
the build fails.

I think it's not a great experience likes this.

So should we mention that go is not needed for all targets (for both Mac and PC description)?

Or as an alternative, should we use another default target, that doesn't need go? And then mention go at the section where it is explained how to build specific targets? Maybe web or hw ?

test/unit/udp_test.c
peppergrayxyz: fixed!

README.md
francescolavra: I would keep the default target as-is, I just think "choose a different target" in the readme file is too vague ("target" might be interpreted as meaning a Makefile target, which is a different thing than the TARGET variable used when invoking make).
I'm OK if for simplicity you just add Go to the dependencies for building on Linux; but if you want to also add a note to say that Go is not needed when building/running a non-default target, this note should be more precise, e.g. mention the TARGET variable.

peppergrayxyz: https://github.com/nanovms/nanos/pull/2071/commits/8970938f3271613e485463483066fb536bdd88c3
- added a note about go to the mac and linux build instructions
- added TARGET variable explanation & link to overview in example section 
- added an overview of examples in the test/runtime dir

peppergrayxyz:
should be all resolved now, thanks for your help!

francescolavra (approved):
Thanks!

PR #2074, opened 2024-10-18
epoll: add support for polling on epoll file descriptors

### LINKED ISSUE #1192: epoll file descriptor is not poll/epoll/selectable
According to the epoll(7) man page, an epoll file descriptor should itself be poll/epoll/selectable. More precisely, if it has events waiting, then it should indicate as being readable.
The current implementation doesn't support polling an epoll file descriptor, because the relevant `events` closure is not implemented.

francescolavra:
In addition, epoll_ctl() now returns -EINVAL if called on a non-epoll file descriptor, or if the file descriptor being added, deleted, or modified is the same as the epoll file descriptor on which epoll_ctl() is called.

Closes #1192.

### DIFF
--- a/src/unix/notify.c
+++ b/src/unix/notify.c
@@ -77,7 +77,7 @@ u64 notify_get_eventmask_union(notify_set s)
     return u;
 }
 
-void notify_dispatch_with_arg(notify_set s, u64 events, void *arg)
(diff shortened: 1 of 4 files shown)


francescolavra:
I amended the code so that a polling syscall on a set of file descriptors that contains an epoll instance properly returns when the epoll instance becomes readable.

PR #2079, opened 2024-11-15
Logging: add user-adjustable log levels ("err", "warn", "info")

francescolavra:
The msg_{err,warn,info} macros have been repurposed so that in kernel code they expand to log_printf() function calls, which print the supplied message to the console if the log level (supplied as an additional argument) is lower than or equal to the configured kernel log level. The default log level is "err".
Example Ops configuration to set the log level to "info":
```
  "ManifestPassthrough": {
    "log_level": "info"
  }
```


eyberg:
https://github.com/nanovms/ops-documentation/commit/673612b66e925aad32bb8db0b6511d611ae85de2

eyberg (approved):

Issue #2080, opened 2024-11-20
Firecracker: Snapshot load causes a lot of firecracker virtio block error logs


jeromegn:
I don't know if these logs are important because my app appears to be working properly (though it isn't accessing the filesystem post-snapshot-load).

```
2024-11-20T12:07:50.095125693 [anonymous-instance:main] Failed to execute Out virtio block request: FileEngine(Sync(Transfer(IOError(Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }))))
2024-11-20T12:07:50.095131604 [anonymous-instance:main] Failed to execute Out virtio block request: FileEngine(Sync(Transfer(IOError(Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }))))
2024-11-20T12:07:50.095137595 [anonymous-instance:main] Failed to execute Out virtio block request: FileEngine(Sync(Transfer(IOError(Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }))))
```

I haven't seen anybody else report these errors to the firecracker repository so I am assuming it is something specific to `nanos`. I'm happy to make a firecracker issue as well if I am misguided.

To reproduce, I believe:
- Build (using `ops`) + start any `deno` app in a firecracker microvm
- Pause it
- Take a snapshot
- Stop the firecracker
- Start a new firecracker without any config
- Load the snapshot in the new firecracker

I see these logs before seeing any logs from my app:

```
2024-11-20T14:18:59.530639927 [anonymous-instance:main] Running Firecracker v1.10.1
2024-11-20T14:19:02.397687996 [anonymous-instance:fc_api] The API server received a Put request on "/snapshot/load" with body "{\n            \"snapshot_path\": \"./snapshot_file\",\n            \"mem_backend\": {\n                \"backend_path\": \"./mem_file\",\n                \"backend_type\": \"File\"\n            },\n            \"enable_diff_snapshots\": true,\n            \"resume_vm\": true\n    }".
2024-11-20T14:19:02.397716760 [anonymous-instance:main] [DevPreview] Virtual machine snapshots is in development preview.
2024-11-20T14:19:02.397852002 [anonymous-instance:main] Host CPU vendor ID: [65, 117, 116, 104, 101, 110, 116, 105, 99, 65, 77, 68]
[...]

PR #2081, opened 2024-11-22
virtio_balloon: add support for MMIO devices

francescolavra:
This allows the virtio memory balloon driver to work with AWS Firecracker, which instantiates MMIO devices.

### DIFF
--- a/src/virtio/virtio.c
+++ b/src/virtio/virtio.c
@@ -122,7 +122,7 @@ status virtio_register_config_change_handler(vtdev dev, thunk handler)
 {
     switch (dev->transport) {
     case VTIO_TRANSPORT_MMIO:
-        return timm("status", "not implemented");
+        return vtmmio_register_config_change_handler((vtmmio)dev, handler);
     case VTIO_TRANSPORT_PCI:
         return vtpci_register_config_change_handler((vtpci)dev, handler);
     default:
--- a/src/virtio/virtio_balloon.c
+++ b/src/virtio/virtio_balloon.c
@@ -416,5 +416,12 @@ void init_virtio_balloon(kernel_heaps kh)
 {
     virtio_balloon_debug("%s\n", func_ss);
     heap h = heap_locked(kh);
-    register_pci_driver(closure(h, vtpci_balloon_probe, h, heap_linear_backed(kh), heap_physical(kh)), 0);
+    backed_heap backed = heap_linear_backed(kh);
+    id_heap physical = heap_physical(kh);
+    pci_probe probe = closure(h, vtpci_balloon_probe, h, backed, physical);
+    if (probe == INVALID_ADDRESS) {
+        msg_err("%s: out of memory", func_ss);
+        return;
+    }
+    register_pci_driver(probe, 0);
 }
--- a/src/virtio/virtio_balloon.c
+++ b/src/virtio/virtio_balloon.c
@@ -414,6 +414,20 @@ closure_function(3, 1, boolean, vtpci_balloon_probe,
     return virtio_balloon_attach(bound(general), bound(backed), bound(physical), v);
 }
 
+closure_function(3, 1, void, vtmmio_balloon_probe,
+                 heap, general, backed_heap, backed, id_heap, physical,
+                 vtmmio dev)
+{
+    virtio_balloon_debug("MMIO probe\n", func_ss);
+    if ((vtmmio_get_u32(dev, VTMMIO_OFFSET_DEVID) != VIRTIO_ID_BALLOON) ||
+        (dev->memsize < VTMMIO_OFFSET_CONFIG + sizeof(struct virtio_balloon_config)))
+        return;
+    heap general = bound(general);
+    backed_heap backed = bound(backed);
+    if (attach_vtmmio(general, backed, dev, VIRTIO_BALLOON_DRV_FEATURES))
+        virtio_balloon_attach(general, backed, bound(physical), &dev->virtio_dev);
+}
+
 void init_virtio_balloon(kernel_heaps kh)
 {
     virtio_balloon_debug("%s\n", func_ss);
@@ -426,4 +440,5 @@ void init_virtio_balloon(kernel_heaps kh)
         return;
     }
     register_pci_driver(probe, 0);
+    vtmmio_probe_devs(stack_closure(vtmmio_balloon_probe, h, backed, physical));
 }
--- a/src/virtio/virtio_balloon.c
+++ b/src/virtio/virtio_balloon.c
@@ -26,7 +26,9 @@
 #define VIRTIO_BALLOON_PAGE_ORDER PAGELOG
 
 /* These are units that we allocate from the physical heap. */
-#define VIRTIO_BALLOON_ALLOC_ORDER 21
+/* The maximum number of pages supported by AWS Firecracker in a single inflate descriptor is 256,
+ * which corresponds to 1 MB of memory, and the balloon allocation unit is sized accordingly. */
+#define VIRTIO_BALLOON_ALLOC_ORDER 20
 #define VIRTIO_BALLOON_ALLOC_SIZE U64_FROM_BIT(VIRTIO_BALLOON_ALLOC_ORDER)
 #define VIRTIO_BALLOON_PAGES_PER_ALLOC U64_FROM_BIT(VIRTIO_BALLOON_ALLOC_ORDER - \
                                                     VIRTIO_BALLOON_PAGE_ORDER)
--- a/src/virtio/virtio_balloon.c
+++ b/src/virtio/virtio_balloon.c
@@ -86,6 +86,8 @@ struct virtio_balloon_config {
 #define VIRTIO_BALLOON_R_NUM_PAGES (offsetof(struct virtio_balloon_config *, num_pages))
 #define VIRTIO_BALLOON_R_ACTUAL    (offsetof(struct virtio_balloon_config *, actual))
 
+#define VIRTIO_BALLOON_DRV_FEATURES (VIRTIO_BALLOON_F_STATS_VQ | VIRTIO_BALLOON_F_MUST_TELL_HOST)
+
 static inline boolean balloon_must_tell_host(void)
 {
     return (virtio_balloon.dev->features & VIRTIO_BALLOON_F_MUST_TELL_HOST) != 0;
@@ -406,8 +408,7 @@ closure_function(3, 1, boolean, vtpci_balloon_probe,
 
     virtio_balloon_debug("   attaching\n", __func__);
     vtdev v = (vtdev)attach_vtpci(bound(general), bound(backed), d,
-                                  (VIRTIO_BALLOON_F_STATS_VQ |
-                                   VIRTIO_BALLOON_F_MUST_TELL_HOST));
+                                  VIRTIO_BALLOON_DRV_FEATURES);
     return virtio_balloon_attach(bound(general), bound(backed), bound(physical), v);
 }
 
--- a/src/virtio/virtio_mmio.c
+++ b/src/virtio/virtio_mmio.c
@@ -186,6 +186,18 @@ closure_func_basic(thunk, void, vtmmio_irq)
     }
 }
 
+static void vtmmio_irq_setup(vtmmio dev, range cpu_affinity)
+{
+    u64 irq_vector = allocate_mmio_interrupt();
+    assert(irq_vector != INVALID_PHYSICAL);
+    register_interrupt(irq_vector, init_closure_func(&dev->irq_handler, thunk, vtmmio_irq),
+                       ss("vtmmio"));
+    dev->irq_vector = irq_vector;
+#ifdef __x86_64__
+    ioapic_set_int(dev->irq, irq_vector, irq_get_target_cpu(cpu_affinity));
+#endif
+}
+
 status vtmmio_alloc_virtqueue(vtmmio dev, sstring name, int idx, range cpu_affinity,
                               struct virtqueue **result)
 {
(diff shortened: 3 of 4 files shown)


jeromegn:
Thank you for working on this!

How did you test it? I can't seem to get it to do much (though it doesn't error anymore).

I don't expect the behaviour to be the same as Linux's, but from previous experience with the ballooning device, it behaved like this:
- Initially set the size of the balloon to 0
- Run a the firecracker microvm
- Let it allocate a bunch of memory
- Inflate the balloon to reclaim memory on the host

If there's nothing preventing freeing the pages inside the microvm, then that's all there is to it. If it can't free all the requested pages, then the kernel starts logging balloon-related information. I believe it applies pressure for these pages to be freed and future allocations by the kernel are severely affected (read: impossible and everything slows down to a crawl).

Testing this branch locally, I'm unable to get any memory freed from the host's perspective. Even inflating the balloon to the total memory allowed in the firecracker doesn't appear to do much.

I'm running a Deno app and I suspect a lot of memory should be free-able. It's a little hard to tell because I'm unable to gather RSS from the host:

```javascript
let mem = Deno.memoryUsage();
console.log(`rss: ${mem.rss}, heap total: ${mem.heapTotal}, heap used: ${mem.heapUsed}, external: ${mem.external}`);
```

```
rss: 0, heap total: 13991936, heap used: 12160248, external: 3003513
```

I'm also using the `sysinfo` crate in my VM and it shows quite a lot of memory used (it does memory total - memory available):

```
=> system:
total memory: 533139456 bytes
used memory : 413736960 bytes
```

These numbers are with the balloon inflated to 512MB and appears to have no effect. Inflating it to other numbers also didn't appear to change anything.

Firecracker process from the host side:

```
❯ cat /proc/42517/status
Name:   firecracker-v1.
Umask:  0022
State:  S (sleeping)
Tgid:   42517
Ngid:   0
Pid:    42517
PPid:   42516
TracerPid:      0
Uid:    0       0       0       0
Gid:    0       0       0       0
FDSize: 524288
Groups: 0 
NStgid: 42517
NSpid:  42517
NSpgid: 42517
NSsid:  42516
Kthread:        0
VmPeak:   531548 kB
VmSize:   531528 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:    246528 kB
VmRSS:    246528 kB
RssAnon:          244736 kB
RssFile:            1792 kB
RssShmem:              0 kB
VmData:   529132 kB
VmStk:       152 kB
VmExe:      1836 kB
VmLib:         8 kB
VmPTE:       668 kB
VmSwap:        0 kB
```

Somehow the memory inside the VM is reported as more than outside which feels odd to me, but I suspect there's some weirdness regarding how memory is reported from inside the guest.

jeromegn:
Ah yes, sorry, I was just including the Deno output as a data point, but wasn't exactly relying on it. I found it interesting it couldn't report on RSS at all. Is that a limitation of nanos? Might just be an incompatible implementation in Deno.

I'll have to test again on Linux, but I don't recall the balloon showing as memory usage. My own memory could be wrong though! It would make sense that it'd show as usage.

> I don't know what Firecracker does with the memory in the balloon (e.g. whether it releases the memory to the host OS, or does something else), so I don't know if we can gather any relevant information from the Firecracker process memory stats.

The main purpose, in my opinion, of the balloon is that it would reclaim memory on the host. Firecracker does not reclaim allocated memory. If the guest bursts to 300 / 512MB and then goes back down to 100MB, then the host will show the process as using 300MB RSS. Inflating the balloon reclaims the now-unused memory from the host and lowers the RSS of the firecracker process.

I started my firecracker with `amount_mib: 0` and I saw this:

```
=> system:
total memory: 533139456 bytes
used memory : 239341568 bytes
```

Then I inflated the balloon to 300 MiB and saw:

```
=> system:
total memory: 533139456 bytes
used memory : 433418240 bytes
```

Then I deflated it back to 0 and saw:

```
=> system:
total memory: 533139456 bytes
used memory : 282423296 bytes
```

The host's RSS for firecracker did not move (or it did by maybe 1-3MB) through these various operations.

I short-circuited my program to allocate 200MB of random bytes of memory (allocating zeroes did nothing for some reason, maybe as an optimization in `nanos` or in Rust?) and I got the following:

```
=> system:
total memory: 533139456 bytes
used memory : 91029504 bytes

allocated 209715200 bytes

=> system:
total memory: 533139456 bytes
used memory : 300630016 bytes
```

Looking from the host, it did use ~238MB of RSS. I don't know why that figure would be smaller than what's reported inside the guest, but I can ignore that for now :)

After 30 seconds, I deallocated the `Vec<u8>` and saw this:

```
deallocated

=> system:
total memory: 533139456 bytes
used memory : 90910720 bytes
```

So far so good!

As expected, by my own mental model of Firecracker, the RSS for the process on the host remained stable at 238MB.

I then inflated the balloon to 400MiB and saw:

```
=> system:
total memory: 533139456 bytes
used memory : 456646656 bytes
```

Which would fit what you've said in your last comment.

The host has reclaimed the memory and is sitting at around 57MB. Again, that figure is a bit surprising to me because it's lower than what the guest reports.

---

Now, given all of that, it seems like memory ballooning does work as expected! My bad.

This is leading me to think that the base memory used by my own app is high and unreclaimable by the balloon device (and I assume: the kernel). It seems like running the same Deno app on then host uses roughly 128MB (RSS). So maybe the 90MB based I got in my program + 128MB sounds about right. I was hoping my own program would be closer to 128MB and have practically zero overhead.

My program is 103MB and `deno` compiled with the same flags is 123MB. In theory I _could_ be using less memory inside the guest, but in practice I am not and there's overhead unaccounted for (likely on my end).

---

Do you have any idea why RSS reported from the host for the firecracker process could be lower than the memory reported inside the host? In our own experience, that's never the case.

francescolavra:
> Ah yes, sorry, I was just including the Deno output as a data point, but wasn't exactly relying on it. I found it interesting it couldn't report on RSS at all. Is that a limitation of nanos? Might just be an incompatible implementation in Deno.

Deno retrieves the RSS from the /proc/self/statm file, which doesn't exist in Nanos, that's why the reported RSS value is 0.

> I short-circuited my program to allocate 200MB of random bytes of memory (allocating zeroes did nothing for some reason, maybe as an optimization in nanos or in Rust?)

Usually when a program allocates a large amount of memory, what happens under the hood is that an mmap() syscall is invoked, which allocates the requested amount of _virtual_ address space but does not map it directly to physical RAM, so RAM space is not allocated right away; the actual allocation (which shows up in the used memory) happens only when the program accesses the memory (in your case, when the memory is filled with random values). Allocating zeros does not require accessing the memory (because mmap()ed memory is always initialized with zeros), that's why it doesn't shows up in the used memory.

> Do you have any idea why RSS reported from the host for the firecracker process could be lower than the memory reported inside the host?

Nanos internally allocates guest physical memory for many different purposes, but not all of that memory may be used immediately, and only the memory that has actually been used (i.e. read from or written to) shows up in Firecracker's RSS. For example, Nanos has internal heaps that use different caches for different allocation sizes; if e.g. one 32-byte allocation is made inside the kernel, the relevant cache allocates a 2-MB region of physical memory, but only a fraction of that memory is going to be used immediately: in this scenario, Nanos reports the entire 2 MB as in-use memory (because it has been allocated in the guest physical memory and is therefore unavailable for other uses), while only the fraction that's actually used (which could be as low as 4 KB) shows up in Firecracker's RSS. Another example is the set of buffers pre-allocated by the network interface driver for reception of network packets: all the memory occupied by the buffers is considered in-use memory by Nanos, but only the buffers that have actually been used for receiving incoming network packets show up in the RSS.

I'm not sure I fully understood the issue you mentioned when comparing the RSS of your app when run directly on the host with the RSS of Firecracker when running your app under Nanos, but it seems to me you are considering the used memory reported by the guest (e.g. 90 MB) as separate from Firecracker's RSS, and you are adding the two figures. In reality, the two figures refer to the same memory (if we ignore the additional memory needed by the VMM itself to create and run the guest VM), and their differences can be explained with what I wrote above. (That's when the memory balloon is empty; if you inflate the balloon, its memory is accounted for in the guest used memory but not in Firecracker's RSS, of course.)
There would be an issue if the RSS when running your app under Nanos was substantially higher than the RSS of your app when run directly on the host, but as long as that doesn't happen I think we are seeing the expected behavior.

jeromegn:
Thanks for the explanations. This is insightful.

> I'm not sure I fully understood the issue you mentioned when comparing the RSS of your app when run directly on the host with the RSS of Firecracker when running your app under Nanos, but it seems to me you are considering the used memory reported by the guest (e.g. 90 MB) as separate from Firecracker's RSS, and you are adding the two figures.

I'm essentially trying to measure the overhead of `nanos`. Firecracker should using using a few MBs (6?) and my app uses ~120MB RSS, so I'm trying to account for the ~100MB extra I'm seeing from running my app with the `nanos` kernel.

I haven't yet had a chance to test this using Linux, but i suspect there's overhead there as well. I just don't expect it to be that much.

By using the balloon my hope was to reclaim as much as possible. It seems like there's more unreclaimable memory when using Deno vs. just allocating 200MB of random bytes on the heap in a Rust program. I'm saying this because of my test comparing the RSS of firecracker before and after allocating and before and after inflating the balloon.

I would expect to be able to reclaim a lot more memory on the host for an app that uses ~120MB of RSS.

From your explanation, it sounds like there's some form of allocation that's not being deallocated when the balloon expands.

The reason I am specifically testing for low memory footprint is that I want to fit as many of these as possible on a host. The difference between 128MB and 200MB+ of memory allocated for a firecracker process is big, almost halving the number that can fit on a host.

> Another example is the set of buffers pre-allocated by the network interface driver for reception of network packets: all the memory occupied by the buffers is considered in-use memory by Nanos, but only the buffers that have actually been used for receiving incoming network packets show up in the RSS.

My app should only have 1 TCP listener and I'm running Deno with `--cache-only` to prevent downloading packages at runtime.

I did notice that making a single request will increase the memory usage by a little bit.

is there a way to inspect what "kinds" of allocations (and how much) `nanos` is doing, at runtime?

My confusion comes from the fact that a test program's memory, allocating 200MB of random bytes, can be fully reclaimed when using the balloon device. But Deno's or V8's memory can't be reclaimed by the host in the same way. So I imagine it has something to do with the "kinds" of allocations it is doing.

On Linux, I suspect the balloon creates memory pressure and the kernel drops buffers and caches more aggressively. I'm wondering if `nanos` is not aggressive enough in this case. If I am applying memory pressure, then I expect to reclaim as much as possible, even at the cost of performance, if only temporarily.

> There would be an issue if the RSS when running your app under Nanos was substantially higher than the RSS of your app when run directly on the host, but as long as that doesn't happen I think we are seeing the expected behavior.

How do I measure that? The RSS stat is not available from inside.

As far as I can tell, there's roughly 2x overhead in memory by running my app in a Firecracker with the `nanos` kernel. The RSS of the `firecracker` process is what I care about because it's what the host sees and uses to determine used / available memory. If the `firecracker` RSS is artificially too high, then a host can come into an OOM situation (though I wouldn't let it go that far since I'd be tracking RSS).

francescolavra:
> is there a way to inspect what "kinds" of allocations (and how much) `nanos` is doing, at runtime?

You can get memory-related info from the guest via the virtio-balloon statistics (see https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md#virtio-balloon-statistics).
To get more detailed info on kernel internals, we have the in-kernel management interface that implements a telnet server that can be queried to retrieve various data, including memory heap statistics; but this interface is not enabled in production kernels, needs to be enabled explicitly by rebuilding the kernel with `MANAGEMENT=telnet`, and in order to make sense of the info you get from it you would need advanced knowledge of kernel internals.

> How do I measure that? The RSS stat is not available from inside.

I was referring to the RSS of the firecracker process, not the RSS of the deno process inside the guest.

So it seems you are seeing ~100MB extra RSS in firecracker compared to running your deno app directly on the host; and it seems you are unable to reclaim that memory via the virtio-ballon device. This sounds like something that could be improved.
If I run the simple app in main.ts at https://github.com/nanovms/nanos/issues/2080#issue-2677006591, I see roughly 25 MB of difference between the two RSS values (52 MB when running the app in the host, vs 77 MB when running it with Firecracker), and that seems pretty reasonable, given that we have to account for the memory needed by Nanos to operate the guest VM, as well as the memory needed by the Firecracker process itself. Do you have sample code that shows the ~100 MB unreclaimable overhead when run with Nanos and Firecracker? If we can replicate this behavior, we can take a look and see what can be done about it.

jeromegn:
> If I run the simple app in main.ts at [#2080 (comment)](https://github.com/nanovms/nanos/issues/2080#issue-2677006591), I see roughly 25 MB of difference between the two RSS values (52 MB when running the app in the host, vs 77 MB when running it with Firecracker), and that seems pretty reasonable, given that we have to account for the memory needed by Nanos to operate the guest VM, as well as the memory needed by the Firecracker process itself. Do you have sample code that shows the ~100 MB unreclaimable overhead when run with Nanos and Firecracker? If we can replicate this behavior, we can take a look and see what can be done about it.

That does seem reasonable, yes.

I'll try and reproduce the memory bloat with my simple app.

The one that does appear to be problematic is a [`fresh`](https://fresh.deno.dev/) boilerplate app. Basically:

```
deno run -A -r https://fresh.deno.dev
```

Then, if you're using `deno compile` (btw, I am using Deno 2.x for this), then you can point it at the `main.ts` file as an entrypoint.

francescolavra:
With #2085, Nanos is more effective in allowing the hypervisor to release host memory. @jeromegn now if I run your executable with Firecracker and Nanos and I inflate the balloon, Firecracker's RSS first decreases by a couple of MB (which correspond to non-file-backed memory that has first been allocated and then deallocated by the executable and/or the kernel); then, if I inflate more, the RSS decreases by another ~40 MB, which is the result of caches being dropped (you can see the difference in the "disk_caches" value returned when you query the balloon statistics).
To test the kernel with these changes, you can use `--nanos-version 9c5ce16 ` Ops command line option.
There is still room for improvement: Nanos currently only drops the file cache for files that are not mapped in the process address space; if we implement a more aggressive cache eviction algorithm that can un-cache also mapped files, there would be more memory available for the balloon and thus more RSS savings for the host.

francescolavra:
#2097 enhances the cache eviction functionality to be able to un-cache both mapped and non-mapped files, so when the host requests the virtio balloon to be inflated, the guest can release virtually all the memory used by the page cache, which minimizes the RSS of the firecracker process.
Note: there is a 16-MB threshold in the balloon driver that prevents it from inflating the balloon further when the amount of available memory goes below the threshold; this mechanism is in place in order to prevent the driver from exhausting all guest memory (which could potentially cause the user process to be terminated when running out of memory).

PR #2092, opened 2025-02-24
fix: file read - offset and length checks

### LINKED ISSUE #2091: bug: syscall/file_read (also file_readv probably) wrong return code when offset > file length
- `ops run pread64-offset`
```
running local instance
....
FileSize: 0 - offset 0 - reading 100 bytes - read 0 bytes: 
FileSize: 0 - offset 24 - reading 16 bytes - error: too many open files
FileSize: 24 - offset 24 - reading 16 bytes - read 0 bytes: 
FileSize: 24 - offset 25 - reading 16 bytes - error: operation not permitted
```
- code to reproduce (also anything involving **SQLite** should be affected, when creating a new **empty** db file only)
```go
package main

import (
	"fmt"
	"os"
	"syscall"
)

func main() {
	fileName := "pread64_offset.txt"

	// Step 1: Check if the file exists, and create it if it does not
	file, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE, 0666)
	if err != nil {
		fmt.Printf("Error opening or creating file: %v\n", err)
		return
	}
	defer file.Close()

	fileInfo, _ := file.Stat()

	// Step 2: Try to read 100 bytes starting from offset 0
	// Nanos - trace - OK
	//    2 file_read: f 0xffffc000001917b0, dest 0x000000c00001a0e0, offset 0 (specified), length 100, file length 0
	//  [2] pread64(3, 0x000000c00001a0e0, 100, 0) = 0
	offset := int64(0)
	size := 100
	buf := make([]byte, size)
	n, err := syscall.Pread(int(file.Fd()), buf, offset)
	if err != nil {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - error: %v\n", fileInfo.Size(), offset, size, err)
	} else {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - read %d bytes: %s\n", fileInfo.Size(), offset, size, n, buf[:n])
	}

	// Step 3: Try to read 16 bytes starting from offset 24
	// Nanos - trace - WRONG instead of -EMFILE = -24 = -(offset - file_length), response should be 0
	//     2 file_read: f 0xffffc000001917b0, dest 0x000000c000010090, offset 24 (specified), length 16, file length 0
	//   [2] pread64(3, 0x000000c000010090, 16, 24) = -EMFILE
	offset = int64(24)
	size = 16
	buf = make([]byte, size)
	n, err = syscall.Pread(int(file.Fd()), buf, offset)
	if err != nil {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - error: %v\n", fileInfo.Size(), offset, size, err)
	} else {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - read %d bytes: %s\n", fileInfo.Size(), offset, size, n, buf[:n])
	}

	// Step 4: Write 24 bytes and try to read 16 bytes starting from offset 24
	data := []byte("123456789012345678901234")
	_, err = file.Write(data)
	if err != nil {
		fmt.Printf("FileSize: %d - Error writing to file: %v\n", fileInfo.Size(), err)
		return
	}

	fileInfo, _ = file.Stat()

	// Attempt to read 16 bytes starting from offset 24 again
	// Nanos - trace - OK
	//     4 file_read: f 0xffffc000001917b0, dest 0x000000c000010090, offset 24 (specified), length 16, file length 24
	//   [4] pread64(3, 0x000000c000010090, 16, 24) = 0
	n, err = syscall.Pread(int(file.Fd()), buf, offset)
	if err != nil {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - error: %v\n", fileInfo.Size(), offset, size, err)
	} else {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - read %d bytes: %s\n", fileInfo.Size(), offset, size, n, buf[:n])
	}

	// Step 5: Attempt to read 16 bytes starting from offset fileInfo.Size() + 1
	// Nanos - trace - WRONG instead of -EPERM = -1 = -(offset - file_length), response should be 0
	//     2 file_read: f 0xffffc000001917b0, dest 0x000000c000010090, offset 25 (specified), length 16, file length 24
	//   [2] pread64(3, 0x000000c000010090, 16, 25) = -EPERM
	offset = fileInfo.Size() + 1
	n, err = syscall.Pread(int(file.Fd()), buf, offset)
	if err != nil {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - error: %v\n", fileInfo.Size(), offset, size, err)
	} else {
		fmt.Printf("FileSize: %d - offset %d - reading %d bytes - read %d bytes: %s\n", fileInfo.Size(), offset, size, n, buf[:n])
	}
}
```

rinor:
- check offset validity before reading
- adjust read length based on file size limits

resolve #2091


francescolavra:
There is a bug that affects both reads and writes when the offset is greater than the file length; at least for the read part, it is caused by https://github.com/nanovms/nanos/blob/26ee8dc117974867e1e61f9c448d9817eef9ee50/src/unix/syscall.c#L455, and I think it should be fixed by changing the file_io_complete() function so that it avoids adjusting the return value when 0 bytes have been read, e.g. by changing the preceding if () condition to something like `if ((len > 0) && (r.start + len > file_len))`. I'm not sure that would fix the write issue too, though.

rinor:
well writes are a bit weird and would argue that nanos is more posix compliant than linux (bug) when it comes to `pwrite64`.

From https://www.man7.org/linux/man-pages/man2/pwrite.2.html :
**Bugs**
**POSIX** requires that opening a file with the **O_APPEND** flag `should have no effect on the location at which pwrite() writes data`. _However_, on Linux, **if a file is opened with O_APPEND**, pwrite() `appends data to the end of the file, regardless of the value of offset`.

that means that right now this works the same on both nanos/linux alowing and you can seek and fill the "holes" with `0x00`
```go
file, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE, 0666)
offset := fileInfo.Size() + 1 // put offset outside file length
n, err = syscall.Pwrite(int(file.Fd()), []byte("12345"), offset)
```

but this one behaves different, on nanos behaves like the POSIX expects, but on Linux offset is not taken in consideration and data is just appended to the end of file without "holes"
```go
file, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
offset := fileInfo.Size() + 1 // put offset outside file length
n, err = syscall.Pwrite(int(file.Fd()), []byte("12345"), offset)
```

Is this what you are referring to about writes?
>There is a bug that affects both reads and writes when the offset is greater than the file length;

src/unix/syscall.c
```diff
@@ -451,8 +451,12 @@ static void file_io_complete(file f, range r, boolean is_file_offset, sg_list sg
         u64 len = range_span(r);
         len -= sg_total_len(sg);
         u64 file_len = fsfile_get_length(f->fsf);
-        if (r.start + len > file_len)   /* can happen with direct I/O */
+        if ((len < 0) || (r.start >= file_len))
```
francescolavra, line 454: `len` can never be less than 0, so this should be simply `if (r.start >= file_len)`

src/unix/syscall.c
```diff
@@ -451,8 +451,12 @@ static void file_io_complete(file f, range r, boolean is_file_offset, sg_list sg
...
+            len = 0;
+        else if (r.start + len > file_len)   /* can happen with direct I/O */
             len = file_len - r.start;
+        if (len < 0)
+            len = 0;
```
francescolavra, line 459: Same here, `len < 0` cannot happen, so the above two lines can be removed.

francescolavra (commented):
Sorry, somehow I was under the impression that pwrite() was suffering from the same bug as pread(), but I checked better now and that's not the case, so pwrite() is fine as it is.
I have just 2 trivial comments on the changes for fixing pread().

PR #2111, opened 2025-10-27
Add klib to populate environment from IMDS userdata

tpjg:
A new klib 'userdata_env' is added. This will attempt to read the 'userdata' from various cloud providers and use it to set environment variables.

This can be used to separate preparing images and populating the environment variables.

Cloud provider typically have IMDS (Instance Metadata service) that is ONLY reachable from within the instance on 169.254.169.254.

While providers have warnings like the following (from Azure):
"
Warning
User data will not be encrypted, and any process on the VM can query this data. You should not store confidential information in user data.
"

Since nanos does not have multiple processes or users this does not apply for nanovms.

See also https://github.com/nanovms/ops/pull/1725 to add support to 'ops' for supplying UserData.

### DIFF
--- a/klib/userdata_env.c
+++ b/klib/userdata_env.c
@@ -24,6 +24,7 @@ enum cloud_provider {
     PROVIDER_AZURE,
     PROVIDER_DIGITALOCEAN,
     PROVIDER_OPENSTACK,
+    PROVIDER_HETZNER,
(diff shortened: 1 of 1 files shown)


tpjg:
Or consider https://github.com/nanovms/nanos/pull/2112 instead.

PR #2123, opened 2026-01-04
net: add DHCP Option 121 (Classless Static Routes) support

tpjg:
Implement RFC 3442 classless static routes received via DHCP to support receiving multiple routes (not just a default gateway). This is especially useful for servers deployed with multiple interfaces, e.g. one 'public' and one 'private' network interface.

Amongst other things this can be used to receive the route to reach the http://169.254.169.254 endpoint.

This has been tested on OpenStack / OVHcloud. Note that this also requires a change in the vendored LWIP (the makefile now pulls from my fork).

Not sure if this is interesting to get into upstream and it definitely needs (a lot of) review, sharing nonetheless.

### DIFF
--- a/src/net/lwipopts.h
+++ b/src/net/lwipopts.h
@@ -200,11 +200,3 @@ static inline int net_ip_input_hook(struct pbuf *pbuf, struct netif *input_netif
 }
 
 #define LWIP_DHCP_CLASSLESS_STATIC_ROUTES  1
-
-/*
- * Static route lookup hook for DHCP Option 121 (Classless Static Routes).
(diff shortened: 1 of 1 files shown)


tpjg:
Hi @francescolavra ; it's been a while as I was focussing on other things. I have now rewritten this and I agree this is much cleaner. Almost everything is done within the forked lwip, almost nothing in the nanos repo,  only the makefiles and a one line to enable the option 121: `#define LWIP_DHCP_CLASSLESS_STATIC_ROUTES 1`.

PR #2135, opened 2026-02-12
Fix readonly_rootfs blocking writes to device files

### LINKED ISSUE #2134: readonly_rootfs blocks open(O_WRONLY|O_RDWR) on device files
When readonly_rootfs is enabled, open() with O_WRONLY or O_RDWR on device files like /dev/null returns EROFS. Such kernel interfaces should be exempt from the read-only filesystem checks because writes are to the device driver, not the readonly filesystem.

To reproduce, manifest with readonly_rootfs:true then build and run with QEMU:

```c
#include <stdio.h>
#include <fcntl.h>
#include <errno.h>
#include <string.h>
#include <unistd.h>

int main() {
    int fd;
    fd = open("/dev/null", O_WRONLY);
    if (fd < 0) printf("FAIL: open(/dev/null, O_WRONLY) = %s (errno %d)\n", strerror(errno), errno);
    else { printf("OK: open(/dev/null, O_WRONLY) = fd %d\n", fd); close(fd); }

    fd = open("/dev/null", O_RDWR);
    if (fd < 0) printf("FAIL: open(/dev/null, O_RDWR) = %s (errno %d)\n", strerror(errno), errno);
    else { printf("OK: open(/dev/null, O_RDWR) = fd %d\n", fd); close(fd); }

    fd = open("/dev/null", O_RDONLY);
    if (fd < 0) printf("FAIL: open(/dev/null, O_RDONLY) = %s (errno %d)\n", strerror(errno), errno);
    else { printf("OK: open(/dev/null, O_RDONLY) = fd %d\n", fd); close(fd); }

    return 0;
}
```

This incorrectly prints:

```sh
FAIL: open(/dev/null, O_WRONLY) = Read-only file system (errno 30)
FAIL: open(/dev/null, O_RDWR) = Read-only file system (errno 30)
OK: open(/dev/null, O_RDONLY) = fd 3
```

niklasfemerstrand:
Fixes https://github.com/nanovms/nanos/issues/2134

`open()` with `O_WRONLY` or `O_RDWR` on device files like `/dev/null` returns `EROFS` when `readonly_rootfs` is enabled. Device writes go to the driver, not the filesystem, so special files should be exempt from the read-only check.

### DIFF
--- a/src/unix/syscall.c
+++ b/src/unix/syscall.c
@@ -856,7 +856,7 @@ int file_open(filesystem fs, tuple n, tuple parent, int flags, fsfile fsf)
         break;
     case O_WRONLY:
     case O_RDWR:
-        if (filesystem_is_readonly(fs)) {
+        if (filesystem_is_readonly(fs) && !get(n, sym(special))) {
             return -EROFS;
         }
         if (!(file_meta_perms(p, n) & ACCESS_PERM_WRITE)) {


francescolavra (changes requested):
This is a sensible fix, but I would like to take this opportunity to do some cleanup.
The `get()` call is somewhat expensive, as is does a table lookup, and we already have a couple of similarly expensive lookups in the same function in the form of `is_dir()` calls; then we have a `file_type_from_tuple()` call which can potentially re-do the same lookups again. It would be better to move the `file_type_from_tuple()` call at the top of the function, so we have the `type` variable which we can compare against known values in the rest of the function: so the `is_dir()` calls become `type == FDESC_TYPE_DIRECTORY`, and your fix would become `if (type == FDESC_TYPE_REGULAR && filesystem_is_readonly(fs))` (regular files are the only file types where a write modifies the filesystem, so this is more accurate than testing against special files).

Thanks!

francescolavra:
Superseded by https://github.com/nanovms/nanos/pull/2164

Issue #2136, opened 2026-02-12
Broken Xen PVH boot


niklasfemerstrand:
Nanos already provides `pvh_start()` and the `__xen_guest` ELF note but booting Nanos as a PVH guest triple faults during early init because:

1. `pvh_start()` is called with RSP=0.
2. `hvm_start_info` memory map is ignored.
3. ACPI init is called unconditionally but PVH has no ACPI tables.
4. PVH has no local APIC but `init_apic()` tries to read MADT and APIC MMIO.
5. Running a Xen PVH guest exposes a latent bug in `init_physical_heap()` where `break` after finding bootstrap regions skips low-memory trimming of remaining regions.

## `init_physical_heap` bug detail

Nanos currently performs trimming and bootstrap search in one loop:

```c
for_regions(e) {
    if (e->type == REGION_PHYSICAL) {
        if (e->base < MB) { /* trim */ }
        if (length >= BOOTSTRAP_SIZE) {
            /* found bootstrap */
            break;  // remaining regions never get trimmed
        }
    }
}
```

Xen allocates RAM starting at address 0x0. After `setup_initmap` splits around the kernel load address, there are two regions:

```
Region 0: [0x00000000, 0x00200000)  2 MB
Region 1: [0x00600000, 0x10000000)  ~250 MB
```

`for_regions` visits Region 1 first (large enough for bootstrap) and `break`. Region 0 never gets trimmed, instead its sub-1 MB pages enter the heap, get allocated, and fault because identity mapping below 1 MB are forbidden.

## Testing

All existing QEMU tests pass (16 unit + 31 runtime via `make test-noaccel`).

Boot any PVH guest on Xen 4.20:

```
name    = "nanos-pvh"
type    = "pvh"
memory  = 256
vcpus   = 1
kernel  = "/kernel.img"
disk    = ["file:/app.img,xvda,w"]
vif     = ["bridge=xenbr0"]
```

PR #2137, opened 2026-02-12
Fix Xen PVH boot and init_physical_heap low-memory trimming bug

niklasfemerstrand:
https://github.com/nanovms/nanos/issues/2136

### DIFF
--- a/platform/pc/service.c
+++ b/platform/pc/service.c
@@ -78,6 +78,7 @@ range kern_get_elf(void)
 
 BSS_RO_AFTER_INIT static boolean have_rdseed;
 BSS_RO_AFTER_INIT static boolean have_rdrand;
+BSS_RO_AFTER_INIT boolean pvh_boot;
 
 static boolean hw_seed(u64 * seed, boolean rdseed)
 {
@@ -205,8 +206,9 @@ void start_secondary_cores(kernel_heaps kh)
 
 void count_cpus_present(void)
 {
-    /* Read ACPI tables for MADT access */
-    init_acpi_tables(get_kernel_heaps());
+    /* Read ACPI tables for MADT access (not available on PVH) */
+    if (!pvh_boot)
+        init_acpi_tables(get_kernel_heaps());
 
 #ifdef SMP_ENABLE
     if (acpi_walk_madt(stack_closure_func(madt_handler, count_processors_handler))) {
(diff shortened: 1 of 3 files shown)


platform/pc/service.c
```diff
@@ -205,8 +206,9 @@ void start_secondary_cores(kernel_heaps kh)
 
 void count_cpus_present(void)
 {
-    /* Read ACPI tables for MADT access */
-    init_acpi_tables(get_kernel_heaps());
+    /* Read ACPI tables for MADT access (not available on PVH) */
+    if (!pvh_boot)
+        init_acpi_tables(get_kernel_heaps());
```
francescolavra, line 211: It should be possible to call both init_acpi_tables() and init_acpi() on a guest without ACPI, in fact Nanos runs just fine on ARM guests without ACPI. I also tested a modified kernel where the AcpiOsGetRootPointer() function in src/x86_64/acpi.c always returns 0 (to simulate an x86 guest without ACPI) and it boots just fine, both with and without KVM.
If the kernel crashes on a Xen PVH guest when trying to initialize ACPI, I would rather debug and fix that, and avoid the pvh_boot global variable.

src/x86_64/init.s
```diff
@@ -107,6 +107,8 @@ extern pvh_start
 
 bits 32
 pvh_start32:
+        ; ESP is undefined at PVH entry, we need a temporary stack before calls
+        mov esp, 0xa000
```
francescolavra, line 111: Good catch, but since there already is a stack pointer initialization a few instructions below this, I would remove that in favor of this one (and update/move the relevant comment).

src/x86_64/interrupt.c
```diff
@@ -383,8 +385,9 @@ void init_interrupts(kernel_heaps kh)
     assert(v != INVALID_PHYSICAL);
     spurious_int_vector = v;
 
-    /* APIC initialization */
-    init_apic(kh);
+    /* APIC initialization, skip for PVH (Xen uses event channels) */
+    if (!pvh_boot)
+        init_apic(kh);
```
francescolavra, line 390: I looked at the Xen source, and as far as I can tell, Xen PVH guests do have a local APIC (see https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/arch/x86/domain.c;h=9ba2774762ccfc2b79d6892c55eaa06e5e66ac29;hb=HEAD#l784, where the LAPIC emulation flag is marked as required for PVH domU). So if there are issues getting it to work, I prefer to sort them out instead of disabling the APIC code.

francescolavra (changes requested):
Thanks for your PR! A few comments inline

PR #2139, opened 2026-02-22
netsock_poll(): poll all network interfaces

francescolavra:
When the destination address of a network packet coincides with the address of a network interface, the packet is queued in the loopback queue of that interface. The netsock_poll() function is only polling the loopback interface, therefore it misses packets whose destination is the address of a non-loopback interface.
This change set ensures that the netsock_poll() function polls all interfaces instead of just the loopback interface.
This first commit is an unrelated fix for a build error that occurs with binutils version 2.46.

### DIFF
--- a/platform/pc/boot/Makefile
+++ b/platform/pc/boot/Makefile
@@ -149,7 +149,7 @@ msg_uefi_ld=    LD	$@
 cmd_uefi_ld=    $(LD) $(LDLAGS-uefi) $(OBJS-uefi) -o $@
 
 msg_uefi_objcopy=    OBJCOPY	$(OBJDIR)/bootx64.efi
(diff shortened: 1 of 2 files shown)


francescolavra:
This fixes also the following warning when running Payara 7.2025.2 on Azul JRE 25.0.1:

> [#|2026-02-22T20:04:27.687+0000|WARNING|Payara 7.2025.2|javax.enterprise.system.jmx|_ThreadID=173;_ThreadName=Thread-18;_TimeMillis=1771790667687;_LevelValue=900;_MessageID=NCLS-JMX-00007;|
  Cannot start JMX connector JmxConnector config: { name = system, Protocol = rmi_jrmp, Address = 0.0.0.0, Port = 8686, AcceptAll = false, AuthRealmName = admin-realm, SecurityEnabled = false} due to exception java.io.IOException: Cannot bind to URL [rmi://0.0.0.0:8686/jmxrmi]: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: Exception creating connection to: 0.0.0.0; nested exception is: 
	java.net.SocketException: Software caused connection abort]|#]

PR #2171, opened 2026-06-21
aarch64: Fix context refcount leak on SVC dispatch

robertream:
`synchronous_handler` acquires a refcount but only releases it on the fault path, leaking one ref per syscall. Since syscall_handler takes its own refcount, we can safely release the initial refcount before syscall dispatch.

### DIFF
--- a/src/aarch64/interrupt.c
+++ b/src/aarch64/interrupt.c
@@ -217,6 +217,10 @@ void synchronous_handler(void)
     int ec = field_from_u64(esr, ESR_EC);
     if (ec == ESR_EC_SVC_AARCH64 && (esr & ESR_IL) &&
         field_from_u64(esr, ESR_ISS_IMM16) == 0) {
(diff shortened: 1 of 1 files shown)


francescolavra (changes requested):
Nice find!
However, it would be better to not acquire the refcount at all (in  `synchronous_handler()`) when taking the syscall path, instead of acquiring it and then releasing it. Also, setting `f[FRAME_FULL] = true` is redundant in the syscall path, since `syscall_handler()` does that.
So I think the correct fix would be to move the following 2 lines in `synchronous_handler()` so they are just before `fault_handler fh = ctx->fault_handler`.
```
    f[FRAME_FULL] = true;
    context_reserve_refcount(ctx);
```

By the way, the same bug is also present in the riscv64 code, in `trap_exception()`, and the same fix should be applied there too.

francescolavra:
Superseded by https://github.com/nanovms/nanos/pull/2178

PR #2179, opened 2026-08-20
Enable JVM ZGC Uncommit - kernel: do not wait for the vmap lock deaf to the flush that needs an answer

AngeloRubens:
Why this is here. This is one of two changes that together let a JVM heap
shrink on nanos. ZGC keeps its heap in a memfd, maps it, and has a thread that
punches holes in it to return memory to the system; the companion pull request
makes fallocate(FALLOC_FL_PUNCH_HOLE) on a memory filesystem give the pages
back, which is what makes the collector's uncommit do anything at all. This
patch is what makes that survivable: with the punch returning memory and the
collector faulting the same file back in, the deadlock below fires within
seconds, every time.

The defect. Servicing a TLB shootdown in nanos is not an acknowledgement, it
is a join and a wait: page_invalidate_sync(..., rendezvous) spins until
joined == total_processors because the initiator runs the completion with every
other processor stopped. A processor that is spinning for a lock with interrupts
disabled never takes the flush interrupt, never joins, and the initiator waits
for it forever.

That is what spin_lock_irq(&p->vmap_lock) does in the fault path. Caught with
the stub on four processors, on the binary nanos ships today:

    CPU1  page_invalidate_sync(rendezvous=1)   flush.c:196, waiting for every
          processor to join
          <- pagecache_node_unmap_pages_sync <- vmap_unmap <- munmap,
          holding the process's vmap lock
    CPU2, CPU3  parked in flush_gen_rlocked, flush.c:65 -- they joined
    CPU0  unix_fault_handler, unix.c:320, spinning on that same vmap lock with
          interrupts off -- it can never join

So the wait for the vmap lock is made to service a pending flush while it spins,
which is what the flush machinery already expects of anyone holding a lock it
has to reach. Nothing else changes: the lock is acquired and released exactly
where it was.

This is not a defect of the new work. The unmapping side is plain munmap
and madvise of a shared file mapping, and the faulting side is any thread
touching the same file. It predates all of it and reproduces on an unmodified
tree.

Evidence. The test added here, vmap_flush_race, stops the machine in its
first round without the change and runs its sixty-four rounds with it, at
VCPUS=2, which is what the CI runs. Under the workload it was found in -- a
JVM with ZGC and an elastic heap, four processors, hardware virtualisation --
the guest wedged in between eight and thirty-three seconds on four attempts out
of four without this change, and ran a full thirty minutes with it: 1802
seconds, 738 uncommits, 21190 MB returned, 1.96 billion pages written and read
back through mappings of the file, nothing read back wrong and no fault or
assertion in the kernel. A second run of the same shape returned 50404 MB.

### DIFF
--- a/src/kernel/kernel.h
+++ b/src/kernel/kernel.h
@@ -179,6 +179,29 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
     irq_restore(flags);
 }
 
+/* For a lock whose holder can end up parked in a TLB flush rendezvous. Waiting for one of those
+ * with interrupts off is a deadlock: the flush IPI never arrives, so this processor never joins,
+ * so the rendezvous never ends, so its holder never lets go. The wait does the flush work itself. */
+static inline u64 spin_lock_irq_flush(spinlock l)
+{
+    u64 flags = irq_disable_save();
+    volatile u64 *p = (volatile u64 *)&l->w;
+#ifdef LOCK_STATS
+    u64 spins = 0;
+#endif
+    while (*p || !compare_and_swap_64(&l->w, 0, 1)) {
+#ifdef LOCK_STATS
+        spins++;
+#endif
+        page_invalidate_flush();
+        kern_pause();
+    }
+#ifdef LOCK_STATS
+    LOCKSTATS_RECORD_LOCK(l->s, true, spins, 0);
+#endif
+    return flags;
+}
+
 static inline u64 spin_wlock_irq(rw_spinlock l)
 {
     u64 flags = irq_disable_save();
--- a/src/unix/unix_internal.h
+++ b/src/unix/unix_internal.h
@@ -828,7 +828,7 @@ boolean unix_timers_init(unix_heaps uh);
 
 #define sysreturn_from_pointer(__x) ((s64)u64_from_pointer(__x));
 
-#define vmap_lock(p) u64 _savedflags = spin_lock_irq(&(p)->vmap_lock)
+#define vmap_lock(p) u64 _savedflags = spin_lock_irq_flush(&(p)->vmap_lock)
 #define vmap_unlock(p) spin_unlock_irq(&(p)->vmap_lock, _savedflags)
 
 extern sysreturn syscall_ignore();
--- a/test/runtime/Makefile
+++ b/test/runtime/Makefile
@@ -140,6 +140,7 @@ LDFLAGS-ktest=		-static
 
 SRCS-mmap= \
 	$(CURDIR)/mmap.c \
+	$(CURDIR)/vmap_flush.c \
 	$(SRCDIR)/unix_process/unix_process_runtime.c \
 	$(RUNTIME)
 LDFLAGS-mmap=		-static
--- a/test/runtime/vmap_flush.c
+++ b/test/runtime/vmap_flush.c
@@ -1,0 +1,128 @@
+/* More than one thread giving pages back at the same time, while others fault
+   on them and the file they came from is synced.
+
+   tlbshootdown covers the neighbouring case: one thread unmaps while the others
+   read, and each of them takes the fault it should. What it never has is two
+   threads returning pages at once, and that is what closes a cycle -- madvise
+   takes the process's vmap lock with interrupts disabled and then unmaps, which
+   opens a shootdown every processor may have to join synchronously; a second
+   thread waiting for that same lock with interrupts disabled never services the
+   shootdown, so it never joins, so the first never finishes and never lets the
+   lock go.
+
+   Two processors are enough. */
+
+#define _GNU_SOURCE
+#include <pthread.h>
+#include <sched.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/sysinfo.h>
+#include <unistd.h>
+
+#include "../test_utils.h"
+
+#define MAX_CPUS    16
+#define PAGESIZE    4096
+#define MAP_PAGES   2048
+#define MAP_BYTES   (MAP_PAGES * PAGESIZE)
+#define ROUNDS      64
+
+static pthread_t threads[MAX_CPUS];
+static volatile uint8_t *m;
+static volatile int done;
+static int np;
+
+static int kidcnt;
+static pthread_cond_t kid_cv;
+static pthread_cond_t sync_cv;
+static pthread_mutex_t sync_mut;
+
+static void wait_for_sync(void)
+{
+    pthread_mutex_lock(&sync_mut);
+    kidcnt++;
+    pthread_cond_signal(&kid_cv);
+    pthread_cond_wait(&sync_cv, &sync_mut);
+    pthread_mutex_unlock(&sync_mut);
+}
+
+static void wait_for_children(void)
+{
+    pthread_mutex_lock(&sync_mut);
+    while (kidcnt < np)
+        pthread_cond_wait(&kid_cv, &sync_mut);
+    kidcnt = 0;
+    pthread_mutex_unlock(&sync_mut);
+}
+
+static void wake_children(void)
+{
+    pthread_mutex_lock(&sync_mut);
+    pthread_cond_broadcast(&sync_cv);
+    pthread_mutex_unlock(&sync_mut);
+}
+
+/* Bring a chunk of the mapping in, then give it back: it takes both, because a
+   range that is already gone is unmapped without a shootdown. */
+static void *cpu_thread(void *v)
+{
+    int id = (int)((uintptr_t)v);
+    size_t share = MAP_BYTES / np;
+    size_t base = share * id;
+
+    while (!done) {
+        wait_for_sync();
+        if (done)
+            break;
+        for (size_t off = base; off < base + share; off += PAGESIZE)
+            m[off] = (uint8_t)(id + 1);
+        test_assert(madvise((void *)(m + base), share, MADV_DONTNEED) == 0);
+    }
+    return NULL;
+}
+
+void test_vmap_flush(int argc, char *argv[])
+{
+    np = get_nprocs();
+    if (np > MAX_CPUS)
+        np = MAX_CPUS;
+    if (np < 2)
+        /* With one processor there is nobody to wait for the lock, and the
+           cycle cannot exist. */
+        return;
+
+    pthread_cond_init(&kid_cv, NULL);
+    pthread_cond_init(&sync_cv, NULL);
+    pthread_mutex_init(&sync_mut, NULL);
+
+    int fd = memfd_create("vmap_flush", 0);
+    test_assert(fd >= 0);
+    test_assert(ftruncate(fd, MAP_BYTES) == 0);
+    m = mmap(NULL, MAP_BYTES, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+    test_assert(m != MAP_FAILED);
+
+    for (long i = 0; i < np; i++)
+        test_assert(pthread_create(&threads[i], NULL, cpu_thread, (void *)i) == 0);
+
+    for (int round = 0; round < ROUNDS; round++) {
+        wait_for_children();
+        wake_children();
+        /* The other way into the shootdown, from this thread, while they are in
+           the middle of theirs. */
+        test_assert(msync((void *)m, MAP_BYTES, MS_SYNC) == 0);
+        test_assert(fsync(fd) == 0);
+    }
+
+    wait_for_children();
+    done = 1;
+    wake_children();
+    for (int i = 0; i < np; i++)
+        test_assert(pthread_join(threads[i], NULL) == 0);
+
+    test_assert(munmap((void *)m, MAP_BYTES) == 0);
+    test_assert(close(fd) == 0);
+}
--- a/src/kernel/flush.c
+++ b/src/kernel/flush.c
@@ -75,9 +75,14 @@ static boolean flush_gen_rlocked(word gen, cpuinfo ci, boolean full_flush)
 static void _flush_handler(void)
 {
     cpuinfo ci = current_cpu();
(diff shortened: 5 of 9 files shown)


src/kernel/kernel.h
```diff
@@ -179,6 +179,39 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
     irq_restore(flags);
 }
 
+/* For a lock whose holder can end up parked in a TLB flush rendezvous. Waiting for one of those
+ * with interrupts off is a deadlock: the flush IPI never arrives, so this processor never joins,
+ * so the rendezvous never ends, so its holder never lets go. Caught in the act with qemu's stub,
```
francescolavra: "Caught in the act with qemu's stub" and everything that follows in this comment adds more noise than useful info. Remove it.

src/kernel/kernel.h
```diff
@@ -179,6 +179,39 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
...
+ * all four processors stopped: one owned the rendezvous and waited for everyone to join; one held
+ * the process's vmap lock and had joined, waiting for the rendezvous to end (madvise() unmapping
+ * pages, page.c:626); one was here, spinning for that same vmap lock with interrupts off and
+ * unable to answer anything. So the wait does the flush work itself, which is what
+ * get_page_flush_entry() already does when it spins for the same reason. */
+static inline u64 spin_lock_irq_flushing(spinlock l)
```
francescolavra: I would call it `spin_lock_irq_flush`

src/kernel/kernel.h
```diff
@@ -179,6 +179,39 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
...
+{
+    u64 flags = irq_disable_save();
+    /* Read before the exchange, as spin_lock() does: a lock this hot -- every mmap, munmap and
+     * fault takes it -- would otherwise have every waiter writing the same line on every pass.
+     * The flush work is asked for only when there is some, so a spin that is merely waiting its
+     * turn costs what it always cost. */
```
francescolavra: This comment has nothing to do with flushing. Remove it

src/kernel/kernel.h
```diff
@@ -179,6 +179,39 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
...
+    volatile u64 *p = (volatile u64 *)&l->w;
+#ifdef LOCK_STATS
+    u64 spins = 0;
+#endif
+    while (*p || !compare_and_swap_64(&l->w, 0, 1)) {
+#ifdef LOCK_STATS
+        spins++;
+#endif
+        if (page_invalidate_pending())
+            page_invalidate_flush();
```
francescolavra: Instead of having a separate page_invalidate_pending() function, I would modify _flush_handler() in flush.c to check if there are any flushes pending before taking the flush lock. Something like `u64 gen_diff = inval_gen - ci->inval_gen; if (!gen_diff) return;`.
So here we just call page_invalidate_flush()

src/unix/unix_internal.h
```diff
@@ -828,7 +828,9 @@ boolean unix_timers_init(unix_heaps uh);
 
 #define sysreturn_from_pointer(__x) ((s64)u64_from_pointer(__x));
 
-#define vmap_lock(p) u64 _savedflags = spin_lock_irq(&(p)->vmap_lock)
+/* Taken and then held across the unmapping of pages, which waits for every processor to flush:
```
francescolavra: "which may wait for every processor to flush" (it doesn't always do)

src/unix/unix_internal.h
```diff
@@ -828,7 +828,9 @@ boolean unix_timers_init(unix_heaps uh);
...
+ * see spin_lock_irq_flushing(), which is why this one is not spin_lock_irq(). */
```
francescolavra: This is not useful info, remove it

test/runtime/Makefile
```diff
@@ -40,6 +40,7 @@ PROGRAMS= \
 	thread_test \
 	time \
 	tlbshootdown \
+	vmap_flush_race \
```
francescolavra: Instead of a separate test program, I would integrate this test into the existing mmap runtime test. Keep this as a separate source file (mmap.c is becoming too big), and call its main function from main() in mmap.c.
It would be good to reuse the TEST_FUNC() macro that is currently in tests/runtime/filesystem.c: move it to test_utils.h, then just call `TEST_FUNC(vmap_flush)` from main() in mmap.c.

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
+/* More than one thread giving pages back at the same time, while others fault
+   on them and the file they came from is synced.
+
+   tlbshootdown covers the neighbouring case: one thread unmaps while the others
+   read, and each of them takes the fault it should. What it never has is two
+   threads returning pages at once, and that is what closes a cycle -- madvise
+   takes the process's vmap lock with interrupts disabled and then unmaps, which
+   opens a shootdown every processor has to join; a second thread waiting for
```
francescolavra: "opens a shootdown every processor may have to join synchronously"

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+   that same lock with interrupts disabled never services the shootdown, so it
+   never joins, so the first never finishes and never lets the lock go. The page
+   tables are locked the same way by the fault and by the unmap, so the same
+   cycle closes there too.
```
francescolavra: This is incorrect, the page tables are not locked while handling a shootdown.
And everything that follows in this comment is more noise than anything else. Remove it.

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+#define PAGESIZE    4096
+#define MAP_PAGES   2048
+#define MAP_BYTES   (MAP_PAGES * PAGESIZE)
+#define ROUNDS      64
+
+static pthread_t threads[MAX_CPUS];
+static volatile uint8_t *m;
+static volatile int done;
+static int np;
+static int fd;
```
francescolavra: No reason to make this a static variable, it's used in the main function only

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+
+    while (!done) {
+        wait_for_sync();
+        if (done)
+            break;
+        if (id & 1) {
+            for (size_t off = base; off < base + share; off += PAGESIZE)
+                m[off] = (uint8_t)(id + 1);
+        } else {
+            test_assert(madvise((void *)(m + base), share, MADV_DONTNEED) == 0);
```
francescolavra: No need to have two kinds of threads; all of them can do the same thing (i.e. fault in a chunk of memory and give it back), and that is enough to trigger the issue. So no need for `if (id & 1)`: all threads do both the for loop and the madvise() call

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+            test_assert(madvise((void *)m, MAP_BYTES, MADV_DONTNEED) == 0);
```
francescolavra: This is not needed and can be removed

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+    }
+    return NULL;
+}
+
+int main(int argc, char **argv)
+{
+    np = get_nprocs();
+    if (np > MAX_CPUS)
+        np = MAX_CPUS;
+    printf("There are %d processors available\n", np);
```
francescolavra: Since we want this to be integrated in the CI test suite, it should be quiet by default, i.e. print something only if it needs attention. So remove this printf

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+                m[off] = (uint8_t)(id + 1);
+        } else {
+            test_assert(madvise((void *)(m + base), share, MADV_DONTNEED) == 0);
+            test_assert(madvise((void *)m, MAP_BYTES, MADV_DONTNEED) == 0);
+        }
+    }
+    return NULL;
+}
+
+int main(int argc, char **argv)
```
francescolavra: This can be changed to `void test_vmap_flush(int argc, char **argv)`, so that it can be called via the TEST_FUNC() macro

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+    pthread_cond_init(&kid_cv, NULL);
+    pthread_cond_init(&sync_cv, NULL);
+    pthread_mutex_init(&sync_mut, NULL);
+
+    fd = memfd_create("race", 0);
+    test_assert(fd >= 0);
+    test_assert(ftruncate(fd, MAP_BYTES) == 0);
+    m = mmap(NULL, MAP_BYTES, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+    test_assert(m != MAP_FAILED);
+    memset((void *)m, 0x11, MAP_BYTES);
```
francescolavra: This memset is not needed

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+
+    for (long i = 0; i < np; i++)
+        test_assert(pthread_create(&threads[i], NULL, cpu_thread, (void *)i) == 0);
+
+    for (int round = 0; round < ROUNDS; round++) {
+        printf("round %d\n", round);
```
francescolavra: No printf

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+        fflush(stdout);
+        wait_for_children();
+        wake_children();
+        /* The other way into the shootdown, from this thread, while they are in
+           the middle of theirs. */
+        test_assert(msync((void *)m, MAP_BYTES, MS_SYNC) == 0);
+        test_assert(fsync(fd) == 0);
+        for (size_t off = 0; off < MAP_BYTES; off += PAGESIZE)
+            m[off] = 0x44;
```
francescolavra: This for loop is also not needed

test/runtime/vmap_flush_race.c
```diff
@@ -0,0 +1,152 @@
...
+
+    wait_for_children();
+    done = 1;
+    wake_children();
+    for (int i = 0; i < np; i++)
+        test_assert(pthread_join(threads[i], NULL) == 0);
+
+    test_assert(munmap((void *)m, MAP_BYTES) == 0);
+    test_assert(close(fd) == 0);
+    printf("vmap flush race test passed\n");
```
francescolavra: No printf

francescolavra (changes requested):
Commit messages should be concise and to the point, not noisy and full of irrelevant (or incorrect) info.
For the first commit, everything starting from "madvise() and munmap() hold the vmap lock across ..." can be replaced by a single sentence: "Ensure that a thread waiting for the vmap lock answers any pending flush."
For the second commit, "Add a test case that exercises the race fixed by the previous commit" is enough.

AngeloRubens:
Hi @francescolavra , Thanks -- all applied.

- `page_invalidate_pending()` is gone: `_flush_handler()` now returns early when there is
  nothing pending, as you suggested, and the spin just calls `page_invalidate_flush()`.
- Renamed to `spin_lock_irq_flush()`.
- Dropped the qemu stub story, the comment about reading before the exchange, and the comment
  on `vmap_lock`. Also dropped the sentence claiming the page tables are locked while handling a
  shootdown, which was wrong.
- The test is no longer a program of its own: `TEST_FUNC()` moved from `filesystem.c` to
  `test_utils.h`, and `vmap_flush.c` exposes `test_vmap_flush()`, called from `main()` in
  `mmap.c`. All threads now do the same thing, the extra madvise, the memset and the trailing
  loop are gone, `fd` is local, and nothing is printed. `mmap.manifest` gained the tmpfs and
  shmem klibs, which `memfd_create()` needs.
- Commit messages cut down to the two sentences you gave.

src/kernel/kernel.h
```diff
@@ -179,6 +179,30 @@ static inline void spin_unlock_irq(spinlock l, u64 flags)
     irq_restore(flags);
 }
 
+/* For a lock whose holder can end up parked in a TLB flush rendezvous. Waiting for one of those
+ * with interrupts off is a deadlock: the flush IPI never arrives, so this processor never joins,
+ * so the rendezvous never ends, so its holder never lets go. The wait does the flush work itself,
+ * which is what get_page_flush_entry() already does when it spins for the same reason. */
```
francescolavra: No need to mention get_page_flush_entry() here

test/runtime/vmap_flush.c
```diff
@@ -0,0 +1,132 @@
...
+   read, and each of them takes the fault it should. What it never has is two
+   threads returning pages at once, and that is what closes a cycle -- madvise
+   takes the process's vmap lock with interrupts disabled and then unmaps, which
+   opens a shootdown every processor may have to join synchronously; a second
+   thread waiting for that same lock with interrupts disabled never services the
+   shootdown, so it never joins, so the first never finishes and never lets the
+   lock go.
+
+   Two processors are enough. A kernel that closes the cycle stops the machine,
+   so there is no message to print: the test either returns or the log ends.
```
francescolavra: This sentence and the following one in this comment are more noise than anything else. Remove them

francescolavra (changes requested):

AngeloRubens:
Hi @francescolavra ,

All three applied.

- The mention of `get_page_flush_entry()` is out of the comment in `kernel.h`.
- The two sentences you pointed at are out of the comment in `vmap_flush.c`.
- The first commit message is down to the one sentence you gave.

The red `build-pc` is not this change: it is `TestE2E/packages/python_alloc` timing out at 120s,
and pull/2185 failed the same way in the same minutes. The commit before this one, which differs
only in those two comments, was green on all four jobs.

PR #2180, opened 2026-08-20
Enable JVM ZGC Uncommit - fs: give back the pages of a punched hole on tmpfsPr/tmpfs uncommit

AngeloRubens:
What this is for. On nanos a JVM cannot give memory back. ZGC's uncommitter
punches holes in the memfd that backs its heap, and fallocate(PUNCH_HOLE) on a
memory filesystem zeroed the pages where they were and kept them: the file
stopped accounting for them, st_blocks went to zero, and not one kilobyte
returned to the system. The collector was told it had succeeded and got nothing,
so the elastic heap was inert. These three commits make it real.

What they do. fs: give back the pages of a punched hole on tmpfs gives
tmpfs a dealloc method: whole pages inside the hole are dropped and their memory
returned, a partial page at either edge is zeroed in place as before, and the
ranges are taken out of the file's dirty map with the pins taken at write-back
released -- both in batches with the filesystem's mutex dropped in between,
because the node's pages are covered by a spin lock while that mutex is held and
the write-back takes the two the other way round. A page still referenced when
the hole reaches it is zeroed rather than freed, and the range reads back as the
hole it has become either way.

Where the code moved. filesystem_dealloc() lived in tfs.c and did one thing:
zero the range through the page cache. It is now in fs.c, where it asks the
filesystem for a dealloc method and falls back to that same zeroing when there is
none, with the zeroing itself pulled out into a helper, fsfile_zero_range(), so
both paths share it. Only tmpfs implements the method. The filesystem on disk
keeps the behaviour it had, through the code it had, which is why the diff takes
those lines out of tfs.c rather than changing them.

test: a hole punched in a memory filesystem has to give the blocks back is the
reproducer: it fails on an unmodified tree, where the block count after a punch
is the block count before it.

fs: count a tmpfs file's pages instead of walking them at every stat closes a
race the change above makes reachable. st_blocks was answered by walking the file's
dirty range map without a lock, which was harmless while nothing ever removed a
node from that map; punching a hole removes ranges and frees the nodes behind
them, so a stat running beside a punch could follow a record that had gone.
The count is now kept as a running total, moved where the map is moved and under
the same lock, and stat reads it and converts it exactly as before.

It needs the vmap lock fix(https://github.com/nanovms/nanos/pull/2179). With the punch returning memory, the pattern of
a collector faulting pages back in while another thread unmaps them reaches a
pre-existing deadlock between the process's mapping lock and the TLB shootdown
rendezvous, and the guest stops within seconds. That is sent separately as
kernel: do not wait for the vmap lock deaf to the flush that needs an answer,
because it is an independent defect that reproduces without any of this. Read
that one first.

Measured. On x86_64 with hardware virtualisation, against the same kernel
without these changes: 8192kB reclaimed of an 8192kB hole where nothing was
reclaimed before, and 288M allocatable after punching 200M in a 512M guest. With
the vmap lock fix applied, a JVM running ZGC with an elastic heap on four
processors ran thirty minutes of sustained load and returned 21190 MB across 738
uncommits, with 1.96 billion pages written and read back through mappings of the
file, nothing reading back wrong, and no fault or assertion in the kernel; a
second run of the same shape returned 50404 MB across 1164 uncommits.

### DIFF
--- a/klib/tmpfs.c
+++ b/klib/tmpfs.c
@@ -8,6 +8,7 @@
 typedef struct tmpfs_file {
     struct fsfile f;
     struct rangemap dirty;
+    u64 dirty_pages;    /* number of page indices in the dirty range map */
     u64 seals;
     closure_struct(pagecache_node_reserve, reserve);
     closure_struct(thunk, free);
@@ -31,10 +32,13 @@ static void tmpfsfile_read(fsfile f,
 }
 
 closure_function(2, 1, boolean, tmpfsfile_write_handler,
-                 rangemap, dirty, pagecache_node, pn,
+                 tmpfs_file, fsf, pagecache_node, pn,
                  range r)
 {
-    if (rangemap_insert_range(bound(dirty), r)) {
+    tmpfs_file fsf = bound(fsf);
+    if (rangemap_insert_range(&fsf->dirty, r)) {
+        fsf->dirty_pages += range_span(r);
+
         /* Pin the pages so that they cannot be evicted from the page cache. */
         pagecache_nodelocked_pin(bound(pn), r);
 
@@ -43,17 +47,61 @@ closure_function(2, 1, boolean, tmpfsfile_write_handler,
     return true;
 }
 
+/* Removes a range of pages from the dirty range map and drops the pin taken when they were first
+   written, so that the memory they hold can be given back. Called with the page cache node
+   locked, hence the node-locked unpin. */
+static void tmpfsfile_punch(tmpfs_file fsf, range pages)
+{
+    rangemap dirty = &fsf->dirty;
+    pagecache_node pn = fsf->f.cache_node;
+    rmnode n = rangemap_lookup_at_or_next(dirty, pages.start);
+    while ((n != INVALID_ADDRESS) && (n->r.start < pages.end)) {
+        rmnode next = rangemap_next_node(dirty, n);
+        range r = n->r;
+        range i = range_intersection(r, pages);
+        boolean head = r.start < pages.start;
+        boolean tail = r.end > pages.end;
+        if (head && tail) {
+            /* The hole falls inside this range, so no other range intersects it. */
+            assert(rangemap_reinsert(dirty, n, irange(r.start, pages.start)));
+            if (!rangemap_insert_range(dirty, irange(pages.end, r.end))) {
+                assert(rangemap_reinsert(dirty, n, r));  /* out of memory: keep the range whole */
+                return;
+            }
+        } else if (head) {
+            assert(rangemap_reinsert(dirty, n, irange(r.start, pages.start)));
+        } else if (tail) {
+            assert(rangemap_reinsert(dirty, n, irange(pages.end, r.end)));
+        } else {
+            rangemap_remove_range(dirty, n);
+        }
+        pagecache_nodelocked_unpin(pn, i);
+        fsf->dirty_pages -= range_span(i);
+        if (head && tail)
+            return;
+        n = next;
+    }
+}
+
 static void tmpfsfile_write(fsfile f,
                    sg_list sg, range q, status_handler complete)
 {
     tmpfs_file fsf = (tmpfs_file)f;
     tmpfs fs = (tmpfs)fsf->f.fs;
-    range pages = range_rshift_pad(q, fs->page_order);
     int res;
+    if (!sg) {
+        /* The range is being punched out: give its pages back. */
+        filesystem_lock(&fs->fs);
+        tmpfsfile_punch(fsf, range_rshift(q, fs->page_order));
+        filesystem_unlock(&fs->fs);
+        async_apply_status_handler(complete, STATUS_OK);
+        return;
+    }
+    range pages = range_rshift_pad(q, fs->page_order);
     filesystem_lock(&fs->fs);
     do {
         res = rangemap_range_find_gaps(&fsf->dirty, pages,
-                                       stack_closure(tmpfsfile_write_handler, &fsf->dirty,
+                                       stack_closure(tmpfsfile_write_handler, fsf,
                                                      fsf->f.cache_node));
     } while (res == RM_ABORT);
     filesystem_unlock(&fs->fs);
@@ -113,10 +161,7 @@ closure_func_basic(thunk, void, tmpfsfile_free)
 
 static s64 tmpfsfile_get_blocks(fsfile f)
 {
-    s64 pages = 0;
-    rangemap_foreach(&((tmpfs_file)f)->dirty, n) {
-        pages += range_span(n->r);
-    }
+    u64 pages = ((tmpfs_file)f)->dirty_pages;
     return (pages << ((tmpfs)f->fs)->page_order) >> SECTOR_OFFSET;
 }
 
@@ -139,6 +184,7 @@ static int tmpfs_create(filesystem fs, tuple parent, string name, tuple md, fsfi
             return fss;
         }
         init_rangemap(&fsf->dirty, h);
+        fsf->dirty_pages = 0;
         fsf->seals = 0;
         fsf->f.get_blocks = tmpfsfile_get_blocks;
         if (f)
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -1358,6 +1358,12 @@ closure_func_basic(pp_handler, boolean, pagecache_unpin_handler,
     return true;
 }
 
+void pagecache_nodelocked_unpin(pagecache_node pn, range pages)
+{
+    pagecache_nodelocked_traverse(pn, pages,
+                                  stack_closure_func(pp_handler, pagecache_unpin_handler));
+}
+
 void pagecache_node_unpin(pagecache_node pn, range pages)
 {
     pagecache_node_traverse(pn, pages, stack_closure_func(pp_handler, pagecache_unpin_handler));
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -38,6 +38,11 @@ typedef struct pagecache_drain_work {
     int freed_pages;
 } *pagecache_drain_work;
 
+struct pagecache_unmap_range {
+    range v;                    /* virtual */
+    u64 node_offset;
+};
+
 #define pagecache_lock_mappings()   pagecache_lock(global_pagecache)
 #define pagecache_unlock_mappings() pagecache_unlock(global_pagecache)
 
@@ -514,6 +519,19 @@ static pagecache_page page_lookup_nodelocked(pagecache_node pn, u64 n)
     return (pagecache_page)rbtree_lookup(&pn->pages, &k.rbnode);
 }
 
+/* The first page at or after n, which may be none. */
+static pagecache_page page_lookup_at_or_next_nodelocked(pagecache_node pn, u64 n)
+{
+    struct pagecache_page k;
+    k.state_offset = n;
+    pagecache_page pp = (pagecache_page)rbtree_lookup_max_lte(&pn->pages, &k.rbnode);
+    if (pp == INVALID_ADDRESS)
+        return (pagecache_page)rbtree_find_first(&pn->pages);
+    if (page_offset(pp) < n)
+        pp = (pagecache_page)rbnode_get_next(&pp->rbnode);
+    return pp;
+}
+
 static pagecache_page page_lookup_or_alloc_nodelocked(pagecache_node pn, u64 n)
 {
     pagecache_page pp = page_lookup_nodelocked(pn, n);
@@ -568,6 +586,8 @@ static void pagecache_node_traverse(pagecache_node pn, range pages, pp_handler h
     pagecache_nodelocked_traverse(pn, pages, handler);
     pagecache_unlock_node(pn);
 }
+
+static void pagecache_node_punch(pagecache_node pn, range q /* bytes */, status_handler complete);
 
 closure_function(6, 1, void, pagecache_write_sg_finish,
                  pagecache_node, pn, range, q, u64, pi, sg_list, sg, status_handler, completion, context, saved_ctx,
@@ -681,6 +701,24 @@ closure_function(1, 3, void, pagecache_write_sg,
     if (range_span(q) == 0) {
         apply(completion, STATUS_OK);
         return;
+    }
+
+    if (!sg) {
+        /* A null sg punches a hole out of the node: the whole pages of the range are given back,
+           while a partial page at either edge is zeroed by the code below. */
+        range body = irange(pad(q.start, U64_FROM_BIT(pc->page_order)),
+                            q.end & ~MASK(pc->page_order));
+        if (body.end > body.start) {
+            merge m = allocate_merge(pc->h, completion);
+            status_handler sh = apply_merge(m);
+            if (q.start < body.start)
+                apply(pn->cache_write, 0, irange(q.start, body.start), apply_merge(m));
+            if (q.end > body.end)
+                apply(pn->cache_write, 0, irange(body.end, q.end), apply_merge(m));
+            pagecache_node_punch(pn, body, apply_merge(m));
+            apply(sh, STATUS_OK);
+            return;
+        }
     }
 
     u64 start_offset = q.start & MASK(pc->page_order);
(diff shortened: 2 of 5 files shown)


test/runtime/Makefile
```diff
@@ -38,6 +38,7 @@ PROGRAMS= \
 	socketpair \
 	syslog \
 	thread_test \
+	tmpfs_punch \
```
francescolavra: No need for a new test program (and not even for a new source file), we can just add the new test cases to fallocate.c, which is run as part of the filesystem runtime tests. You will have to add the shmem and tmpfs klibs to filesystem.manifest

test/runtime/tmpfs_punch.c
```diff
@@ -0,0 +1,155 @@
...
+   heap by punching holes in a memfd -- which is what ZGC's uncommitter does --
+   frees nothing at all.
+
+   The second is the part that does not look like it needs a test. On a memory
+   filesystem the pages of a file are the file, so writing to one pins its pages
+   and the pages of a punched range are given back and unpinned. If the record of
+   such a page is left behind holding no memory, the next write pins it again and
+   the next unpin releases it a second time, handing an address that was never
+   allocated to the page heap. Hence the write-punch-write-close below: each step
+   is ordinary, and it is only the order that finds it. */
```
francescolavra: This whole set of paragraphs is mostly a repetition of what is written elsewhere, so should be removed

test/runtime/tmpfs_punch.c
```diff
@@ -0,0 +1,155 @@
...
+    test_assert(munmap(p, FILE_SIZE) == 0);
+    test_assert(close(fd) == 0);
+}
+
+int main(int argc, char **argv)
+{
+    test_punch_frees_blocks();
+    test_rewrite_after_punch();
+    test_punch_under_mapping();
+    printf("tmpfs punch test passed\n");
```
francescolavra: The calls to the 3 functions above can go in test_fallocate() in fallocate.c. And no printf() for a test that passed.

test/runtime/tmpfs_punch.c
```diff
@@ -0,0 +1,161 @@
...
+    test_assert(fstat(fd, &st) == 0);
+    return st.st_blocks;
+}
+
+/* The blocks of a file on a memory filesystem are counted from the ranges its
+   write handler has been given, and the page cache hands those over when it
+   commits, not when write() returns -- so a file that has only been written to
+   holds no blocks yet, and fsync is what makes the count mean anything. Keeping
+   the count as a running total instead of walking the ranges does not change
+   that: the total moves where the ranges do, which is inside the write-back. */
```
francescolavra: This comment is a convoluted way to say that we need to call fsync() in order for the allocated pages to show up in a fstat() syscall. Just write this simple sentence above the line that calls fsync(fd), and remove this comment.

francescolavra (changes requested):
It seems to me this functionality can be achieved in a much simpler way by just amending tmpfsfile_write() so that if `sg` is 0 if unpins (whole) pages from the pagecache node and removes them for the dirty rangemap. Yes, the lock inversion is a problem, but it can be solved by introducing a pagecache_nodelocked_unpin() function that does the reverse of what pagecache_nodelocked_pin() does. With these simple changes, we don't need a `dealloc` filesystem callback.
By the way, the sentence "the file stops accounting for them -- st_blocks goes to zero" in the first commit message is wrong.
Also, the third commit is logically tied to the first commit, so I would squash them together; and the description of the commit that introduces the runtime test does not need to explain again everything that is in the first commit: just "Add test cases to exercise FALLOC_FL_PUNCH_HOLE on a memfd" is enough.

AngeloRubens:
Hi @francescolavra ,
Reworked along the lines you suggested: no `dealloc` filesystem callback, a new
`pagecache_nodelocked_unpin()`, and `tmpfsfile_write()` dropping the pin and the dirty ranges
when `sg` is 0. That removes the batching the lock inversion used to need.

One thing I ran into: the null sg does not reach `fs_write` any more. It used to -- 98cae64a
("fallocate / filesystem_dealloc: use null sg to zero data in pagecache, which gets propagated
to tfs storage write for extent trimming") -- but that propagation was lost when the page cache
moved to write-back. Today `pagecache_write_sg()` zeroes the pages and marks them dirty, and
`pagecache_commit_dirty_ranges()` always calls `fs_write` with a real sg, so the `!sg` branch of
`extents_range_handler()` is unreachable: punching a hole on tfs has not been trimming extents
either.

So I restored it, in the page cache rather than per filesystem: a write with a null sg now
punches the whole pages of the range -- the dirty ranges over them are dropped, `fs_write` is
applied with the null sg **with the node locked**, the way the write-back applies it, which is
what makes `pagecache_nodelocked_unpin()` the right primitive, and the memory is given back once
the pages have been unmapped. A partial page at either edge keeps its storage and is only
zeroed, as before. The whole pages of the range no longer have cache pages allocated and zeroed
for them on the way to being freed, which is what the old path did.

Also: the wrong sentence about `st_blocks` is gone, the third commit is squashed into the first,
and the test cases are in `test_fallocate()` in `fallocate.c`, with the shmem and tmpfs klibs
added to `filesystem.manifest` and no output on success.

klib/tmpfs.c
```diff
@@ -43,17 +47,61 @@ closure_function(2, 1, boolean, tmpfsfile_write_handler,
...
+
 static void tmpfsfile_write(fsfile f,
                    sg_list sg, range q, status_handler complete)
 {
     tmpfs_file fsf = (tmpfs_file)f;
     tmpfs fs = (tmpfs)fsf->f.fs;
-    range pages = range_rshift_pad(q, fs->page_order);
     int res;
+    if (!sg) {
+        /* The range is being punched out: give the pages back instead of writing to them. */
```
francescolavra: "instead of writing to them" does not add any useful info

klib/tmpfs.c
```diff
@@ -97,6 +145,7 @@ closure_func_basic(status_handler, void, tmpfsfile_sync_complete,
     rangemap_foreach(dirty, n) {
         pagecache_node_unpin(pn, n->r);
     }
+    fsf->dirty_pages = 0;
```
francescolavra: This function deallocates the whole tmpfs_file structure, there is no point in setting one of its members to 0.

src/kernel/pagecache.c
```diff
@@ -569,6 +569,69 @@ static void pagecache_node_traverse(pagecache_node pn, range pages, pp_handler h
     pagecache_unlock_node(pn);
 }
 
+static void pagecache_node_free_pages(pagecache_node pn, range q /* bytes */);
```
francescolavra: This forward function declaration doesn't seem necessary, you can reorder the function definitions as needed.

src/kernel/pagecache.c
```diff
@@ -569,6 +569,69 @@ static void pagecache_node_traverse(pagecache_node pn, range pages, pp_handler h
...
+
+/* Drops the dirty ranges over a range that is being punched out: a hole has nothing to write
+   back, and the pages of the range are about to be given back. Called with the node and state
```
francescolavra: "a hole has nothing to write back, and the pages of the range are about to be given back" is an obvious and redundant statement

src/kernel/pagecache.c
```diff
@@ -569,6 +569,69 @@ static void pagecache_node_traverse(pagecache_node pn, range pages, pp_handler h
...
+    pagecache pc = pn->pv->pc;
+    pagecache_debug("%s: pn %p, q %R\n", func_ss, pn, q);
+    pagecache_lock_node(pn);
+    pagecache_lock_state(pc);
+    pagecache_punch_dirty_locked(pn, q);
+    pagecache_unlock_state(pc);
+
+    /* Applied with the node locked, as the write-back applies it, so that a filesystem which
+       gives up its own references here (tmpfs unpins the pages it has written) can do so without
+       locking the node a second time. */
```
francescolavra: "so that a filesystem which..." is irrelevant information here. Remove it.

src/kernel/pagecache.c
```diff
@@ -1902,6 +1989,128 @@ void *pagecache_get_page_if_filled(pagecache_node pn, u64 node_offset, boolean p
...
+/* Number of pages freed per hold of the node lock. */
+#define FREE_PAGES_BATCH 32
+
+struct pagecache_unmap_range {
+    range v;                    /* virtual */
+    u64 node_offset;
+};
+
+/* Collects the mappings intersecting a node range, resuming from the virtual address in cursor;
+   the mappings lock cannot be held while unmapping, hence the batching. */
```
francescolavra: Why can't the mappings lock cannot be held while unmapping?

src/kernel/pagecache.c
```diff
@@ -569,6 +569,69 @@ static void pagecache_node_traverse(pagecache_node pn, range pages, pp_handler h
...
+   to release the storage behind the range, and the memory held by the pages of the range is
+   given back. Only whole pages can be punched; a partial page at either edge of the requested
+   range keeps its storage and is only zeroed, by the ordinary write path. */
+static void pagecache_node_punch(pagecache_node pn, range q /* bytes */, status_handler complete)
+{
+    pagecache pc = pn->pv->pc;
+    pagecache_debug("%s: pn %p, q %R\n", func_ss, pn, q);
+    pagecache_lock_node(pn);
+    pagecache_lock_state(pc);
+    pagecache_punch_dirty_locked(pn, q);
```
francescolavra: This does more or less what the existing pagecache_purge_node() function already does (except that pagecache_purge_node() operates on the whole node instead of a selected range). We should find a way to consolidate this functionality into the existing code instead of duplicating code.

src/kernel/pagecache.c
```diff
@@ -1902,6 +1989,128 @@ void *pagecache_get_page_if_filled(pagecache_node pn, u64 node_offset, boolean p
...
+    pagecache_debug("%s: pn %p, q %R\n", func_ss, pn, q);
+
+    struct pagecache_unmap_range maps[16];
+    u64 cursor = 0;
+    while (cursor != infinity) {
+        int count = pagecache_node_collect_maps(pn, q, &cursor, maps, _countof(maps));
+        for (int i = 0; i < count; i++)
+            /* The unmapping must be synchronous: the asynchronous variant releases the reference
+               each mapping holds on a page from a completion that runs after this returns, and it
+               dereferences the page to do so, while the loop below is already freeing it. */
```
francescolavra: The loop below is not freeing pages, is releasing their refcount, so the actual freeing occurs when the last refcount is dropped. So I don't think the unmapping must be synchronous.

src/kernel/pagecache.c
```diff
@@ -1902,6 +1989,128 @@ void *pagecache_get_page_if_filled(pagecache_node pn, u64 node_offset, boolean p
...
+        int n = 0;
+        pagecache_lock_node(pn);
+        /* The first page at or after next: an exact lookup would miss a range whose leading
+           pages were never faulted in, which is the common case for a punch. */
+        struct pagecache_page k;
+        k.state_offset = next;
+        pagecache_page pp = (pagecache_page)rbtree_lookup_max_lte(&pn->pages, &k.rbnode);
+        if ((pp != INVALID_ADDRESS) && (page_offset(pp) < next))
+            pp = (pagecache_page)rbnode_get_next(&pp->rbnode);
+        while ((pp != INVALID_ADDRESS) && (n < FREE_PAGES_BATCH)) {
```
francescolavra: I don't think splitting this thing in batches is justified: the operations done on each pages should be pretty fast, except when a page needs to be zeroed (in which case we call pagecache_unlock_state() before zeroing and pagecache_lock_state() afterwards). So I would simplify the code and do without batching.

francescolavra (changes requested):
The code that handles the NULL sg in the page cache is logically separate from the code that handles the NULL sg in tmpfs, so the two sets of changes should be in two separate commits (first the page cache changes, then the tmpfs changes).

AngeloRubens:
Hi @francescolavra ,

Split as you asked -- the page cache changes are in the first commit and the tmpfs ones in the
second, and each builds on its own.

- The dirty-drop loop is shared with `pagecache_purge_node()` now: it is
  `pagecache_purge_dirty_locked()`, and `purge_range_handler()` is reduced to calling it and
  removing the range. What is left of `pagecache_punch_dirty_locked()` is the range map
  splitting, which a whole-node purge does not need, so it moved into `pagecache_node_punch()`
  and the separate function is gone.
- No batching: the node lock is held for the walk and the state lock per page, released only
  around the zeroing. The per-page reference stays, since it is what keeps `kvirt` valid across
  that release.
- Both forward declarations are gone, the punch being defined after
  `pagecache_node_unmap_pages()` now. One is left, for `pagecache_node_punch()` itself, which
  `pagecache_write_sg()` calls well before that point; removing it would mean moving
  `pagecache_node_unmap_pages_sync()` and everything it rests on above the write path.
- `fsf->dirty_pages = 0` in the deallocation is gone, along with the comments you pointed at.

Two things I did not change, and why.

The unmapping does have to be synchronous, though not for the reason the comment gave -- you are
right that the loop only releases references. It is that
`pagecache_node_unmap_pages_complete()` drops the reference each mapping holds on a page, and
with the asynchronous variant it does so after this returns: the loop would then read every page
as still referenced and zero it instead of giving it back, which is the whole point of the
patch. The comment says that now.

On the mappings lock: it is dropped before unmapping because the unmapping waits for a TLB flush
that the processors holding it may have to join. That is what every other user of it does --
`pagecache_drain()`, `pagecache_scan_shared_mappings()`, `pagecache_node_scan()` and
`pagecache_node_unmap_pages()` all collect under the lock, release it, then call
`page_invalidate_sync()`. The comment says that instead of asserting it.

AngeloRubens:
One measurement, since the batching turned out not to be free.

The test: a 256 MB memfd with every page resident, mapped shared. One thread punches 28 MB
holes in a loop and stops after a fixed 20000 of them, so each build is asked for the same
number of punches; three other threads read pages of the same file in a stride and time each
read. A punched page has been freed and unmapped, so reading it faults, and the fault waits for
the node the punch is holding. No madvise and no timers, so the run is the same work every time.
With the puncher idle the same reads average 209 ns and never fault, which is the floor.

Median of three runs each, four processors, KVM, run locally and not in the CI:

```
  batching       397k reads   mean  77 us   p90 334 us   p99 599 us
  no batching    273k reads   mean 101 us   p90 378 us   p99 838 us
```

The ranges do not overlap. The punches are equal in number but not in work: each one frees only
what the readers have faulted back in since the last pass over that window, so the build with
the batching, whose readers got further, also had more pages to free. That understates the
difference rather than inflating it. A third build carrying everything else in this rework, with
only the batching restored, lands at 361k -- so what costs is the node lock held across the
walk, not the dirty-drop loop now shared with pagecache_purge_node().

The rate matters more than the number: that is 2000 punches a second. The ZGC soak this is for
uncommitted 0.4 times a second, 28 MB at a time -- I counted uncommits, not fallocate calls, so
the call rate is that or a small multiple of it. Orders of magnitude below either way, and at
that rate none of this is observable, which is why the code is as you asked for it.

I can post the bench if it is worth having.

src/kernel/pagecache.c
```diff
@@ -2070,6 +2108,138 @@ void pagecache_node_unmap_pages(pagecache_node pn, range v /* bytes */, u64 node
...
+
+    range pages = range_rshift(q, page_order);
+    pagecache_lock_node(pn);
+    /* The first page at or after the start of the hole: an exact lookup would miss a range whose
+       leading pages were never faulted in, which is the common case for a punch. */
+    struct pagecache_page k;
+    k.state_offset = pages.start;
+    pagecache_page pp = (pagecache_page)rbtree_lookup_max_lte(&pn->pages, &k.rbnode);
+    if ((pp != INVALID_ADDRESS) && (page_offset(pp) < pages.start))
+        pp = (pagecache_page)rbnode_get_next(&pp->rbnode);
```
francescolavra: It seems to me the above lines can be replaced by `pagecache_page pp = (pagecache_page)rangemap_lookup_at_or_next(&pn->pages, pages.start);`

src/kernel/pagecache.c
```diff
@@ -1313,6 +1345,12 @@ closure_func_basic(pp_handler, boolean, pagecache_unpin_handler,
     return true;
 }
 
+void pagecache_nodelocked_unpin(pagecache_node pn, range pages)
```
francescolavra, line 1361: This function definition (and its declaration in the header file) belongs to the commit that uses it, i.e. the second commit

test/runtime/fallocate.c
```diff
@@ -26,6 +28,116 @@ static void test_tmpfile(void)
...
+    for (size_t i = 0; i < sizeof(buf); i++)
+        test_assert(buf[i] == 0xa5);
+
+    test_assert(fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, 0, MEMFD_SIZE) == 0);
+    test_assert(blocks_of(fd) < holed);
+
+    test_assert(close(fd) == 0);
+}
+
+/* Writing over a punched range, and then dropping the file: the pages of the range are pinned by
```
francescolavra: "and then dropping the file" is not something specific done by this function, it's just a cleanup thing done by the other functions too, so there is no reason to mention it

francescolavra (changes requested):

AngeloRubens:
`pn->pages` is an rbtree, not a rangemap, so `rangemap_lookup_at_or_next()` does not apply to it. The lookup was wrong all the same, and for more than tidiness: `rbtree_lookup_max_lte()` returns `INVALID_ADDRESS` when no page precedes the hole -- as `test/unit/rbtree_test.c` itself asserts -- and the walk was then skipped entirely. A punch that starts before the first page a file has, which is what a collector uncommitting from the start of its heap does, gave nothing back and left the pages readable with their old contents.

It is now `page_lookup_at_or_next_nodelocked()`, next to `page_lookup_nodelocked()`, falling back to the first node the way `rangemap_lookup_at_or_next()` does internally. The case for it is at the head of `test_memfd_punch()`, on a file whose first pages were never written: with the old lookup restored it fails on both arm64 and x86_64, at the read that must come back zero.

`pagecache_nodelocked_unpin()` and its declaration have moved to the second commit, and the comment in `fallocate.c` is gone. Rebased on master.

francescolavra (approved):

PR #2181, opened 2026-08-21
THP for tmpfs: Memory mapping: extend Transparent Huge Pages to shared tmpfs mappings

AngeloRubens:
Stacked on [2179](https://github.com/nanovms/nanos/pull/2179) and [2180](https://github.com/nanovms/nanos/pull/2180): the first five commits are those two pull requests, and the
last one is what this asks to review. It is built on them because it touches the same
file that 2180 does, and because the flush rendezvous test it runs among its neighbours
is red without 2179. Once those are in, this rebases to a single commit.

Memory areas in a user process address space are paged on demand: the process faults on a
virtual address and one 4 KB page is allocated and mapped. The THP support added in [2037](https://github.com/nanovms/nanos/pull/2037)
lets the kernel allocate and map a block of up to 2 MB instead, which cuts the number of
faults a process takes over a large area and improves TLB usage, one entry covering 2 MB
where 512 were needed before.

This change extends the same treatment to shared mappings of tmpfs files, and it takes its
cue from [2037](https://github.com/nanovms/nanos/pull/2037) rather than inventing a second way of doing the same thing: the block
descriptors and the splitting helper that work added to the per-architecture page tables
are what this uses to lay a 2 MB entry down and to break one back into its pages. It adds
no architecture specific code of its own, and one path serves every architecture the kernel
builds for. What is new is only where the block comes from -- the page cache of a tmpfs
node, instead of the anonymous allocator.

Who it is for. Processes that keep a large working area in a memfd and map it shared, which
on this kernel is how tmpfs is reached: shared memory arenas, and above all a JVM running
ZGC. ZGC keeps its heap in a memfd and maps it shared, so its heap is file backed and gets
none of the above, while the collectors that keep their heap in anonymous memory -- G1,
Parallel, Shenandoah -- have had it since 2037. The bigger the area and the more often it is
faulted back in, the more that asymmetry costs, and a collector that hands memory back and
takes it again does precisely that for as long as the process runs.

What the change does. A tmpfs node can take one contiguous 2 MB block from the page heap and
lay it under a whole window of its cache pages, keeping the 512 ordinary 4 KB records over
it. The page size, the cache's page order, the pins, the dirty map and the punch are all
untouched: only the block carrying the memory is bigger, so a shared mapping of that window
can be described by a single entry. ZGC commits and uncommits in granules of exactly 2 MB,
a compile time constant of the collector, so a window is either whole or gone and never half
of each.

Three of the walks the page cache runs had to learn that one entry can cover more than one
page: the scan that writes dirty pages back now marks every page an entry covers, ageing
leaves such entries alone, and the walk that unmaps a range splits a block back into its
pages before it starts -- outside the node's lock, so that servicing a flush cannot park a
processor inside a critical section.

A window is promoted only when it holds no memory yet, which is where a collector leaves it
after a punch. Everything else falls back to the page at a time path, one window at a time:
an offset that does not line up, a private mapping, a window already partly resident, or no
contiguous block to be had. The transparent_hugepage tuple still turns the whole thing off,
2 MB is never mandatory, and nothing changes for anonymous mappings.

The tests this adds. test/runtime/tmpfs_thp asks the question from user space rather than
from inside the kernel: it touches one page of a window and uses mincore to count how many
pages of that window are resident. An entry covering 2 MB answers 512; a page at a time
mapping answers one. It covers six cases -- the aligned shared mapping that should be
promoted, an offset that does not line up, a private mapping, a 4 KB punch inside a promoted
window, an unmap of half a window, and a window already partly resident -- and then runs the
same kernel again with transparent_hugepage:never, which must put the count back to one. So
the test also demonstrates that the gate is the tuple that already exists.

test/runtime/tmpfs_thp_perf runs the cycle a collector runs against its heap -- commit,
fault, punch -- and times it in both configurations.

What those two tests measure, run by the workflow on hosted runners under TCG at VCPUS=2,
which is the CI configuration:

  x86_64    first fault    0.374 s  ->  0.098 s     342 MiB/s  ->  1304 MiB/s
            punch + fault  1.494 s  ->  0.416 s     343 MiB/s  ->  1230 MiB/s

  arm64     first fault    0.356 s  ->  0.144 s     359 MiB/s  ->   891 MiB/s
            punch + fault  1.427 s  ->  0.608 s     359 MiB/s  ->   843 MiB/s

These are timings under emulation, where a page fault costs more than it does on hardware,
so the ratios are reported and never gated. What they are offered for is the direction and
the agreement between them: four measurements on two architectures, the throughput figures
tracking the elapsed ones, and the same kernel binary answering differently only because of
the tuple.

Regression tests. These are not here to show that the change works -- the two tests above do
that. They are here to catch what it might have broken elsewhere, which is the question
worth worrying about, because three of the edits are in walks that are not tmpfs code at
all. Every filesystem goes through them, and a mistake there would not surface as a wrong
page count; it would surface somewhere unrelated to huge pages.

So the workflow also runs the neighbours that exercise those paths, on both architectures:
tmpfs_punch, tmpfs_commit_race and tmpfs_punch_race for the punch and write-back races,
release_page_race for the page release path, vmap_flush_race for the flush rendezvous --
which is exactly where the splitting had to be moved out from under the node's lock --
shmem for the memfd path, and mmap for the mapping calls themselves. They pass, and they
pass with the same output on a tree without this change, which is the point of running them:
it says the result is unchanged, not merely that the tests are green.

What is deliberately not done. Nothing is exposed under sysfs and statfs still does not
report a filesystem magic. A JVM that discovers the host supports huge pages for shared
memory switches its commit path to mmap, madvise, touch and munmap on every commit, which is
slower than the fallocate it uses now, and a JVM that discovers one of the two but not the
other fails to start at all. The point of doing this in the kernel is that the collector
needs to know nothing and change nothing.

### DIFF
--- a/src/kernel/page.c
+++ b/src/kernel/page.c
@@ -280,6 +280,34 @@ void update_map_flags(u64 vaddr, u64 length, pageflags flags)
 #endif
 }
 
+/* called with lock held */
+closure_function(1, 3, boolean, split_large_entry,
+                 flush_entry, fe,
+                 int level, u64 vaddr, pteptr entry)
+{
+    pte old_entry = pte_from_pteptr(entry);
+    if ((level == PT_PTE_LEVEL) || !pte_is_present(old_entry) ||
+        !pte_is_mapping(level, old_entry))
+        return true;
+    if (pte_split(entry, irange(0, 0)) == INVALID_PHYSICAL)
+        return false;
+    page_invalidate(bound(fe), vaddr);
+    return true;
+}
+
+/* Replaces any block mapping intersecting the range with a table of PTE-level mappings of the
+   same memory. What is mapped does not change; only the level at which it is described does.
+   The page cache walks its mappings expecting one entry per cache page, so a range it is about
+   to unmap has to be described that way before it is walked. */
+boolean split_large_mappings(u64 vaddr, u64 length)
+{
+    page_debug("vaddr 0x%lx, length 0x%lx\n", vaddr, length);
+    flush_entry fe = get_page_flush_entry();
+    boolean success = traverse_ptes(vaddr, length, stack_closure(split_large_entry, fe));
+    page_invalidate_sync(fe, 0, false);
+    return success;
+}
+
 static boolean map_level(u64 *table_ptr, int level, range v, u64 *p, u64 flags, flush_entry fe);
 
 /* called with lock held */
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -1893,13 +1893,14 @@ closure_function(3, 1, void, get_page_finish,
                  status s)
 {
     pagecache_page_handler handler = bound(handler);
+    range kvirt = irange(0, 0);
     if (is_ok(s)) {
         pagecache_page pp = bound(pp);
-        void *kvirt = !bound(private) ? pp->kvirt : pagecache_get_private_page(pp);
-        apply(handler, kvirt);
-    } else {
-        apply(handler, INVALID_ADDRESS);
+        void *p = !bound(private) ? pp->kvirt : pagecache_get_private_page(pp);
+        if (p != INVALID_ADDRESS)
+            kvirt = irangel(u64_from_pointer(p), cache_pagesize(pp->node->pv->pc));
     }
+    apply(handler, kvirt);
     closure_finish();
 }
 
@@ -1923,7 +1924,7 @@ void pagecache_get_page(pagecache_node pn, u64 node_offset, boolean private,
                     func_ss, pn, node_offset, handler, pp);
     if (pp == INVALID_ADDRESS) {
         pagecache_unlock_node(pn);
-        apply(handler, INVALID_ADDRESS);
+        apply(handler, irange(0, 0));
         return;
     }
     merge m = allocate_merge(pc->h, closure(pc->h, get_page_finish, pp, private, handler));
@@ -1934,20 +1935,20 @@ void pagecache_get_page(pagecache_node pn, u64 node_offset, boolean private,
 }
 
 /* no-alloc / no-fill path */
-void *pagecache_get_page_if_filled(pagecache_node pn, u64 node_offset, boolean private)
+range pagecache_get_page_if_filled(pagecache_node pn, u64 node_offset, boolean private)
 {
+    pagecache pc = pn->pv->pc;
     pagecache_lock_node(pn);
-    pagecache_page pp = page_lookup_nodelocked(pn, node_offset >> pn->pv->pc->page_order);
+    pagecache_page pp = page_lookup_nodelocked(pn, node_offset >> pc->page_order);
     pagecache_debug("%s: pn %p, node_offset 0x%lx, pp %p\n", func_ss, pn, node_offset, pp);
-    void *kvirt;
-    if (pp == INVALID_ADDRESS) {
-        kvirt = INVALID_ADDRESS;
+    range kvirt = irange(0, 0);
+    if (pp == INVALID_ADDRESS)
         goto out;
+    if (touch_or_fill_page_nodelocked(pn, pp, 0)) {
+        void *p = !private ? pp->kvirt : pagecache_get_private_page(pp);
+        if (p != INVALID_ADDRESS)
+            kvirt = irangel(u64_from_pointer(p), cache_pagesize(pc));
     }
-    if (touch_or_fill_page_nodelocked(pn, pp, 0))
-        kvirt = !private ? pp->kvirt : pagecache_get_private_page(pp);
-    else
-        kvirt = INVALID_ADDRESS;
   out:
     pagecache_unlock_node(pn);
     return kvirt;
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -1358,6 +1358,12 @@ closure_func_basic(pp_handler, boolean, pagecache_unpin_handler,
     return true;
 }
 
+void pagecache_nodelocked_unpin(pagecache_node pn, range pages)
+{
+    pagecache_nodelocked_traverse(pn, pages,
+                                  stack_closure_func(pp_handler, pagecache_unpin_handler));
+}
+
 void pagecache_node_unpin(pagecache_node pn, range pages)
 {
     pagecache_node_traverse(pn, pages, stack_closure_func(pp_handler, pagecache_unpin_handler));
--- a/src/kernel/pagecache.c
+++ b/src/kernel/pagecache.c
@@ -861,6 +861,12 @@ closure_function(3, 3, boolean, pagecache_check_old_page,
     pte old_entry = pte_from_pteptr(entry);
     boolean abort = false;
     if (pte_is_present(old_entry) && pte_is_mapping(level, old_entry) && !pte_is_dirty(old_entry)) {
+        /* An entry that stands for many cache pages is left to the unmapping path: ageing one
+           page per entry would queue a single page for eviction and then clear an entry that
+           still maps the rest, and the pages of such an entry are each referenced by it anyway,
+           so eviction could not take their memory back. */
+        if (level != PT_PTE_LEVEL)
+            return true;
         u64 pi = (pcm->node_offset + vaddr - pcm->n.r.start) >> pc->page_order;
         pagecache_node pn = pcm->pn;
         pagecache_lock_node(pn);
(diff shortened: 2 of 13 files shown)


francescolavra (changes requested):
I'm not convinced having a separate interface (with `pagecache_get_huge_window` and `pagecache_release_huge_window`) for huge pages is the best choice. I would prefer the page cache to handle pages of different sizes more transparently, so that `pagecache_get_page_if_filled()` returns an address range instead of an address value, and `pagecache_get_page()` takes a preferred page size as an additional parameter. The page cache code decides what size to allocate for a requested page (based on the `huge` attribute, which should be assigned to the pagecache volume, not to every single node), does a regular read from the underlying node (which for tmpfs means just zeroing the memory), and passes the address range (not just the base address value) to the `pagecache_page_handler` completion closure.

This change does not look simple to implement (it would likely need to be split into multiple reviewable commits), but in my opinion it would lead to cleaner and more extensible code.

AngeloRubens:
Hi @francescolavra ,

Redone the way you described, and split into six commits:

1. `pagecache_get_page_if_filled()` and `pagecache_page_handler` hand out a `range` instead of an
   address -- refactor only, nothing hands out more than a page yet.
2. `split_large_mappings()`, which describes a range at page level before a walk that expects one
   entry per page.
3. `pagecache_get_page()` takes a preferred size, the `huge` attribute lives on the volume, and
   the cache decides: for such a volume it lays one contiguous block under a whole aligned window
   of the node's pages, fills it with a single **regular read of the node**, and hands out the
   range of the whole of it. The walks that assumed one page per entry were taught otherwise.
4. tmpfs marks its volume.
5. the fault path asks for a whole window and maps the range it gets.
6. tests.

`pagecache_get_huge_window()` and `pagecache_release_huge_window()` are gone; a caller that finds
the window already mapped releases the pages with `pagecache_release_page()`, as it did for one.

One deliberate limit worth flagging: `pagecache_get_page_if_filled()` returns the range of a
single page. It does not try to recognise an already resident contiguous run, so the whole huge
path goes through `pagecache_get_page()` with the preferred size.

src/unix/mmap.c
```diff
@@ -327,10 +336,12 @@ static status demand_filebacked_page(process p, context ctx, u64 vaddr, vmap vm,
...
         return STATUS_OK;
     kvirt = (*pf)->page_kvirt;
-    if (kvirt == INVALID_ADDRESS)
+    if (range_span(kvirt) == 0)
         s = timm_oom;
     else
-        s = mmap_filebacked_page(vm, page_addr, flags, kvirt);
+        /* What came back may cover more than one page, in which case it is mapped at the base of
+           the window the faulting page falls in. */
+        s = mmap_filebacked_page(vm, page_addr & ~(range_span(kvirt) - 1), flags, kvirt);
```
francescolavra, line 344: The `& ~(range_span(kvirt) - 1)` operation should be done inside mmap_filebacked_page(), because it's needed for any call to mmap_filebacked_page(). And the above comment can be dropped.

src/kernel/page.c
```diff
@@ -280,6 +280,34 @@ void update_map_flags(u64 vaddr, u64 length, pageflags flags)
...
+        !pte_is_mapping(level, old_entry))
+        return true;
+    if (pte_split(entry, irange(0, 0)) == INVALID_PHYSICAL)
+        return false;
+    page_invalidate(bound(fe), vaddr);
+    return true;
+}
+
+/* Replaces any block mapping intersecting the range with a table of PTE-level mappings of the
+   same memory. What is mapped does not change; only the level at which it is described does.
```
francescolavra, line 299: In this comment, everything after "the same memory" is either obvious, or irrelevant and confusing. The same goes for the commit message.
And I fail to see how the title of the commit message relates to this code.

src/kernel/pagecache.c
```diff
@@ -861,6 +861,12 @@ closure_function(3, 3, boolean, pagecache_check_old_page,
     pte old_entry = pte_from_pteptr(entry);
     boolean abort = false;
     if (pte_is_present(old_entry) && pte_is_mapping(level, old_entry) && !pte_is_dirty(old_entry)) {
+        /* An entry that stands for many cache pages is left to the unmapping path: ageing one
+           page per entry would queue a single page for eviction and then clear an entry that
+           still maps the rest, and the pages of such an entry are each referenced by it anyway,
+           so eviction could not take their memory back. */
+        if (level != PT_PTE_LEVEL)
+            return true;
```
francescolavra, line 869: This comment can be made more readable by just saying that ageing of pages mapped at block level is not implemented.
And this code can go to the beginning of the function (together with the comment).

src/kernel/pagecache.c
```diff
@@ -1620,20 +1626,30 @@ closure_function(3, 3, boolean, pagecache_check_dirty_page,
     if (pte_is_present(old_entry) &&
         pte_is_mapping(level, old_entry) &&
         pte_is_dirty(old_entry)) {
-        range r = irangel(sm->node_offset + (vaddr - sm->n.r.start), cache_pagesize(pc));
+        /* One entry may stand for many cache pages, and its dirty bit is the only record that
+           any of them was written, so every page it covers is dirtied. The entry's own base can
+           precede the mapping, hence the intersection. */
+        range v = range_intersection(irangel(vaddr, pte_map_size(level, old_entry)), sm->n.r);
```
francescolavra, line 1632: Yet another less than useful comment. Remove it

src/kernel/pagecache.c
```diff
@@ -1898,17 +1914,240 @@ closure_function(3, 1, void, get_page_finish,
...
+            kvirt = irangel(u64_from_pointer(p), bound(size));
     }
     apply(handler, kvirt);
     closure_finish();
 }
 
+/* The first page of the node at or after an index, or INVALID_ADDRESS when the node has none.
+   rbtree_lookup_max_lte() answers with what precedes the index and has nothing to say when the
+   whole node sits after it, in which case the least page of the node is the answer. */
+static pagecache_page page_at_or_after_nodelocked(pagecache_node pn, u64 pi)
```
francescolavra, line 1926: Can use page_lookup_at_or_next_nodelocked() instead of duplicating its functionality

src/kernel/pagecache.c
```diff
@@ -1898,17 +1914,240 @@ closure_function(3, 1, void, get_page_finish,
...
+        return (pagecache_page)rbtree_find_first(&pn->pages);
+    if (page_offset(pp) < pi)
+        return (pagecache_page)rbnode_get_next(&pp->rbnode);
+    return pp;
+}
+
+/* Whether no page of a window holds memory, which is what makes it possible to lay one block
+   under the whole of it: a page that holds memory may be mapped. Walked from the first page at
+   or after the window rather than looked up one index at a time, since the pages of a window are
+   neighbours in the node's tree. */
```
francescolavra, line 1941: Replace this comment with "Returns whether no page of a window holds memory"

src/kernel/pagecache.c
```diff
@@ -1898,17 +1914,240 @@ closure_function(3, 1, void, get_page_finish,
...
+        if (pp->kvirt != INVALID_ADDRESS)
+            return false;
+    return true;
+}
+
+/* Gives a page its slice of the block. Called with the node and state locks held. */
+static void window_page_fill_locked(pagecache pc, pagecache_page pp, void *kvirt)
+{
+    /* The caller has established that the page holds no memory, which is the contract
+       realloc_pagelocked() states: such a page has a refcount of exactly zero. */
```
francescolavra, line 1956: Useless comment

src/kernel/pagecache.c
```diff
@@ -1898,17 +1914,240 @@ closure_function(3, 1, void, get_page_finish,
...
+    if (pp == INVALID_ADDRESS)
+        return pp;
+    init_rbnode(&pp->rbnode);
+    init_refcount(&pp->read_refcount, 0,
+                  init_closure_func(&pp->read_release, thunk, pagecache_page_read_release));
+    pp->state_offset = ((u64)PAGECACHE_PAGESTATE_ALLOC << PAGECACHE_PAGESTATE_SHIFT) | pi;
+    pp->kvirt = INVALID_ADDRESS;
+    pp->node = pn;
+    pp->l.next = pp->l.prev = 0;
+    pp->refcount = 0;
```
francescolavra, line 1982: This is setting pp->kvirt and pp->refcount just to make the relevant asserts in window_page_fill_locked() happy. I would remove those asserts and get rid of this double initialization.

src/kernel/pagecache.c
```diff
@@ -2066,6 +2307,11 @@ static void pagecache_node_unmap_pages_sync(pagecache_node pn, range v, u64 node
     thunk completion = stack_closure(pagecache_node_unmap_pages_complete, pn, unmap_entries,
                                      shared_mappings, true, &remaining);
     boolean done, progress;
+    /* The walk below records one cache page per entry, and traverse_ptes() hands it the base of
+       the entry rather than the range asked for, so a block mapping would have it work out the
+       wrong page -- and one page where the entry holds many. */
+    if (pn->pv->huge)
+        split_large_mappings(v.start, range_span(v));
```
francescolavra, line 2314: Splitting a large mapping just before unmapping is pretty inefficient if the whole mapping is to be torn down. The splitting should be done only when necessary

src/kernel/pagecache.c
```diff
@@ -2110,6 +2356,11 @@ void pagecache_node_unmap_pages(pagecache_node pn, range v /* bytes */, u64 node
         deallocate_buffer(unmap_entries);
         return pagecache_node_unmap_pages_sync(pn, v, node_offset, false);
     }
+    /* Before the node is taken: this ends in a flush that waits for every processor, and waiting
+       for that while holding the node is what parks a processor inside the flush with the node
+       in hand. */
+    if (pn->pv->huge)
+        split_large_mappings(v.start, range_span(v));
```
francescolavra, line 2363: Same as above, plus I fail to see what the comment has to do with the code.

francescolavra (changes requested):

AngeloRubens:
Thanks for the review. Rebased on master now that 2180 is in, so the branch is down to the 6 huge window commits.

The splitting is now done only where needed: split_large_mappings() is gone, and the page walk in pagecache_unmap_page_nodelocked() takes a block entry whole when the unmapped range covers it, recording the number of cache pages it maps so that the completion dirties and releases all of them. Only a block entry that the range cuts through is split, from within the walk via the new split_mapping(), and traverse_ptes() then walks the new table, as it already does for update_pte_flags(). This also removes the separate flush that the old comment was about.

The other points are addressed as suggested: the window alignment is done in mmap_filebacked_page(), the lookup uses page_lookup_at_or_next_nodelocked(), the double initialization and the asserts in window_page_fill_locked() are removed, the ageing check is at the beginning of the function, and the comments and the commit message of the page commit are trimmed.

AngeloRubens:
One more change since the reply above. A soak with a real JVM (Temurin 25, ZGC uncommitting every second while other threads write and read back every page) found a race in the unwind of a window: when another thread faulted a page into the same window while the node was unlocked for the block allocation, the pages already laid out were released from the reading state, which trips the assertion in change_page_state_locked(). They are now taken back to the allocated state before being released; the fix is in the page cache commit, and with it the soak no longer hits the assertion.

